PR #60, opened 2021-04-01
Build out proof of concept for full auth flow

kaitoo1:
all API interaction is mocked bc no backend yet
this pr adds the following:
  - retrieve nonce from backend for connected wallet address
  - request connected wallet to sign nonce message
  - exchange signature for jwt from backend
  - set jwt in storage to set loggedIn state
  
  - add sign out button and show instead of sign in button if user is loggedIn


src/components/GlobalNavbar/GlobalNavbar.tsx
```diff
@@ -1,12 +1,15 @@
+import { useAuthState } from 'contexts/auth/AuthContext';
 import styled from 'styled-components';
-import SigninButton from './SigninButton';
+import SignInButton from './SignInButton';
+import SignOutButton from './SignOutButton';
 
 function GlobalNavbar() {
+  const authState = useAuthState();
+  console.log(typeof authState);
+  const isLoggedIn = typeof authState === 'object';
```
Robinnnnn, line 9: we should probably write a [typeguard](https://basarat.gitbook.io/typescript/type-system/typeguard) for this, which will be a function like `function isLoggedIn(state: authState) ...` and returns a boolean

Robinnnnn: function could live in contexts/auth/types.ts

src/components/WalletSelector/WalletSelector.tsx
```diff
@@ -24,17 +26,26 @@ function WalletSelector() {
     error,
   } = context;
 
+  const headerMessage = active ? 'Log in with wallet' : 'Connect your wallet';
+  let signer;
+  if (library && account) {
+    signer = library.getSigner(account);
+  }
```
Robinnnnn, line 33: prefer to useMemo here instead of setting a variable on the fly, like:
```js
const signer = useMemo(() => {
  return (library && account) ? library.getSigner(account) : undefined
}, [library, account])
```

Robinnnnn (approved):
nicely done!!

src/components/GlobalNavbar/GlobalNavbar.tsx
kaitoo1: done

src/components/WalletSelector/WalletSelector.tsx
kaitoo1: done

PR #74, opened 2021-04-05
Streamline sign in flow by reducing to one visible step

kaitoo1:


src/components/WalletSelector/WalletButton.tsx
```diff
@@ -1,50 +1,68 @@
...
 };
 
 function WalletButton({ walletName, activate, connector }: WalletButtonProps) {
+  const { setStateToLoading } = useAuthActions();
+  const [isConnecting, setIsConnecting] = useState(false);
+  const authState = useAuthState();
+  const isLoadingAuth = isLoadingState(authState);
+
   const handleClick = useCallback(() => {
+    setStateToLoading();
```
Robinnnnn, line 23: hmm what will this accomplish? my worry is that setting to loading will throw the entire app into a loading state- like "APP IS LOADING"

Robinnnnn: wondering if we can accomplish this with just the local `isConnecting` state?

src/components/WalletSelector/WalletSelector.tsx
```diff
@@ -27,32 +30,35 @@ function WalletSelector() {
...
+  const { logIn } = useAuthActions();
+  const { hideModal } = useModal();
+
+  useEffect(() => {
+    if (account && isLoadingAuth && !isLoggedIn && signer) {
+      signMessageAndAuthenticate(account, signer).then((jwt) => {
+        logIn(jwt);
+        hideModal();
+      });
+    }
```
Robinnnnn, line 46: ohhh i see, so `isLoadingAuth` is set from the previous screen so that this effect can go through? if so i'm wondering if there's a better word / separate state we can use

also nitpick: i prefer `async / await` vs `.then` syntax

src/components/WalletSelector/WalletSelector.tsx
```diff
@@ -62,50 +68,60 @@ type LogInButtonProps = {
...
+      // TODO: handle case where user rejects sign request
+      throw new Error(err.message);
+    });
+
+  if (!jwt) {
+    // TODO: handle error exchanging signature for jwt
+    throw new Error('no jwt received from backend');
+  }
+  return jwt;
+};
```
Robinnnnn, line 123: dope simulator

src/contexts/auth/types.ts
```diff
@@ -1,5 +1,13 @@
 import { AuthState } from './AuthContext';
 
+export const LOGGED_OUT = Symbol('LOGGED_OUT');
+export const LOADING = Symbol('LOADING');
+export const UNKNOWN = Symbol('UNKNOWN');
+
 export function isLoggedInState(state: AuthState) {
   return typeof state === 'object';
 }
```
Robinnnnn, line 9: we should be a bit more explicit here, since `null` in javascript technically is a typeof object. let's use this:

```js
return typeof state === 'object' && state !== null && 'jwt' in state
```

that last part comes from the fact that `type LoggedIn = { jwt: string };`, so we not only wanna check for object but the fact that `jwt` key exists in the object

we also might as well move `type LoggedIn` into this file

kaitoo1: done ✅

src/components/WalletSelector/WalletSelector.tsx
kaitoo1: yeah I think we can use a state that's more specific to the WalletSelector modal.

I removed  the use of `isLoadingAuth` for now because I realized that with the new order of things, I didn't actually need to check the state to trigger the authentication methods. So we would just want some state for disabling the UI.
 
I want it to represent a state where the user has clicked on a Wallet option (like metamask). That way we can disable the whole modal and all the Wallet option buttons while the Metamask prompt is open for example

src/components/WalletSelector/WalletButton.tsx
kaitoo1: Good point, I wasn't sure what we wanted the Auth Loading state to represent, but I think I have a better idea now.

I've replied on the related comment in WalletSelector.tsx

Robinnnnn (approved):

PR #79, opened 2021-04-06
Add Dropdown and DropdownMenu components

kaitoo1:
This PR adds a DropdownMenu component that takes an array of menu item options and renders a dropdown with nav links. 
It uses the Dropdown component, which contains the logic for dropdown open/close functionality.
Example usage of DropdownMenu can be found in /scenes/Gallery/components/Header.tsx.

It also includes a minor change to rename Link component to ActionText


Robinnnnn (approved):

kaitoo1:
TODO: rename Link to ActionText

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+import TextButton from '../Button/TextButton';
+import colors from '../colors';
+
+type Props = {
+  mainText: string;
+  children?: ReactNode;
+};
+
+const refContainsEventTarget = (ref: RefObject<HTMLDivElement>, event: any) => {
+  return ref.current && ref.current.contains(event.target);
```
Robinnnnn, line 19: cool trick with optional chaining: `return ref.current?.contains(event.target)`

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+      }
+    }
+
+    // we only need to add document listener if dropdown is open
+    if (isDropdownVisible) {
+      document.addEventListener('mousedown', handleDocumentClick);
+    }
+    return () => {
+      document.removeEventListener('mousedown', handleDocumentClick);
+    };
```
Robinnnnn, line 55: gorgeous

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+  }, [dropdownMenuRef, isDropdownVisible]);
```
Robinnnnn, line 56: confused about these deps:
1. dropdownMenuRef is a react ref, which means it shouldn't need to be included as a dependency 🤔 
2. if `isDropdownVisible` is part of the dep array, does it mean whenever the state changes we run the effect? ideally we only run this effect once. i think i gotta study this a bit more

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+      );
+      const clickedOnButton = refContainsEventTarget(dropdownButtonRef, event);
+
+      if (clickedOutsideMenu && !clickedOnButton) {
+        setIsDropdownVisible(false);
+      }
+    }
+
+    // we only need to add document listener if dropdown is open
+    if (isDropdownVisible) {
```
Robinnnnn, line 50: don't think we need this if statement if we have something similar on L34. probably safe to always register `handleDocumentClick`

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const handleClick = useCallback(() => {
+    setIsDropdownVisible(!isDropdownVisible);
+  }, [isDropdownVisible]);
+
+  const dropdownMenuRef = useRef<HTMLDivElement>(null);
+  const dropdownButtonRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    function handleDocumentClick(event: any) {
```
Robinnnnn, line 33: i think this function should be defined as a `useCallback`, and we should pass an argument into the function about whether it's open (e.g. `handleDocumentClick(event: any, isDropdownVisible: boolean)`. this will make it so the callback doesn't have to depend on any react state, and it'll be a "pure function"

we want a stable function in memory so that `document.addEventListener(x, function)` will register that function once and only once, even if called multiple times.

example:
```js
// adds 3 event listners
document.addEventListener('mousedown', () => console.log('hi'))
document.addEventListener('mousedown', () => console.log('hi'))
document.addEventListener('mousedown', () => console.log('hi'))

// adds only 1 event listener!
const sayHi = () => console.log('hi')
document.addEventListener('mousedown', sayHi)
document.addEventListener('mousedown', sayHi)
document.addEventListener('mousedown', sayHi)
```

so basically within our effect if we don't want to worry about setting `handleDocumentClick` listener too many times, it should be a callback defined outside of the `useEffect` that's defined only once

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return () => {
+      document.removeEventListener('mousedown', handleDocumentClick);
+    };
+  }, [dropdownMenuRef, isDropdownVisible]);
+
+  return (
+    <StyledDropdown>
+      <div
+        style={{ width: 'fit-content', alignSelf: 'flex-end' }}
```
Robinnnnn, line 61: 1. it's not ideal to pass an object in directly, it should be memoized with useMemo...
2. that said, could we just make this a styled component like `StyledDropdownButtonContainer` instead of having to use inline styles?

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+        ref={dropdownButtonRef}
+      >
+        <StyledDropdownButton text={mainText} onClick={handleClick} />
+      </div>
+      <div ref={dropdownMenuRef} style={{ position: 'relative' }}>
```
Robinnnnn, line 66: same comment as above

src/scenes/Gallery/components/Header.tsx
```diff
@@ -28,6 +29,16 @@ function Header({ usernameOrWalletAddress }: Props) {
...
         </StyledUserDetails>
       </StyledLeftContainer>
+      <StyledRightContainer>
+        <DropdownMenu
+          mainText="Edit Profile"
+          options={[
+            { label: 'Edit name', value: '/edit/name' },
+            { label: 'Edit bio', value: '/edit/bio' },
+            { label: '+ New Collection', value: '/add/collection' },
+          ]}
```
Robinnnnn, line 39: again we would want to memoize this array, so that it's stable.

since this array is static and doesn't depend on any variables, we could move it outside of the component completely:

```js
const options = [{}, {}, {}]

function Header() {
    ...
    <DropdownMenu options={options}
}

src/components/core/Dropdown/Dropdown.tsx
kaitoo1: yeaah we can discuss in person.
I debated the "set listener only once, on component mount" vs this approach

With this approach, the document listener gets set when the dropdown opens, and is removed when it's closed. That way it's not listening for document clicks when we don't care about it. It seems a bit extra compared to setting the listener once and forgetting about it, but it also feels cleaner to me.

That was also part of the reason for not passing in a `isDropdownVisible` argument to handleDocumentClick - we should only need to handleDocumentClick if the dropdown is open/visible, so if it's called only when necessary, `isDropdownVisible` will/should always be true.

But I also like your suggestion to memoize the function and want to make sure I'm adhering to best practices. a further optimization could be to make it so the effect only runs if `isDropdownVisible === true`. We can discuss and i'll continue tinkering

src/components/core/Dropdown/Dropdown.tsx
kaitoo1: good all- I left a more detailed reply below

src/components/core/Dropdown/Dropdown.tsx
kaitoo1: 1. ok I'll memoize
2. So these non styled-component divs are a workaround I found to set a ref for the button and box properly. Styled Components didn't like it when I set the ref on them directly, so I wrapped them in a div. [Though it seems like I should be able to set it directly](https://styled-components.com/docs/advanced#refs)
And once they were wrapped I had to add these styles to get the child elements to play nice being absolute positioned and what not..

src/scenes/Gallery/components/Header.tsx
kaitoo1: yup, good call. Do you think it makes sense to have the DropdownMenu links passed in with this format? ie just passing in two values, display text and the path?

src/components/core/Dropdown/Dropdown.tsx
kaitoo1: 🧠🧠🧠

PR #84, opened 2021-04-10
Handle case where user doesnt sign message, allow retry

kaitoo1:
This PR modifies WalletSelector to
- catch errors during `signMessageAndAuthenticate`
- set an errorCode that we define, like `REJECTED_SIGNATURE`
- display an error message associated with the errorCode to the user 
- allow the user to retry connecting a wallet

### DIFF
--- a/src/components/WalletSelector/WalletSelector.tsx
+++ b/src/components/WalletSelector/WalletSelector.tsx
@@ -8,23 +8,53 @@ import { useAuthActions, useAuthState } from 'contexts/auth/AuthContext';
 import { useModal } from 'contexts/modal/ModalContext';
 import WalletButton from './WalletButton';
 import { isLoggedInState } from 'contexts/auth/types';
+import colors from 'components/core/colors';
 
 const walletConnectorMap: Record<string, AbstractConnector> = {
   Metamask: injected,
   WalletConnect: walletconnect,
 };
 
+type ErrorMessage = {
+  heading: string;
+  body: string;
+};
+
+const ERROR_MESSAGES: { [key: string]: ErrorMessage } = {
+  REJECTED_SIGNATURE: {
+    heading: 'Signature rejected',
+    body: 'Please sign the message with your wallet to continue',
+  },
+  UNKNOWN_ERROR: {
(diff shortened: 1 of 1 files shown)


src/components/WalletSelector/WalletSelector.tsx
```diff
@@ -34,13 +64,32 @@ function WalletSelector() {
   const { hideModal } = useModal();
 
   useEffect(() => {
-    if (account && !isLoggedIn && signer) {
-      signMessageAndAuthenticate(account, signer).then((jwt) => {
-        logIn(jwt);
-        hideModal();
-      });
+    if (account && isConnecting && !isLoggedIn && signer) {
```
kaitoo1, line 67: added `isConnecting` to this condition so that we only automatically trigger `signMessageAndAuthenticate` if the user has clicked a wallet option in the Wallet modal to set `isConnecting` to true.

src/components/WalletSelector/WalletSelector.tsx
```diff
@@ -8,24 +8,54 @@ import { useAuthActions, useAuthState } from 'contexts/auth/AuthContext';
...
+  UNKNOWN_ERROR: {
+    heading: 'There was an error connecting',
+    body: 'Please try again',
+  },
+};
+
+function getErrorMessage(errorCode: string) {
+  return ERROR_MESSAGES[errorCode]
+    ? ERROR_MESSAGES[errorCode]
+    : ERROR_MESSAGES.UNKNOWN_ERROR;
```
Robinnnnn, line 37: here you could use something fun called [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator): `return ERROR_MESSAGES[errorCode] ?? ERROR_MESSAGES.UNKNOWN_ERROR`

Robinnnnn (approved):

src/components/WalletSelector/WalletSelector.tsx
kaitoo1: done ✅

Issue #95, opened 2021-04-18
Stepper feature


Robinnnnn:
A step-through feature, with footer that controls steps

<img width="831" alt="Screen Shot 2021-04-17 at 10 31 23 PM 1" src="https://user-images.githubusercontent.com/12162433/115132271-b85a4300-9fcc-11eb-9536-220a1a677447.png">

PR #100, opened 2021-04-18
Wizard component

Robinnnnn:
Addresses #95 

https://user-images.githubusercontent.com/12162433/115134915-71774800-9fe2-11eb-8b7e-477c69355bb3.mov


src/components/core/Text/Text.tsx
```diff
@@ -9,7 +9,7 @@ export const Title = styled.p`
 `;
 
 type TextProps = {
-  light?: boolean;
+  color?: colors;
```
kaitoo1, line 12: oo nice call to add this as an option

src/scenes/CreateCollection/CreateCollection.tsx
```diff
@@ -0,0 +1,28 @@
...
+import OrganizeCollections from './steps/OrganizeCollections';
+
+function CreateCollection(_: RouteComponentProps) {
+  return (
+    <Wizard
+      render={({ step, next, previous }) => {
+        return (
+          <>
+            <Steps>
+              <Step id="create" render={FirstCollection} />
```
kaitoo1, line 16: maybe `FirstCollection` can be named `CreateFirstCollection` so all the steps in the wizard have a verb and it's explicit what the intended action is

kaitoo1 (approved):
approved with minor comment about step naming convention

src/scenes/CreateCollection/CreateCollection.tsx
Robinnnnn: great idea on naming convention- will make the fix and merge!

PR #106, opened 2021-04-20
Add Wallet Step to landing page

kaitoo1:
The wallet step appears after the user enters the password. After they connect their wallet, they are authenticated and are redirected to the Welcome screen.

additional changes:
  - add placeholder welcome page shown before /create
  - redirect non logged in users from /create to /

### DIFF
--- a/src/scenes/CollectionCreationFlow/CollectionCreationFlow.tsx
+++ b/src/scenes/CollectionCreationFlow/CollectionCreationFlow.tsx
@@ -1,12 +1,21 @@
 import { memo } from 'react';
-import { RouteComponentProps } from '@reach/router';
+import { Redirect, RouteComponentProps } from '@reach/router';
 import { Wizard, Steps, Step } from 'react-albus';
 import WizardFooter from './WizardFooter';
 import CreateFirstCollection from './steps/CreateFirstCollection';
 import AddNfts from './steps/AddNfts/AddNfts';
 import OrganizeCollections from './steps/OrganizeCollections';
+import { useAuthState } from 'contexts/auth/AuthContext';
+import { isLoggedInState } from 'contexts/auth/types';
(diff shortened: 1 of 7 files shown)


src/scenes/CollectionCreationFlow/CollectionCreationFlow.tsx
```diff
@@ -1,12 +1,21 @@
...
 import AddNfts from './steps/AddNfts/AddNfts';
 import OrganizeCollections from './steps/OrganizeCollections';
+import { useAuthState } from 'contexts/auth/AuthContext';
+import { isLoggedInState } from 'contexts/auth/types';
 
 function CollectionCreationFlow(_: RouteComponentProps) {
+  const authState = useAuthState();
+  const isLoggedIn = isLoggedInState(authState);
+
+  if (!isLoggedIn) {
```
kaitoo1, line 15: maybe we want to put the redirect somewhere else so we don't have add it to every page

Robinnnnn: Great call, I think reach router should have something around authenticated routing

Robinnnnn: https://github.com/Robinnnnn/good-vibes-only-client/blob/master/src/scenes/Routes.tsx#L10

This is how I did it before, but I'm not sure if it's the right way

Robinnnnn: It would be cool if we made a higher order component that's like PrivateRoute. Maybe we can pair on this next time

kaitoo1:
merging PR as is-  authenticated routes will be implemented as a follow up as discussed

PR #125, opened 2021-05-01
Fix bug where removing and adding nft staged it twice

kaitoo1:


src/scenes/CollectionCreationFlow/steps/AddNfts/Sidebar/NftPreviewIcon.tsx
```diff
@@ -16,12 +16,19 @@ function NftPreviewIcon({
...
+  // only run effect on component update, not initial render
+  // https://reactjs.org/docs/hooks-faq.html#can-i-run-an-effect-only-on-updates
+  const mountRef = useRef(false);
+  useEffect(() => {
+    if (mountRef.current) {
+      isSelected ? onStageNft(nft) : onUnstageNft(nft.id);
+    } else {
+      mountRef.current = true;
+    }
+  }, [isSelected, nft, onStageNft, onUnstageNft]);
```
Robinnnnn, line 31: this might be a pattern we want to have in the long run in other places. this might be over-engineering but we could do something like this:

```ts
export default function useEffectAfterMount(handleUpdate: () => void) {
  const mountRef = useRef(false);
  useEffect(() => {
    if (mountRef.current) {
      handleUpdate()
      return
    }
    mountRef.current = true;
  }, [handleUpdate])
}

...

import useEffectAfterMount from 'hooks/useEffectAfterMount'

function Component() {
  const handleUpdate = useCallback(() => {
    isSelected ? onStageNft(nft) : onUnstageNft(nft.id);
  }, [isSelected, nft, onStageNft, onUnstageNft])

  useEffectAfterMount(handleUpdate)
}
```

Robinnnnn (approved):

src/scenes/CollectionCreationFlow/steps/AddNfts/Sidebar/NftPreviewIcon.tsx
kaitoo1: good idea, done

PR #128, opened 2021-05-02
Make AddUserInfo trigger next if you hit Enter

kaitoo1:


src/scenes/CollectionCreationFlow/steps/AddUserInfo.tsx
```diff
@@ -28,15 +33,36 @@ function AddUserInfo() {
...
-      <StyledTextArea onChange={handleBioChange} placeholder="Bio (optional)" />
+      <StyledForm onSubmit={onSubmit}>
+        <StyledText>Add username &amp; bio</StyledText>
+        <BigInput onChange={handleUsernameChange} placeholder="Username" />
+        <Spacer height={20} />
+        <StyledTextArea
+          onChange={handleBioChange}
+          placeholder="Bio (optional)"
+        />
+      </StyledForm>
```
Robinnnnn, line 56: question: should we just throw a form around the "next" button in the WizardFooter? so basically at any point in the flow the user can click Enter to go to the next step?

this will also prevent the need to infuse this onSubmit logic into each individual step component

kaitoo1: Thats a good idea, maybe we can add it if we find ourselves doing it a lot? Because I can also think of steps where we might not want to do that, like the AddNfts step, or "reorder collections" step. 
it could be the case that "hit Enter to go to the next step" is something we only want for a limited number of steps

PR #176, opened 2021-05-07
start Nft Detail page

kaitoo1:
- Make Nft Detail Page to display a single Nft and description /info
- Responsive
- added transitions and enums file in components/core to hold all transitions and enums
- added navigation to and from gallery page <-> nft detail page
- added navigation component for going to next or prev nft in collection
- added Gradient component to Nft on Gallery page
- made fields on Nft type camelcase


src/components/GlobalNavbar/GlobalNavbar.tsx
```diff
@@ -6,8 +6,7 @@ import useIsAuthenticated from 'contexts/auth/useIsAuthenticated';
 function GlobalNavbar() {
   const isAuthenticated = useIsAuthenticated();
 
-  // hiding navbar for now
-  return true ? null : (
+  return (
```
kaitoo1, line 9: in master, GlobalNavbar is rendered in AppContainer, which was removed from main routes: https://github.com/gallery-so/gallery/blob/main/src/scenes/PossiblyDeprecated_AppContainer/PossiblyDeprecated_AppContainer.tsx



should be able to render the GlobalNavbar somewhere directly in [Routes](https://github.com/gallery-so/gallery/blob/main/src/scenes/Routes.tsx), just above the <Router> but within <Location>

example: https://codesandbox.io/s/w2m1m67nj5?file=/src/App.js

PR #177, opened 2021-05-07
Transition routes, auth refactor

kaitoo1:
- fade transitioner for routes
- make /password its own route
- handle more auth combinations (password auth vs. web3 auth)
- wallet button hover fx


src/components/WalletSelector/WalletSelector.tsx
```diff
@@ -74,28 +71,22 @@ function WalletSelector() {
   useEffect(() => {
     if (account && isPending && signer) {
       signMessageAndAuthenticate(account, signer)
-        .then((jwt) => {
-          logIn(jwt);
-          navigate('/welcome');
-        })
```
kaitoo1, line 80: we don't need to navigate here since redirect will be handled by the parent once context authenticates

src/components/FadeTransitioner/SaveForLater__Fader.tsx
```diff
@@ -0,0 +1,69 @@
+import { useEffect, useState } from 'react';
```
kaitoo1, line 1: ignore this file, committing in case we want more control over transitions in the future

src/scenes/Home/Home.tsx
```diff
@@ -1,51 +1,60 @@
...
-    return (
-      <StyledHome>
-        <WalletSelector />
-      </StyledHome>
-    );
+  if (!isPasswordValidated) {
+    return <Redirect to="/password" />;
+  }
+
+  if (!isAuthenticated) {
```
kaitoo1, line 27: I don't think we should auto redirect to the auth step, the user should see the "enter" button

Screen Shot 2021-05-08 at 11 27 01 AM

Issue #191, opened 2021-05-11
Refactor password flow


mikeydub:
Show enter button automatically as soon as a user types in the password

ie,

1) password box shown on landing page
2) type password
3) on type show enter button
4) click enter button or enter key to continue

PR #193, opened 2021-05-12
Update password flow

kaitoo1:
Addresses #191

### DIFF
--- a/src/scenes/Password/Password.tsx
+++ b/src/scenes/Password/Password.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState, useRef } from 'react';
+import { useCallback, useState, useRef, useEffect } from 'react';
 import { navigate, Redirect, RouteComponentProps } from '@reach/router';
 import styled from 'styled-components';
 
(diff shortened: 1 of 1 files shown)


src/scenes/Password/Password.tsx
```diff
@@ -13,6 +13,7 @@ import { validatePassword } from 'utils/password';
 
 function Password(_: RouteComponentProps) {
   const [storedPassword, storePassword] = usePersistedState('password', null);
+  const enteredPassword = useRef('');
```
kaitoo1, line 16: Using a ref to store the currently entered password at any given time. This allows `handleEnterGallery` to be a stable function without updating every time the entered password is changed

src/scenes/Password/Password.tsx
```diff
@@ -26,25 +27,36 @@ function Password(_: RouteComponentProps) {
...
-        setIsFormVisibleAndUnlocked(true);
-      }
-    },
-    [storePassword]
-  );
+  const handlePasswordChange = useCallback((event) => {
+    event.preventDefault();
+    const value = event.target.value;
+    enteredPassword.current = value;
+    setIsFormVisibleAndUnlocked(validatePassword(value));
```
kaitoo1, line 34: display the enter button as soon as there's a valid value

Issue #135, opened 2021-05-05
Username + bio creation error view


Robinnnnn:
e.g. username characters invalid, username taken, bio too long, etc.

Issue #132, opened 2021-05-05
Button loading state


Robinnnnn:
Button should enter a loading state when applicable, e.g. while collection is being created.

Need design

PR #216, opened 2021-05-16
Button loader, wizard button submit refactor, error handling, various other

kaitoo1:
- update UserInfoForm to be a controlled component (away from uncontrolled) [[explanation](https://stackoverflow.com/a/42522792)]. update parents in 1) welcome flow, 2) edit info modal to use new form
- centralized Loader component with customizable config (#135)
- update Button component to handle loading flag, which using above Loader (#132)
- update WizardFooter to handle async onNext submissions, displaying a loader while fetching
- update TextArea to take default value
- turn off autocomplete, autocorrect, autocapitalize, spellcheck on all inputs
- character length checker for bio (#136)
- misc

### DIFF
--- a/src/components/Profile/UserInfoForm.tsx
+++ b/src/components/Profile/UserInfoForm.tsx
@@ -15,8 +15,6 @@ type Props = {
   onUsernameChange: (username: string) => void;
   bio: string;
   onBioChange: (bio: string) => void;
-  // whether the bio form is tall
-  tall?: boolean;
 };
 
 export const BIO_MAX_CHAR_COUNT = 500;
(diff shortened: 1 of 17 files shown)


src/components/Profile/UserInfoForm.tsx
```diff
@@ -1,58 +1,82 @@
...
+export const BIO_MAX_CHAR_COUNT = 500;
+
 function UserInfoForm({
   className,
   onSubmit,
-  handleIsValidChange,
+  username,
+  onUsernameChange,
+  bio,
+  onBioChange,
```
kaitoo1, line 28: these have to be passed in rather than set on the component, since higher-level parent may want to handle submission / error handling differently

PR #222, opened 2021-05-18
Form error handling

fbrcsftnr:
- generic ErrorText component
- update BigInput to handle errors
- hook for user info form logic
- fix ugly collection naming form

### DIFF
--- a/src/components/Profile/useUserInfoForm.ts
+++ b/src/components/Profile/useUserInfoForm.ts
@@ -9,6 +9,9 @@ type Props = {
 };
 
 export default function useUserInfoForm({ onSuccess }: Props) {
+  // TODO__v1: auto-populate username and bio from useSwr()
+  // const { username, bio } = useSwr('/...')
+
   const [username, setUsername] = useState('');
   const [usernameError, setUsernameError] = useState('');
 
--- a/src/components/Profile/useUserInfoForm.ts
+++ b/src/components/Profile/useUserInfoForm.ts
@@ -74,7 +74,7 @@ export default function useUserInfoForm({ onSuccess }: Props) {
         return;
       }
       setGeneralError(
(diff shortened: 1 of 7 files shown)


src/components/Profile/useUserInfoForm.ts
```diff
@@ -0,0 +1,109 @@
...
+    setGeneralError('');
+
+    //------------ client-side checks ------------
+    if (!username.length) {
+      setUsernameError('Required.');
+      return;
+    }
+
+    if (username.length < 2 || username.length > 20) {
+      setUsernameError('Username must be between 2 and 20 characters.');
```
kaitoo1, line 33: Maybe we can put these error messages in a single const like a map

src/components/core/BigInput/BigInput.tsx
```diff
@@ -1,40 +1,52 @@
...
   );
 }
 
 const StyledBigInput = styled.input`
   padding: 0 2px;
   border: 0;
-  font-size: 32px;
-  font-family: 'Times New Roman';
-  line-height: 44px;
+  font-size: 40px;
```
kaitoo1, line 47: I changed these to font-size: 32px and line-height: 44px yesterday based on Jon's feedback in Markup, but figma says 40px and 48px 🤔
https://app.markup.io/markup/9170d884-2ef5-4b68-9b43-203780ceac06

kaitoo1: gonna ask in discord

kaitoo1 (approved):

src/components/Profile/useUserInfoForm.ts
Robinnnnn: Agreed. In an ideal world we have a set of Validators that run against the input, and each validator returns a specific error const.

Example array of validators: [required, minLength, maxLength, ...] then run your input through them all and return first error (or if no error return true)

PR #235, opened 2021-05-20
Add footer, adjust fade transition positioning

kaitoo1:


src/components/FadeTransitioner/FadeTransitioner.tsx
```diff
@@ -29,7 +28,7 @@ const childNodeStyles = {
  */
 function FadeTransitioner({ nodeKey = '', children }: Props) {
   return (
-    <TransitionGroup>
+    <TransitionGroup style={{ minHeight: '100vh' }}>
```
fbrcsftnr, line 31: since it's static, this can be defined above the component

kaitoo1: good call, done

PR #259, opened 2021-05-26
Use react-spring for parallax intro 

kaitoo1:


kaitoo1:
We might want to make the browser load these images earlier in the flow to ensure they're loaded and ready by the time we render the parallax scene

PR #272, opened 2021-06-10
On auth page, redirect existing users to profile page

kaitoo1:


src/scenes/Auth/Auth.tsx
```diff
@@ -5,23 +5,19 @@ import { RouteComponentProps, Redirect } from '@reach/router';
...
-  // const username = user?.username;
-  const userExists = false;
-  const username = '0x70d04384b5c3a466ec4d8cfb8213efc31c6a9d15';
+  const user = useAuthenticatedUser();
+  const userExists = !!user;
 
   if (isAuthenticated) {
     // if user exists in DB, send them to their profile
-    if (userExists) return <Redirect to={`/${username}`} />;
+    if (userExists) return <Redirect to={`/${user?.username}`} />;
```
fbrcsftnr, line 20: we should just be able to do `user.username` here because we ensured `user` exists

fbrcsftnr (approved):

src/scenes/Auth/Auth.tsx
kaitoo1: Compiler doesn't like it :( "Object is possibly 'null'."
I could use `user!.username` instead, either seems ok for here

PR #275, opened 2021-06-10
Prepopulate username and bio on EditUsernameInfoModal

kaitoo1:
Get the authenticatedUser and set existing username and description as default values for respective states


fbrcsftnr (approved):
YES

PR #282, opened 2021-06-19
auth update

mikeydub:


src/scenes/Auth/Auth.tsx
```diff
@@ -13,20 +13,22 @@ function Auth(_: RouteComponentProps) {
   // whether the user is web3-authenticated
   const isAuthenticated = useIsAuthenticated();
   const user = useAuthenticatedUser();
-  const userExists = !!user;
+  const username = user?.username;
+
+  if (!isPasswordValidated) {
+    return <Redirect noThrow to="/password" />;
+  }
```
mikeydub, line 20: moving this logic up so that users who haven't entered password but are authenticated get redirected to password.

PR #283, opened 2021-06-23
Update proxy to support local go server

fbrcsftnr:
`yarn start` will now automatically try to connect to go server. if it's not running, it'll spin up the local mock server

### DIFF
--- a/config/paths.js
+++ b/config/paths.js
@@ -64,7 +64,7 @@ module.exports = {
   appJsConfig: resolveApp('jsconfig.json'),
   yarnLockFile: resolveApp('yarn.lock'),
   testsSetup: resolveModule(resolveApp, 'src/setupTests'),
-  proxySetup: resolveApp('src/setupProxy.js'),
+  proxySetup: resolveApp('dev_server/server.js'),
   appNodeModules: resolveApp('node_modules'),
   swSrc: resolveModule(resolveApp, 'src/service-worker'),
   publicUrlOrPath,
--- a/dev_server/checkPort.js
+++ b/dev_server/checkPort.js
@@ -1,0 +1,23 @@
+const net = require('net');
+
+function checkPort(port) {
+  return new Promise((resolve, reject) => {
+    const server = net.createServer();
+
+    server.once('error', function (err) {
+      if (err.code === 'EADDRINUSE') {
+        reject(err);
+      }
+    });
+
+    server.once('listening', function () {
+      // if port is not in use, close silently
+      server.close();
+      resolve();
+    });
+
+    server.listen(port);
+  });
+}
+
+module.exports = checkPort;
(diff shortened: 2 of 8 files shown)


config/paths.js
```diff
@@ -64,7 +64,7 @@ module.exports = {
   appJsConfig: resolveApp('jsconfig.json'),
   yarnLockFile: resolveApp('yarn.lock'),
   testsSetup: resolveModule(resolveApp, 'src/setupTests'),
-  proxySetup: resolveApp('src/setupProxy.js'),
+  proxySetup: resolveApp('dev_server/server.js'),
```
fbrcsftnr, line 67: moved setupProxy to dedicated server directory outside of `src`, assuming we'll continue to add features

dev_server/checkPort.js
```diff
@@ -0,0 +1,23 @@
+const net = require('net');
```
fbrcsftnr, line 1: helper that detects whether a port is in use

dev_server/mocks.js
```diff
@@ -0,0 +1,115 @@
+const IMAGE_NFT = {
```
fbrcsftnr, line 1: moved mocks to own file

dev_server/routes.js
```diff
@@ -0,0 +1,110 @@
+const express = require('express');
+const MOCK_DB = require('./mocks');
+
+const router = express.Router();
```
fbrcsftnr, line 4: moved routes to own file

dev_server/server.js
```diff
@@ -0,0 +1,80 @@
...
+module.exports = function (app) {
+  app.use(
+    '/api',
+    createProxyMiddleware({
+      target: baseurl,
+      changeOrigin: true,
+      pathRewrite: {
+        // drop `/api` from path prefix, since its only purpose is to route local requests here
+        '^/api': '/',
+      },
```
fbrcsftnr, line 30: simplified path rewrite.

basically, since go server expects `/glry/v1/*`, the proxy would cause a bug by dropping this.

now we'll always have the client attach `/glry/v1`, and this is accounted for in our local server here: https://github.com/gallery-so/gallery/pull/283/files#diff-55ded419532a0b5a6cc541f4dd2a15f16e430acbd203359c699738836e3591e2R79

dev_server/server.js
```diff
@@ -0,0 +1,80 @@
...
+      console.log('Running against local mock server');
+      initializeMockServer();
+    })
+    .catch((e) => {
+      if (e.code === 'EADDRINUSE') {
+        console.log('Running against local backend');
+        return;
+      }
+      console.log('Error initializing mock server:', e);
+    });
```
fbrcsftnr, line 57: only run mock server if go server isn't running

kaitoo1 (approved):

PR #284, opened 2021-07-06
gitignore .env and replace with .env.sample

kaitoo1:
Although our `.env` file doesn't currently have sensitive values, it's best practice not to check in .env files. 

I added `.env` to our gitignore, and replaced it with a sample file that future devs can copy values from to create their own .env file locally. I also added instructions to the README for copying values from `.env.sample` to `.env` locally.

The main reasons for not checking .env files into git are to avoid publishing sensitive information, and to avoid coupling the application with any specific environment.

sources for .env best practice:
https://stackoverflow.com/questions/42014558/why-dotenv-files-should-not-be-checked-in
https://www.freecodecamp.org/news/heres-how-you-can-actually-use-node-environment-variables-8fdf98f53a0a/#:~:text=env%20files%20allow%20you%20to,in%20there%20on%20different%20lines.&text=To%20read%20these%20values%2C%20there,the%20dotenv%20package%20from%20npm.


kaitoo1:
Going to merge this to get dev env working

PR #290, opened 2021-07-24
Handle error http codes in fetcher

kaitoo1:
Since we changed our http responses to use error codes (4xx, 5xx) rather than embed the error message inside a successful response, our Fetcher should be able to detect these cases and throw an error when appropriate.

### DIFF
--- a/src/contexts/swr/fetcher.ts
+++ b/src/contexts/swr/fetcher.ts
@@ -1,5 +1,4 @@
 import { JWT_LOCAL_STORAGE_KEY } from 'contexts/auth/constants';
-import { isErrorResponse } from 'types/ApiResponse';
 
 const baseurl = process.env.REACT_APP_API_BASE_URL;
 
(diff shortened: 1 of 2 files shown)


src/contexts/swr/fetcher.ts
```diff
@@ -24,12 +23,11 @@ export default async function fetcher<ResponseData, RequestBody = {}>(
     requestOptions.method = 'POST';
     requestOptions.body = JSON.stringify(body);
   }
-
   const res = await fetch(`${baseurl}/glry/v1${path}`, requestOptions);
   const payload = await res.json();
 
-  if (isErrorResponse(payload)) {
-    throw new Error(payload.data.handler_error_user_msg);
+  if (!res.ok) {
```
fbrcsftnr, line 29: Is res.ok always built into 200 responses?

kaitoo1: Yup it should!
res.ok will be true if the response code is 2xx
Except on ie11 lol

https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

PR #293, opened 2021-07-27
Add API call to UserInfoForm, fix issues with fetcher and localstorage util

kaitoo1:
- Add API call to UserInfoForm to update username and description
- Update Fetcher to handle responses with no Response body
- Update localstorage to only stringify Objects, so that we don't stringify strings and add unnecessary double quotes to values

### DIFF
--- a/src/components/Profile/useUserInfoForm.ts
+++ b/src/components/Profile/useUserInfoForm.ts
@@ -56,10 +56,6 @@ export default function useUserInfoForm({
     }
     //------------ end client-side checks ------------
 
-    // TODO__v1: send request to server to UPDATE user's username and bio.
-    // it should be an UPDATE because a user entity will have already been
-    // created by this step
-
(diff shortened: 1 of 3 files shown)


src/components/Profile/useUserInfoForm.ts
```diff
@@ -58,6 +59,10 @@ export default function useUserInfoForm({
     // TODO__v1: send request to server to UPDATE user's username and bio.
     // it should be an UPDATE because a user entity will have already been
     // created by this step
```
fbrcsftnr, line 61: should we remove this comment?

fbrcsftnr (approved):

src/components/Profile/useUserInfoForm.ts
kaitoo1: good call, removed

PR #298, opened 2021-07-29
Add API calls for creating first collection and saving collection name

kaitoo1:
- Wrapped OrganizeCollection step with new component, `OrganizeCollectionWrapper` to provide access to collection context during wizard onNext()
- Added Create Collection response type (for create collection endpoint)
- Add API call to Create 1st Collection (called when user clicks "Create Collection", before modal appears)
- Add API call to save collection name (called when user clicks "Save" in modal)

### DIFF
--- a/src/flows/shared/steps/OrganizeCollection/CollectionNamingForm.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/CollectionNamingForm.tsx
@@ -12,6 +12,7 @@ import ErrorText from 'components/core/Text/ErrorText';
 import { useModal } from 'contexts/modal/ModalContext';
 import { Collection } from 'types/Collection';
 import { pause } from 'utils/time';
+import fetcher from 'contexts/swr/fetcher';
 
 type Props = {
   onNext: WizardContext['next'];
@@ -64,7 +65,12 @@ function CollectionNamingForm({ onNext, collection }: Props) {
 
     setIsLoading(true);
     try {
-      // TODO__v1: send request to server to UPDATE user's collection name, description
+      // TODO this endpoint only updates name. Change endpoint to also update description
+      await fetcher('/collections/update/name', {
+        id: collection.id,
+        name: collectionName,
+      });
+
       if (collectionDescription === 'invalid_desc') {
         await pause(700);
         throw { type: 'ERROR_SOMETHING_GENERIC' };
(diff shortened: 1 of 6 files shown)


src/flows/shared/steps/OrganizeCollection/OrganizeCollection.tsx
```diff
@@ -1,55 +1,74 @@
...
+  return stagedNfts.map((stagedNft: EditModeNft) => stagedNft.nft.id);
+}
+
 function useWizardConfig({ onNext }: ConfigProps) {
   const wizardId = useWizardId();
   const { setOnNext } = useWizardCallback();
   const { showModal } = useModal();
+  const stagedNfts = useStagedNftsState();
+
+  const getStagedNfts = useCallback(() => stagedNfts, [stagedNfts]);
```
kaitoo1, line 34: I think there's a better way to do this ? I wanted to access the staged nfts returned by `useStagedNftsState()` from inside the `useEffect` without adding `stagedNfts` to the `useEffect`'s dependency array. 
The callback seemed like a good way to basically have a function that I could call in the `useEffect` do that. But the callback still gets updated every time the staged nfts change which I don't really want.

Ultimately I only care about what the value of `stagedNfts` is at the time onNext() is called  - when the user clicks "Save Collection"

any ideas?
@fbrcsftnr

src/flows/shared/steps/OrganizeCollection/OrganizeCollection.tsx
```diff
@@ -1,55 +1,72 @@
...
+          }
+        );
+
         showModal(
           <CollectionNamingForm
             onNext={onNext}
-            // TODO: pass in the actual collection being organized
-            collection={{ id: '', nfts: [] }}
+            // TODO: pass in only id, or actual collection
+            collection={{ id: createdCollection.collection_id, nfts: [] }}
```
kaitoo1, line 54: pass collection name and description only instead of whole collection

src/flows/shared/steps/OrganizeCollection/CollectionNamingForm.tsx
```diff
@@ -12,36 +12,43 @@ import ErrorText from 'components/core/Text/ErrorText';
 import { useModal } from 'contexts/modal/ModalContext';
 import { Collection } from 'types/Collection';
 import { pause } from 'utils/time';
+import fetcher from 'contexts/swr/fetcher';
 
 type Props = {
   onNext: WizardContext['next'];
-  collection: Collection;
+  // collection: Collection;
```
fbrcsftnr, line 19: remove comment

fbrcsftnr (approved):

PR #302, opened 2021-08-07
401 signout / redirect

Robinnnnn:
addresses #300 

this PR logs the user out if a 401 is detected on fetch. this is made possible by migrating `fetcher` => `useFetcher`, since logout functionality is provided by `useAuthActions`, and hooks must be called by other hooks. usages of legacy `fetcher` had to be migrated.

**misc**
- noThrow on redirects to [prevent crashes](https://github.com/reach/router/issues/100#issuecomment-415424987)
- error formatter util

**outstanding TODO**
- hook up to a new JWT validation endpoint that's called first thing on app mount; if this fails, log the user out. bonus points if we can refresh


src/contexts/swr/useFetcher.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const ERR_UNAUTHORIZED = 401;
+
+export type FetcherType = <ResponseData, RequestBody = {}>(
+  path: string,
+  body?: RequestBody
+) => Promise<ResponseData>;
+
+export default function useFetcher(): FetcherType {
+  const { logOut } = useAuthActions();
```
Robinnnnn, line 15: in order to use this hook, had to convert fetcher => useFetcher

src/contexts/swr/useFetcher.ts
```diff
@@ -0,0 +1,48 @@
...
+      const res = await fetch(`${baseurl}/glry/v1${path}`, requestOptions);
+
+      // If the response doesnt have a Response stream, such as with a
+      // succesful Update request, res.json() will throw an error
+      const responseBody = await res.json().catch((e) => ({}));
+
+      if (!res.ok) {
+        if (res.status === ERR_UNAUTHORIZED) {
+          logOut();
+        }
```
Robinnnnn, line 41: main change

PR #312, opened 2021-08-13
Refactor error handling, network requests, types

Robinnnnn:
**Requests**
- Introduce `useGet` for resource fetching
  - `useSwr` should never be used directly within a component; it should be wrapped in a custom hook calling `useGet` defined in `hooks/api/*`
- Introduce `usePost` for resource mutations
  - `useFetcher` should never be used directly within a component; it should be wrapped in a custom hook calling `usePost` in `hooks/api/*`
- All requests to the API requires an action key that describes the request (e.g. `"fetch user"`, `"update collection"`). This action will be mapped to the error response if/when they occur
- Fix `res.json()` reader to differentiate legit responses vs. actual errors

**Errors**
- Introduced a custom error type called `ApiError`. This is a standard JS error that's decorated with a `customMessage` field, which is populated by our base level `_fetcher`, which in turn forwards the `action` from the call site
- Error Boundary has been updated to better reflect errors. This uses `formatDetailedError` that can extract custom messages from our internal errors, with sane defaults

**LocalStorage**
- Only cache successful responses locally under `__swr__` prefix; prevent over-eager caching "errors" (failed requests) and "validating" (in-flight requests)

**Directories**
- `types/*` contains primitive types only; API-specific types such as `UserRequest` and `UserResponse` are bundled next to the hooks that actually use them
- In general, types should be defined close to the call site unless it is truly generic. As a result `type EditModeNft` has been moved within `OrganizeGallery`, given that it likely won't be used outside of that view


kaitoo1 (approved):

kaitoo1:
BIG

PR #316, opened 2021-08-14
Null check user.addresses

kaitoo1:
Super edge case (a user without an `addresses` array is an invalid state for user in database) but I encountered it so I thought we might as well handle it gracefully on the front end. Null check everything we can

### DIFF
--- a/src/components/GlobalNavbar/LoggedInNav.tsx
+++ b/src/components/GlobalNavbar/LoggedInNav.tsx
@@ -6,7 +6,10 @@ import TextButton from 'components/core/Button/TextButton';
 import Dropdown from 'components/core/Dropdown/Dropdown';
 import Spacer from 'components/core/Spacer/Spacer';
 import CopyToClipboard from 'components/CopyToClipboard/CopyToClipboard';
-import { useAuthenticatedUser } from 'hooks/api/users/useUser';
+import {
+  useAuthenticatedUser,
+  useAuthenticatedUserAddress,
+} from 'hooks/api/users/useUser';
 import { User } from 'types/User';
 
 function truncate(address: string) {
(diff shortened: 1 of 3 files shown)


src/components/GlobalNavbar/LoggedInNav.tsx
```diff
@@ -17,7 +17,13 @@ function LoggedInNav() {
...
+  let userAddresses = user.addresses || [];
+
+  if (!userAddresses[0]) {
+    console.error('User is missing addresses');
+  }
+
+  const walletAddress = userAddresses[0] || '';
   const truncatedAddress = useMemo(() => {
     return truncate(walletAddress);
   }, [walletAddress]);
```
Robinnnnn, line 29: i think we should just have this in a hook and have it throw an error; this will allow the Suspense boundary to catch the error, and we don't have to do `if X throw error` within the component itself

e.g.
```js
export default function useAuthenticatedUserAddress() {
  const user = userAuthenticatedUser()
  if (!user) {
    throw new Error('Authenticated user not found')
  }
  const address = user.addresses?.[0]
  if (!addresses) {
    throw new Error('No addresses found on user')
  }
  return address
}

...

function LoggedInNav() {
  ...
  const address = useAuthenticatedUserAddress()
  ...
}
```

kaitoo1: brilliant

kaitoo1: It renders the error via the Error Boundary momentarily then unmounts/shows the dev error message, is that expected for local? Console errors indicate the error wasn't caught which is odd `index.js:81 Uncaught Error: No address found on user`
Screen Shot 2021-08-16 at 12 04 17 PM

PR #323, opened 2021-08-22
Edit user modal

Robinnnnn:


src/components/Profile/useUserInfoForm.ts
```diff
@@ -56,9 +58,9 @@ export default function useUserInfoForm({
     //------------ end client-side checks ------------
 
     try {
-      await updateUser(username, bio);
+      await updateUser(userId, username, bio);
 
-      onSuccess();
+      onSuccess(username);
```
Robinnnnn, line 63: updated username is passed to onSuccess in case we need to navigate there

src/hooks/api/users/useUpdateUser.ts
```diff
@@ -1,18 +1,34 @@
...
+        [`/users/get?username=${username}`, 'fetch user'],
+        (user: User) => ({ ...user, username, bio }),
+        false
+      );
+
+      mutate(
+        [`/users/get?user_id=${userId}`, 'fetch user'],
+        (user: User) => ({ ...user, username, bio }),
+        false
+      );
```
Robinnnnn, line 30: mutations are how we allow to client to update immediately: https://swr.vercel.app/docs/mutation

PR #324, opened 2021-08-23
Edit Collection Info Integration

Robinnnnn:


src/flows/shared/steps/OrganizeCollection/CollectionEditInfoForm.tsx
```diff
@@ -17,8 +17,8 @@ import { Collection } from 'types/Collection';
 type Props = {
   onNext: WizardContext['next'];
   collectionId: Collection['id'];
-  collectionTitle?: Collection['title'];
-  collectionDescription?: Collection['description'];
+  collectionTitle?: Collection['name'];
```
kaitoo1, line 20: maybe we can rename the prop names to match as well

kaitoo1: `collectionName`, `collectionNote` (or `collectionCollectorsNote` if we want to match explicitly)

src/flows/shared/steps/OrganizeGallery/CollectionDnd.tsx
```diff
@@ -59,15 +59,16 @@ function CollectionDnd({ collections }: Props) {
   }, []);
 
   const activeCollection = useMemo(() => {
-    return collectionOrder.find(({ id }) => id === activeId);
-  }, [activeId, collectionOrder]);
+    return collections.find(({ id }) => id === activeId);
```
kaitoo1, line 62: So activeCollection now checks the collection from SWR instead of the local state's collection?

kaitoo1 (approved):

src/flows/shared/steps/OrganizeCollection/CollectionEditInfoForm.tsx
Robinnnnn: great call, done in a separate PR!

src/flows/shared/steps/OrganizeGallery/CollectionDnd.tsx
Robinnnnn: yup exactly- this will handle the small edge case where the shadow of the collection being dragged may be stale (e.g. old collection name)

PR #326, opened 2021-08-23
Fix onboarding bugs

Robinnnnn:


src/components/WalletSelector/authRequestUtils.ts
```diff
@@ -157,27 +155,12 @@ async function createUser(
...
-  fetcher: FetcherType
-) {
+async function triggerOpenseaSync(address: string, fetcher: FetcherType) {
   try {
-    /**
-     * upon fetching the user's nfts from opensea, we'll stick them directly in the
-     * "unassigned NFTs" cache. this will allow the Create First Collection step to
-     * immediately display the user's unassigned NFTs by the time they get there
-     */
-    const response = await fetcher<OpenseaSyncResponse>(
```
Robinnnnn, line 171: doing away with attempting optimistic update since the initial opensea sync doesn't return the exact same data that's eventually stored with the user (no IDs attached, etc.)

also, we fetch the unassigned NFTs anyway as the source of truth when the user gets to edit step

kaitoo1 (approved):

PR #333, opened 2021-08-27
Move options under 'Edit Profile' to LoggedInNavbar, and get rid of EditProfile

kaitoo1:
Account options


Robinnnnn (approved):
looks awesome!!

a couple small things:
- i feel like it's more intuitive to see the wallet address first in the list. maybe it's just me?
- i think we should display more of the wallet address, or maybe put more periods in between. basically i think it could be wider lol

kaitoo1:
good idea, I reordered the items to put wallet first and increased the truncated address length

Screen Shot 2021-08-29 at 10 05 40

PR #339, opened 2021-09-13
Add more mixpanel events

kaitoo1:
Added more Mixpanel events


src/flows/OnboardingFlow/steps/AddUserInfo.tsx
```diff
@@ -19,8 +20,6 @@ function useWizardConfig({ onNext }: ConfigProps) {
 
   useEffect(() => {
     setOnNext(onNext);
-
-    return () => setOnNext(undefined);
```
kaitoo1, line 23: I removed this because it was resetting the next step's next(), since it was being called after the next step had already called `setOnNext()`.

PR #347, opened 2021-09-16
Sync opensea on collection editor

Robinnnnn:


src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
```diff
@@ -66,6 +63,9 @@ function StagingArea() {
       >
         <SortableContext items={stagedNfts}>
           <StyledStagedNftContainer>
+            <StyledHeadingWrapper>
+              <Heading>Your collection</Heading>
+            </StyledHeadingWrapper>
```
kaitoo1, line 68: oo if we move the heading into the DND, we could remove the fixed width from the heading, and set it to width: 100% instead
on line 89

kaitoo1 (approved):

PR #357, opened 2021-09-21
Adjust Gallery Profile Page content

kaitoo1:
Changes
- Bump up username header font size
- adjust spacing
- Add `Curator since {created_at}`
- Fix word break/wrapping for bio text  and reduce total width of bio text box

### DIFF
--- a/src/components/core/Text/Text.tsx
+++ b/src/components/core/Text/Text.tsx
@@ -50,7 +50,7 @@ export const Subdisplay = styled(H2)`
   letter-spacing: 0px;
 
   @media only screen and ${breakpoints.tablet} {
-    font-size: 40px;
+    font-size: 48px;
     line-height: 48px;
   }
 `;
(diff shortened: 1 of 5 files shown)


kaitoo1:
Blocked by backend change to return `created_at` for User Get

kaitoo1:
Backend PR: https://github.com/gallery-so/go-gallery/pull/161/files

src/components/core/Text/Text.tsx
```diff
@@ -50,7 +50,7 @@ export const Subdisplay = styled(H2)`
   letter-spacing: 0px;
 
   @media only screen and ${breakpoints.tablet} {
-    font-size: 40px;
+    font-size: 48px;
```
kaitoo1, line 53: `Subdisplay` is only used by the Gallery username header and the ErrorBoundary message so changing the font size here should be ok

Robinnnnn (approved):

PR #361, opened 2021-09-22
Display owner username and link to gallery if exists

kaitoo1:
Changes:
- add owner_username to Nft type
- on NFT Detail Page, display owner_username and link to user's Gallery
- if no owner_username, display owner_address and link to Etherscan


src/scenes/NftDetailPage/NftDetailText.tsx
```diff
@@ -41,6 +34,27 @@ function NftDetailText({ nft }: Props) {
   );
 }
 
+type NftOwnerProps = {
+  owner_address: string;
```
Robinnnnn, line 38: nit: would prefer camel case for props, but not a big deal!

Robinnnnn (approved):

src/scenes/NftDetailPage/NftDetailText.tsx
kaitoo1: oh good point I'll fix!

PR #366, opened 2021-09-24
Coinbase Wallet

Robinnnnn:
Screen Shot 2021-09-23 at 9 21 30 PM


src/components/WalletSelector/WalletButton.tsx
```diff
@@ -56,12 +59,22 @@ function WalletButton({
       return null;
     }
 
+    function convertWalletName(name: string) {
+      const overrides: Record<string, string> = {
```
kaitoo1, line 63: Maybe this `overrides` map can be defined outside of the component

kaitoo1 (approved):

src/components/WalletSelector/WalletButton.tsx
Robinnnnn: was thinking about this, but i think it's kinda nice to have it right where the logic is used (fewer mental hops)

in the future if we need the map in more places we can move

PR #371, opened 2021-09-27
Ensure NFT is rendered properly if image_url links to a video

kaitoo1:
Changes
- Add NftPreviewAsset component that will render an Image by default, but will render a video if the NFT doesn't have an animation_url, and has an image_url that ends with .mp4
- Refactor how we determine media type for a given nft. Consolidate nft-related util funds in utils/nft.ts
- Add test for `getMediaType()` util function


src/components/NftPreview/NftPreviewAsset.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+type Props = {
+  nft: Nft;
+};
+
+function NftPreviewAsset({ nft }: Props) {
+  const setContentIsLoaded = useSetContentIsLoaded();
+  const nftAssetComponent = useMemo(() => {
+    if (getMediaTypeForAssetUrl(nft.image_url) === NftMediaType.VIDEO) {
+      return <StyledVideo src={'https://storage.opensea.io/files/7acbd389bd9bfdd376e428b685e6c15a.mp4'} onLoadStart={setContentIsLoaded}></StyledVideo>;
```
kaitoo1, line 18: update url

PR #380, opened 2021-09-29
Add Helvetica Neue for Windows compatibility

kaitoo1:
Helvetica Neue is not a system font on windows so add it to project


src/components/core/Text/Text.tsx
```diff
@@ -84,7 +84,6 @@ export const BodyRegular = styled(Paragraph)`
   font-family: ${BODY_FONT_FAMILY};
   font-size: 14px;
   line-height: 20px;
-  letter-spacing: 0.4px;
```
kaitoo1, line 87: the new font has more spacing by default

PR #411, opened 2021-10-21
Add Membership Mint Page

kaitoo1:
Changes:
- Added ABI for membership contract
- Changed Wallet Selector flow to allow just connecting a wallet without Gallery auth
- Added support for testnets
- Added MembershipMintPage which will render the contract/membership data, and allow a user to connect their wallet and mint

### DIFF
--- a/src/scenes/MembershipMintPage/MembershipMintPage.tsx
+++ b/src/scenes/MembershipMintPage/MembershipMintPage.tsx
@@ -103,11 +103,11 @@ function MembershipMintPage({ membershipColor }: Props) {
   return (
     <StyledMintPage centered>
       <StyledContent>
-        <StyledMedia>
+        <div>
           <ShimmerProvider>
             <MembershipVideo src={membershipProperties.videoUrl}/>
           </ShimmerProvider>
-        </StyledMedia>
+        </div>
         <StyledDetailText>
           <Heading>{membershipProperties.title}</Heading>
           <Spacer height={16} />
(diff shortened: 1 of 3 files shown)


src/contexts/AppProvider.tsx
```diff
@@ -13,9 +15,12 @@ export default function AppProvider({ children }: Props) {
     <Boundary>
       <ToastProvider>
         <AuthProvider>
-          <SwrProvider>
-            <ModalProvider>{children}</ModalProvider>
-          </SwrProvider>
+          <Web3ProviderNetwork>
```
kaitoo1, line 18: this is the recommended way to maintain another web3 connection - this way we can have a web3 connection to read contract data to render the contract details on a page without a user connecting their wallet,, and a separate connection to do user wallet stuff
https://github.com/NoahZinsmeister/web3-react/tree/v6/docs#createweb3reactroot

src/scenes/MembershipMintPage/MembershipMintPage.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const getPrice = useCallback(async (contract: Contract) => {
+    const priceRes = await contract.getPrice(membershipProperties.tokenId);
+    setPrice(priceRes);
+  }, [membershipProperties.tokenId]);
+
+  useEffect(() => {
+    if (contract) {
+      void getCanMintToken(contract);
+      void getRemainingSupply(contract);
+      void getPrice(contract);
```
Robinnnnn, line 90: try/catch?

PR #467, opened 2021-11-20
MP pageview and console

Robinnnnn:


config/webpack.config.cjs
```diff
@@ -35,10 +35,10 @@ const appPackageJson = require(paths.appPackageJson);
 const shouldUseSourceMap = process.env.GENERATE_SOURCEMAP !== 'false';
```
Robinnnnn, line 35: ignore most changes here, except the one commented

config/webpack.config.cjs
```diff
@@ -258,6 +258,7 @@ module.exports = function (webpackEnv) {
               // Pending further investigation:
               // https://github.com/terser-js/terser/issues/120
               inline: 2,
+              drop_console: true,
```
Robinnnnn, line 261: main change. since eslint is no longer banning console logs, webpack will drop it before shipping to prod

src/contexts/navigation/GalleryNavigationContext.tsx
```diff
@@ -1,18 +1,32 @@
 import { globalHistory, navigate } from '@reach/router';
```
Robinnnnn, line 1: ignore most changes here, except the one commented

src/contexts/navigation/GalleryNavigationContext.tsx
```diff
@@ -40,76 +58,91 @@ type Props = {
...
-      </NavigationActionsContext.Provider>
-    </NavigationStateContext.Provider>
-  );
-});
+    }, [pathname, sanitizedPathname]);
+
+    // track pageviews
+    useEffect(() => {
+      Mixpanel.track('Page view');
+    }, [sanitizedPathname]);
```
Robinnnnn, line 81: only change

src/utils/mixpanel.ts
```diff
@@ -1,14 +1,19 @@
 import mixpanel from 'mixpanel-browser';
```
Robinnnnn, line 1: accidentally hit save in this file

.eslintrc.cjs
```diff
@@ -8,7 +8,7 @@ const rules = {
   'node/prefer-global/process': ['error', 'always'],
   radix: 'off',
   'unicorn/no-array-reduce': 'off',
-  'no-console': ['error', { allow: ['warn', 'error'] }],
+  'no-console': 'off',
```
kaitoo1, line 11: should we change this to `warn`

config/webpack.config.cjs
kaitoo1: oh interesting, does this only drop console.logs or other consoles like error/warn

Robinnnnn: warn wasn't an option, but noticing in my code that it underlines it anyway (without erroring)

.eslintrc.cjs
Robinnnnn: warn wasn't an option, but noticing in my code that it underlines it anyway (without erroring)

config/webpack.config.cjs
Robinnnnn: looks like it drops everything. should we keep some stuff around?

image

Robinnnnn: image

.env.sample
```diff
@@ -10,3 +10,9 @@ MOCK_SERVER_BASE_URL=http://localhost:3500
 
 # Ethirium Chain ID for reading contract data using Network Connector
 REACT_APP_NETWORK_CONNECTOR_CHAIN_ID=4
+
+# Mixpanel API key
```
kaitoo1, line 14: I was thinking we don't need default for these vars in the sample because devs usually wouldn't need to run analytics locally.

If `REACT_APP_MIXPANEL_TOKEN` and `REACT_APP_ANALYTICS_API_URL` are defined it'll try to log the event and sends a network request to to `ourApi/api.bingbonglyrics.co` lmao

could pollute logs if the api base url is pointed to dev or prod api

kaitoo1: Or we could change the switch for enabling analytics to something else

kaitoo1 (approved):

.env.sample
Robinnnnn: lol truuu. okay makes sense, should take that out!

.eslintrc.cjs
kaitoo1: oh ok sweet!

PR #489, opened 2021-12-15
Allow user to retry connecting with Gnosis

kaitoo1:


src/contexts/auth/AuthContext.tsx
```diff
@@ -97,14 +96,14 @@ const AuthProvider = memo(({ children }: Props) => {
         setToken(payload.jwt);
         setUserId(payload.userId);
         setAuthState(payload);
-        setUserSigninAddress(address);
+        setUserSigninAddress(address.toLowerCase());
```
kaitoo1, line 99: this is the only change in this file

src/components/ManageWallets/ManageWallets.tsx
```diff
@@ -26,30 +26,23 @@ function ManageWallets({ newAddress }: Props) {
   }, [showAddWalletModal]);
 
   const addWalletDisabled = useMemo(() => addresses.length >= 5, [addresses]);
-  const [userSigninAddress] = usePersistedState(
-    USER_SIGNIN_ADDRESS_LOCAL_STORAGE_KEY,
-    ''
-  );
+  const [userSigninAddress] = usePersistedState(USER_SIGNIN_ADDRESS_LOCAL_STORAGE_KEY, '');
```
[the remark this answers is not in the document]
Robinnnnn: ty!

src/components/WalletSelector/AddWalletPending/AddWalletPendingGnosisSafe.tsx
```diff
@@ -54,69 +62,140 @@ function AddWalletPendingGnosisSafe({
...
+        } catch (error: unknown) {
+          handleError(error);
+        }
       }
     }
 
-    void authenticate();
-  }, [account, authenticatedUserAddresses, attemptAddWallet]);
+    void initiateAuthentication();
+  }, [
```
Robinnnnn, line 190: may not be necessary, but are there any threads/processes that are being opened as a result of this `useEffect`? if so it's best practice to clean them up as a return func, but not sure if needed here

```js
useEffect(() => {
  // do stuff

  return () => {} // clean stuff up
}, [])
```

src/components/WalletSelector/GnosisSafePendingMessage.tsx
```diff
@@ -1,15 +1,71 @@
...
+import { Web3Provider } from '@ethersproject/providers/lib/web3-provider';
+import { getLocalStorageItem } from 'utils/localStorage';
+
+const GNOSIS_NONCE_STORAGE_KEY = 'gallery_gnosis_nonce';
 
 type Props = {
   pendingState: PendingState;
   userFriendlyWalletName: string;
+  onRestartClick?: () => void;
+  manuallyValidateSignature: () => void;
```
Robinnnnn, line 18: superhypernit: we should have a convention where function props are named with `on`; otherwise this could be confused for a boolean

src/components/WalletSelector/walletUtils.ts
```diff
@@ -122,8 +123,8 @@ export async function listenForGnosisSignature(
   const listenToGnosisSafeContract = new Promise((resolve) => {
     gnosisSafeContract.on(GNOSIS_SAFE_SIGN_MESSAGE_EVENT_NAME, async (msgHash: any) => {
       // Upon detecing the SignMsg event, validate that the contract signed the message
-      const magicValue = await gnosisSafeContract.isValidSignature(messageHash, '0x');
-      const messageWasSigned = magicValue === GNOSIS_VALID_SIGNATURE_MAGIC_VALUE;
+      console.log('Gnosis Safe Contract Event: ', GNOSIS_SAFE_SIGN_MESSAGE_EVENT_NAME);
```
Robinnnnn, line 126: do we want to keep this?

src/components/WalletSelector/walletUtils.ts
```diff
@@ -136,6 +137,25 @@ export async function listenForGnosisSignature(
...
+    .catch(() => false);
+  return magicValue === GNOSIS_VALID_SIGNATURE_MAGIC_VALUE;
+}
+
+export async function validateNonceSignedByGnosis(
+  address: string,
+  nonce: string,
+  library?: Web3Provider
+) {
+  console.log('Querying Gnosis Safe');
```
Robinnnnn, line 155: same comment

src/components/WalletSelector/GnosisSafePendingMessage.tsx
```diff
@@ -1,15 +1,71 @@
...
 import Spacer from 'components/core/Spacer/Spacer';
 import { LISTENING_ONCHAIN, PendingState, PROMPT_SIGNATURE } from 'types/Wallet';
 import colors from 'components/core/colors';
 import styled, { keyframes } from 'styled-components';
+import { useMemo, useEffect } from 'react';
+import { useWeb3React } from '@web3-react/core';
+import { Web3Provider } from '@ethersproject/providers/lib/web3-provider';
+import { getLocalStorageItem } from 'utils/localStorage';
+
+const GNOSIS_NONCE_STORAGE_KEY = 'gallery_gnosis_nonce';
```
Robinnnnn, line 12: looks like we can just import this from storageKeys now!

src/utils/localStorage.ts
```diff
@@ -0,0 +1,5 @@
+export function getLocalStorageItem(key: string) {
+  const item = window.localStorage.getItem(key);
+  const parsedItem = item && (JSON.parse(item) as string);
+  return parsedItem ?? '';
+}
```
Robinnnnn, line 5: not necessary in this PR, but is it worth using `usePersistedState`?

https://github.com/gallery-so/gallery/blob/main/src/hooks/usePersistedState.ts

src/components/WalletSelector/walletUtils.ts
kaitoo1: I thought it might be useful to be able to see that the SignMsg event was detected - otherwise we can't really tell that the event was detected which has slowed down debugging - though do we remove console.logs on other envs?

src/utils/localStorage.ts
kaitoo1: I tried using persistedState at first, but for this use case I wanted a super simple way to just read a value from storage without syncing it to the state or anything, so I thought this util func might be useful

src/components/WalletSelector/AddWalletPending/AddWalletPendingGnosisSafe.tsx
kaitoo1: will fast follow with optimizations here as discussed

PR #493, opened 2021-12-22
Add partnership mint page, refactor mint page

kaitoo1:


Robinnnnn:
Why does this PR have a vercel deployment error while the other one doesn't?

src/contexts/membershipMintPage/MembershipMintPageContext.tsx
```diff
@@ -0,0 +1,113 @@
...
+type Props = { children: ReactNode };
+
+const MembershipMintPageProvider = memo(({ children }: Props) => {
+  const [MembershipMintPageState, setMembershipMintPageState] = useState<MembershipMintPageState>({
+    totalSupply: 0,
+    remainingSupply: 0,
+    price: 0,
+  });
+
+  const setTotalSupply = useCallback((totalSupply: number) => {
```
Robinnnnn, line 59: Supernit: if you have 3 separate setX functions, I'd prefer to keep the state separate (e.g. [totalSupply, setTotalSupply] etc) instead of one big object

Robinnnnn: Then you can avoid having to do the object spreading / merging in each callback

src/scenes/MembershipMintPage/MembershipMintPage.tsx
```diff
@@ -272,7 +228,17 @@ function MembershipVideo({ src }: VideoProps) {
   return <StyledVideo src={src} autoPlay loop playsInline muted onLoadStart={setContentIsLoaded} />;
 }
 
-const StyledMintPage = styled(Page)`
+function MembershipNftVisual({ src }: VideoProps) {
+  return (
+    <div>
```
Robinnnnn, line 233: Is the div needed here?

src/scenes/MembershipMintPage/MerkleTree.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  getProof(el) {
+    let idx = this.bufIndexOf(el, this.elements);
+
+    if (idx === -1) {
+      throw new Error('Element does not exist in Merkle tree');
+    }
+
+    console.log({ idx });
```
Robinnnnn, line 80: There are some console logs in this file

src/scenes/MembershipMintPage/PartnerMembershipMintPage.test.tsx
```diff
@@ -0,0 +1,19 @@
+import { OpenseaAsset, reduceOpenseaAssetsToContractAddresses } from './PartnerMembershipMintPage';
+
+describe('reduceOpenseaAssetsToContractAddresses', () => {
+  test('', () => {
```
Robinnnnn, line 4: Love that this is tested! This test block should have a name, currently its empty string

src/scenes/MembershipMintPage/PartnerMembershipMintPage.tsx
```diff
@@ -0,0 +1,178 @@
...
+// Checks to see if the address owns any partner NFTs by
+// retrieving all NFTs owned by a user from opensea,
+// and filtering the list by whitelisted partner contract addresses
+async function detectOwnedPartnerNftsFromOpensea(
+  account: string,
+  addresses: string[]
+): Promise<string[]> {
+  const params = getContractAddressQueryParams(addresses);
+  const response = await fetch(
+    `https://testnets-api.opensea.io/api/v1/assets?owner=${account}${params}`,
```
Robinnnnn, line 71: Maybe we shouldn't hardcode testnet?

src/scenes/MembershipMintPage/PartnerMembershipMintPage.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const [ownedPartnerNfts, setOwnedPartnerNfts] = useState<string[]>([]);
+  const [ownsGeneralCard, setOwnsGeneralCard] = useState(false);
+  const canMintToken = useMemo(
+    () => !ownsGeneralCard && ownedPartnerNfts.length > 0,
+    [ownedPartnerNfts.length, ownsGeneralCard]
+  );
+
+  // Regardless of how many partner NFTs they own, we just need one contract address to mint with, so pick first one in list.
+  const partnerContractAddress = useMemo(() => ownedPartnerNfts[0], [ownedPartnerNfts]);
```
Robinnnnn, line 113: Nit: this line and the one above we don't have to memo since they're constant time computations- memo is more useful if we're filtering an array or finding for example

src/scenes/MembershipMintPage/PartnerMembershipMintPage.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  // use a ref to track if we called OS, to limit to only calling them once
+  const calledOpenseaRef = useRef(false);
+  useEffect(() => {
+    async function checkIfUserCanMint(account: string, addresses: string[]) {
+      const ownsGeneralCard = await detectOwnedGeneralCardsFromOpensea(account);
```
Robinnnnn, line 119: Confusing to have ownsGeneralCard defined twice, one here and another outside of this scope

src/scenes/MembershipMintPage/PartnerMembershipMintPage.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+      if (ownsGeneralCard) {
+        calledOpenseaRef.current = true;
+        setOwnsGeneralCard(true);
+        return;
+      }
+
+      const ownedPartnerNfts = await detectOwnedPartnerNftsFromOpensea(account, addresses);
+      calledOpenseaRef.current = true;
```
Robinnnnn, line 128: Looks like we're setting this to true both here and in the if statement above. Maybe we just call it once above the if statement? Since it's always getting called

src/contexts/membershipMintPage/MembershipMintPageContext.tsx
kaitoo1: makes sense! done

src/scenes/MembershipMintPage/MembershipMintPage.tsx
kaitoo1: not needed, removed

src/scenes/MembershipMintPage/MerkleTree.ts
kaitoo1: removed

src/scenes/MembershipMintPage/PartnerMembershipMintPage.test.tsx
kaitoo1: added description

src/scenes/MembershipMintPage/PartnerMembershipMintPage.tsx
kaitoo1: Made it configurable via env config!

src/scenes/MembershipMintPage/PartnerMembershipMintPage.tsx
kaitoo1: true, done

src/scenes/MembershipMintPage/PartnerMembershipMintPage.tsx
kaitoo1: good point, renamed internal one to `generalCardDetectedInAccount`

src/scenes/MembershipMintPage/PartnerMembershipMintPage.tsx
kaitoo1: the idea here was to set `calledOpenseaRef.current = true;` once we were finished successfully calling opensea for each case. 

So in the case where `ownsGeneralCard` is true, we don't need to call OS anymore so we set the ref to true immediately in the if block.

Otherwise, we need to call `detectOwnedPartnerNftsFromOpensea` so we wait until that's successfully called, then set the ref to true after.

To try to make this clearer I've renamed the ref to `finishedCallingOpensea`. Lmk if that makes sense

PR #502, opened 2022-01-06
Remove reach router

tbezman:
## Reach Router Removal.

Originally, we planned to move to `next/router` incrementally route by route. But after looking around and trying that out, it felt significantly more complex and buggy than to just do it all one in sweep. I'll outline some of the high level changes I had to make to get this done.

## Build Step Change
Now that we're using `getServerSideProps`. We have to switch to Vercel's Next.JS preset on the build step. Otherwise the project won't build (this is why the build is failing on the PR). @Robinnnnn , once this PR lands, we can switch the build step, and test out on staging. **For now, I recommend you pull the branch locally and run the project to test things out**.

### Checklist
- [x] Add 404 pages
  - UserGalleryPage
  - NftDetailPage
- [x]  Replace usages of navigate with push
  - `push` is the idiomatic equivalent of `navigate` from Reach Router, except in one case for us `navigate(-1)` should be `back()` in Next Router.
- [x]  Remove RouteComponentProps
  - RouteComponentProps is how you get path params in Reach Router. But in Next.JS You get your path params by doing one of the following
    - Using `query` on `useRouter()`. But on the first render, your path param is undefined which is really annoying.
    - The better way is to pull them out of `getServerSideProps` which has them right the first time and then you pass them in as props to the page component.
- [x]  Move routes to appropriate directories
  - You can read more about Next's routing file structure [here](https://nextjs.org/docs/routing/introduction). We no longer have a `Routes` component. All routes are handled by the file structure in Next.JS.
- [x]  Figure out how to get FadeTransitioner working
  - Works the same way as before, but now the `locationKey` comes from the `asPath` on `useRouter()`. This is done inside of `_app.tsx`.
- [x]  Migrate GalleryNavigationContext
    - [x]  Figure out how to do scroll restoration
      - Scroll Restoration is handled for us in Next.JS which is working very well. The one case that gave me issues is a little jump jank when navigating away with the `FadeTransitioner`. You can read about my fix in a comment left in the code `GalleryNavigationProvider.tsx` 
      -  carbon
    - [x]  Track pageview to Mixpanel `Mixpanel.track('Page view')`
      - This is now done in `_app.tsx` as a side effect of the `asPath` off `useRouter()`.
    - [x]  Figure out how to track router stack length (Maybe use window.history?)
      - We just track this using native window `pushState` and `popState` events.


pages/_app.tsx
```diff
@@ -14,18 +18,30 @@ const SafeHydrate: FC = ({ children }) => (
...
-      <Component {...pageProps} />
-    </SafeHydrate>
-  </>
-);
+}> = ({ Component, pageProps }) => {
+  const { asPath } = useRouter();
+
+  useEffect(() => {
+    Mixpanel.track('Page view', { path: asPath });
+  }, [asPath]);
```
tbezman, line 26: New effect to track page views.

pages/_app.tsx
```diff
@@ -14,18 +18,30 @@ const SafeHydrate: FC = ({ children }) => (
...
+    <>
+      <Head>
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <meta name="description" content="Show your collection to the world." />
+
+        <title>Gallery</title>
+      </Head>
+      <SafeHydrate>
+        <AppProvider>
+          <FadeTransitioner locationKey={asPath}>
```
tbezman, line 38: Add `FadeTransitioner` so we can do the cool animations.

src/scenes/_Router/GalleryRedirect.tsx
```diff
@@ -8,9 +9,10 @@ type Props = {
...
 export default function GalleryRedirect({ to }: Props) {
-  return (
-    <Page>
-      <Redirect to={to} noThrow />
-    </Page>
-  );
+  const { replace } = useRouter();
+  useEffect(() => {
+    void replace(to);
+  }, [replace, to]);
```
tbezman, line 15: This is up for debate but I felt like the right behavior here is to replace the current route so they don't press the back button just to get redirected again.

src/scenes/NftDetailPage/NavigationHandle.tsx
```diff
@@ -28,9 +28,12 @@ function NavigationHandle({ direction, nftId }: Props) {
 
   const hoverText = useMemo(() => HOVER_TEXT.get(direction) ?? '', [direction]);
 
+  const { push } = useRouter();
   const handleOnClick = useCallback(() => {
-    void navigate(nftId, { state: { collection: [] } });
-  }, [nftId]);
+    // TODO(Terence): Figure out how to get this state across since Next doesn't support navigation state
+    // void push(nftId, { state: { collection: [] } });
+    void push({ pathname: nftId });
```
tbezman, line 35: @Robinnnnn Can we pair up on this? It looks like this component is commented out and not being used so maybe we can just delete it?

Robinnnnn: this is a fine for now, don't worry about it, we do eventually wanna use this component

kaitoo1 (approved):

PR #503, opened 2022-01-07
Make user + nft page titles dynamic

kaitoo1:
Make the Page title dynamic, so that we include the username on the Gallery Page, and username+artwork name on the Detail Page.

- [x] title, og:title and twitter:title are dynamic on the User Gallery Page, displaying the username. Tab says "Kaito | Gallery" for example
- [x] title, og:title and twitter:title are dynamic on the NFT Detail Page, displaying the artwork name + username. Tab says "Obit #1 - Kaito | Gallery"  for example

example:
Screen Shot 2022-01-04 at 19 37 22

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -19,6 +19,8 @@ const App: FC<{
     <Head>
       <meta name="viewport" content="width=device-width, initial-scale=1" />
       <meta name="description" content="Show your collection to the world." />
+      <meta property="og:title" content="Gallery" key="og:title" />
+      <meta name="twitter:title" content="Gallery" key="twitter:title" />
 
       <title>Gallery</title>
     </Head>
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -19,8 +19,27 @@ const App: FC<{
     <Head>
       <meta name="viewport" content="width=device-width, initial-scale=1" />
       <meta name="description" content="Show your collection to the world." />
+      <meta name="msapplication-TileColor" content="#da532c" />
+      <meta name="theme-color" content="#ffffff" />
+
+      <meta property="og:type" content="website" />
       <meta property="og:title" content="Gallery" key="og:title" />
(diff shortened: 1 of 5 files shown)


pages/_app.tsx
```diff
@@ -19,6 +19,8 @@ const App: FC<{
     <Head>
       <meta name="viewport" content="width=device-width, initial-scale=1" />
       <meta name="description" content="Show your collection to the world." />
+      <meta property="og:title" content="Gallery" key="og:title" />
```
kaitoo1, line 22: I moved `og:title` and `twitter:title` out of _document.tsx and into _app.tsx, because it looks like tags in `_document` don't get overridden even when you use the `key` identifier. So without doing that, the page ended up with two `og:title` tags - one set by _document and one set by NftDetailPage for example.

Next js docs recommended putting tags that can be page-specific in pages and components instead, which worked.


https://stackoverflow.com/questions/68204973/avoid-duplicate-meta-description-and-keywords-in-next-js
https://nextjs.org/docs/advanced-features/custom-document

kaitoo1: and the `key` identifier is used to prevent duplicate tags.

Robinnnnn (approved):
not a nextjs expert but LGTM! 🔥

tbezman:
@kaitoo1 Don't want to block this PR but maybe as a fast follow we can clean up the Head stuff all over the app. Maybe we should move all of the meta tags into `_app.tsx` so we can change them on the page level in the future. I'm not sure I had a good reason to split them between `_app.tsx` and `_document.tsx` the way I did but I remember Next.JS complaining when I had certain `meta` tags in the `_document.tsx` file. 

I think we should just move all of the meta tags into `_app.tsx` so we can change them.

kaitoo1:
@tbezman  Ah gotcha - sounds good, I've moved all the meta tags from `_document.tsx` to `_app.tsx`. thanks for looking

PR #535, opened 2022-01-22
Show 'More Info' button on assets that have external_url

connorrothschild:
Looks like this ([example](https://gallery-n3d8zs5s1-gallery-so.vercel.app/connorr/23d4gAV1jK7EorGmfphDTTD7tuK/23sellWLGTpApQwMi7Vslgj3mqR)): 

image

### DIFF
--- a/src/scenes/NftDetailPage/NftAdditionalDetails.tsx
+++ b/src/scenes/NftDetailPage/NftAdditionalDetails.tsx
@@ -62,14 +62,21 @@ function NftAdditionalDetails({ nft }: Props) {
           <BodyRegular color={colors.gray50}>Token ID</BodyRegular>
           <BodyRegular>{hexHandler(nft.opensea_token_id)}</BodyRegular>
           <Spacer height={16} />
-          {hasContractAddress && (
-            <StyledLinkContainer>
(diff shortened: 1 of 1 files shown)


src/scenes/NftDetailPage/NftAdditionalDetails.tsx
```diff
@@ -62,14 +62,21 @@ function NftAdditionalDetails({ nft }: Props) {
           <BodyRegular color={colors.gray50}>Token ID</BodyRegular>
           <BodyRegular>{hexHandler(nft.opensea_token_id)}</BodyRegular>
           <Spacer height={16} />
-          {hasContractAddress && (
-            <StyledLinkContainer>
-              <StyledLink href={getOpenseaExternalUrl(nft)} target="_blank" rel="noreferrer">
-                <ActionText color={colors.gray50}>View on OpenSea</ActionText>
+          <StyledLinkContainer>
```
connorrothschild, line 65: `StyledLinkContainer` does not take up any space/have any visual appearance if it is blank so this is acceptable even if the asset is missing a contract address and an external URL

Robinnnnn (approved):

PR #537, opened 2022-01-24
Initial Relay Configuration + Memberhsip Page => GraphQL

tbezman:
Look ma, no network requests on the client!

https://user-images.githubusercontent.com/6754223/150749263-d55f0e51-8f84-4a4f-81f6-a7dde0da63e9.mov

## High Level
This PR bootstraps Gallery with Relay and migrates the `/members` page to GraphQL by making a REST request behind the scenes and converting that response to a GraphQL Response.

### Changes Made
- Moved off of SWC back to babel until https://github.com/vercel/next.js/pull/33240 lands.
- Added relay dependencies to our `package.json`
- Added a first pass schema `schema.graphql`. This schema is very malleable at the moment as Ezra and I  keep iterating.
- Added a mechanism for serializing and hydrating the Relay cache from the server (more on this in a moment)

## How are we fetching GraphQL without a GraphQL server setup?
Since we don't have a GraphQL server and I wanted to show y'all what Relay looks like in practice. I setup Relay's fetch function to make a rest request behind scenes and then supply a proper GraphQL response. We will not be using this pattern going forward in production. I see us using it while we're waiting for a backend eng to do their work so we can test locally, but it should almost never be in production (excluding this PR).

## How are we avoiding network requests on the client?
In Relay, every single page has a query. And everything below that query uses a fragment. With this mechanism, we are able to make the request to fetch all of the data for a single page in a single query. In a `getServerSideProps` next to the page component, we create a temporary server side relay environment, fetch the query, and then serialize the cache to be sent as a `__relayCache` prop to the client. Then in `_app.tsx`, we look in `pageProps` and create a client side relay environment from the passed `__relayCache`. The next time relay stumbles across that component, it won't make a request since the data is already in the cache.

## Why isn't the page server side rendered if we are fetching the data on the server?
The short answer is we're not quite there yet. If we want to turn on SSR, we have to do it site wide. This means we have to go around the app and ensure we're not referencing `window` anywhere on the server.

I took a stab at removing some references but quickly broke auth. @kaitoo1 Is there any reason we still need to store JWTs in local storage? It seems like we can remove all of that stuff now that we have cookie based auth.

## Some notes on our schema
If you look in our schema, you may notice that everything is nullable. This is an informed decision from my experience with GraphQL and follows what Facebook does. The philosophy here is that the backend can technically fail to resolve anywhere in the schema. We need to handle these scenarios as gracefully as possible. Keeping the schema nullable means we can let the client determine how to handle a server side failure. 

### How do we avoid annoying null checks if everything is nullable?
Use the `@required` directive from Relay. You can read more [here](https://relay.dev/docs/guides/required-directive/#internaldocs-banner).

### What is this `removeNullValues` I see?
With our fully nullable schema, anything can be nullable, even items in a list. So if we ask GraphQL for a list of Nfts, we may get `[nft1, nft2, null, nft4]` as a result if the GraphQL server had a failure while resolving data around the third nft. `removeNullValues` removes the nulls from the list and fixes your types so you don't have to deal with the nulls. We'll be using this helper function a lot.


src/contexts/swr/useFetcher.ts
```diff
@@ -39,7 +39,8 @@ export const _fetch: FetcherType = async (path, action, parameters = {}) => {
     credentials: 'include',
   };
 
-  const localJwt = window.localStorage.getItem(JWT_LOCAL_STORAGE_KEY);
+  const localJwt =
+    typeof window === 'undefined' ? null : window.localStorage.getItem(JWT_LOCAL_STORAGE_KEY);
```
tbezman, line 43: Had to change this so we can make requests on the server. Hopefully this stuff can go away now that we have Cookie auth

kaitoo1:
Thanks for the detailed write up @tbezman  - we can definitely remove all jwt logic since we moved to cookies. I'll remove this week

.babelrc
```diff
@@ -0,0 +1,4 @@
+{
+  "presets": ["next/babel"],
```
Robinnnnn, line 2: hehe

Robinnnnn (approved):

kaitoo1 (approved):

src/scenes/MemberListPage/MemberListTier.tsx
```diff
@@ -15,21 +18,35 @@ function getPreviewDirection(index: number) {
...
+  const { fadeUsernames, searchQuery } = useMemberListPageState();
 
-  const { searchQuery } = useMemberListPageState();
   const sortedOwners = useMemo(() => {
-    if (!tier.owners) {
-      return [];
-    }
+    const nonNullOwners = removeNullValues(tier.owners ?? []);
+
+    nonNullOwners.sort((a, b) => a.user.username.localeCompare(b.user.username));
```
kaitoo1, line 47: it looks like we lost the username.toLowerCase() here. some usernames are capitalized and we want to ignore case when sorting the list alphabetically

PR #541, opened 2022-01-25
Better handling of hexadecimals 

connorrothschild:
This fixes long string conversion so that it outputs a number instead of scientific notation 🙏 e.g. this shows properly now https://gallery-git-connor-hex-fix-gallery-so.vercel.app/robin/21Qt76R6bIyM7fIDsBsgU9DrKCg/21R0nPXR1nlqXP9UD8hn7CQUsUP

### DIFF
--- a/src/scenes/NftDetailPage/NftAdditionalDetails.tsx
+++ b/src/scenes/NftDetailPage/NftAdditionalDetails.tsx
@@ -12,9 +12,15 @@ type Props = {
 };
 
 // The backend converts all token IDs to hexadecimals; here, we convert back
-// TODO: Fix parseInt() for long strings (> 21 characters?), when JS converts to scientific https://gallery.so/robin/21Qt76R6bIyM7fIDsBsgU9DrKCg/21R0nPXR1nlqXP9UD8hn7CQUsUP
-const hexHandler = (str: string) => parseInt(str, 16);
-
+const hexHandler = (str: string) => {
+  if (str.length % 2) {
+    str = '0' + str;
+  }
+  const bn = BigInt('0x' + str);
+  const d = bn.toString(10);
+  return d
+}
+  
 const getOpenseaExternalUrl = (nft: Nft) => {
   const contractAddress = nft.asset_contract.address;
   const tokenId = hexHandler(nft.opensea_token_id);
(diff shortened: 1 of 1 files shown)


Robinnnnn (approved):

src/scenes/NftDetailPage/NftAdditionalDetails.tsx
```diff
@@ -12,8 +12,15 @@ type Props = {
 };
 
 // The backend converts all token IDs to hexadecimals; here, we convert back
-// TODO: Fix parseInt() for long strings (> 21 characters?), when JS converts to scientific https://gallery.so/robin/21Qt76R6bIyM7fIDsBsgU9DrKCg/21R0nPXR1nlqXP9UD8hn7CQUsUP
-const hexHandler = (str: string) => parseInt(str, 16);
+const hexHandler = (str: string) => {
+  if (str.length % 2) {
+    str = '0' + str;
+  }
```
Robinnnnn, line 18: what exactly is the purpose of this?  if the string length is even we add a `0` in front? i think we should add a comment to explain!

kaitoo1: would love a comment here!

src/scenes/NftDetailPage/NftAdditionalDetails.tsx
```diff
@@ -12,8 +12,16 @@ type Props = {
...
 // The backend converts all token IDs to hexadecimals; here, we convert back
-// TODO: Fix parseInt() for long strings (> 21 characters?), when JS converts to scientific https://gallery.so/robin/21Qt76R6bIyM7fIDsBsgU9DrKCg/21R0nPXR1nlqXP9UD8hn7CQUsUP
-const hexHandler = (str: string) => parseInt(str, 16);
+// https://stackoverflow.com/a/53751162
+const hexHandler = (str: string) => {
+  if (str.length % 2) {
+    str = '0' + str;
+  }
+
+  const bn = BigInt('0x' + str);
```
kaitoo1, line 21: for future readability I think we should use fully written out variable names, that way I can understand what it represents without looking through the rest of the function

PR #542, opened 2022-01-25
Add more detailed events for authentication flow to discern attempt, success, and error cases

kaitoo1:
We currently only track one event during the whole auth flow - when it's successful. 
This alone isn't very useful for us to measure how the flow is performing.

This PR introduces more detailed events:
`trackSignInAttempt`
`trackSignInSuccess`
`trackSignInError`
and equivalent for `AddWallet`

these get recorded for Logging In and Multi Wallet for both EOA and Contract Accounts. 
no PII such as address is recorded.

### DIFF
--- a/src/utils/mixpanel.ts
+++ b/src/utils/mixpanel.ts
@@ -1,4 +1,5 @@
 import mixpanel from 'mixpanel-browser';
+import { isWeb3Error } from 'types/Error';
 
 if (process.env.NEXT_PUBLIC_MIXPANEL_TOKEN && process.env.NEXT_PUBLIC_ANALYTICS_API_URL) {
   mixpanel.init(process.env.NEXT_PUBLIC_MIXPANEL_TOKEN, {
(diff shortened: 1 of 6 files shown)


src/utils/mixpanel.ts
```diff
@@ -1,29 +1,92 @@
...
+  ) => {
+    Mixpanel.track(eventName, {
       wallet_provider: walletName,
       connection_mode: connectionMode,
+      ...(errorMessage && { error: errorMessage }),
     });
   },
+
+  trackSignInAttempt: (walletName: string) => {
+    Mixpanel.trackAuthEvent('Sign In - Attempt', walletName, AUTH_MODE_SIGN_IN);
```
Robinnnnn, line 58: i wonder if the Auth Event should always just be `"Sign In"`, and `success | attempt` can be a metadata field?

like `Mixpanel.track('SignIn', { action: 'success' })`

src/utils/mixpanel.ts
```diff
@@ -1,29 +1,92 @@
...
+    Mixpanel.trackAuthEvent(
+      'Sign In - Error',
+      walletName,
+      AUTH_MODE_SIGN_IN,
+      getAuthErrorMessage(error)
+    );
+  },
+
+  trackAddWalletAttempt: (walletName: string) => {
+    Mixpanel.trackAuthEvent('Add Wallet - Attempt', walletName, AUTH_MODE_ADD_WALLET);
```
Robinnnnn, line 75: same comment here

PR #544, opened 2022-01-25
Support poolsuite allowlist

Robinnnnn:


kaitoo1:
love that the lists are separate!!

PR #548, opened 2022-01-26
Remove usage of jwt and add get current user + logout calls

kaitoo1:
## Context
This PR is mainly JWT related clean up now that we are on cookie based auth. The most significant change is that because we can no longer rely on a JWT in local storage to quickly determine the user's authenticated state on app load, I instead added use of a new endpoint that returns the current user if there is a valid cookie present. There are related changes to handling 401 unauthorized errors as well.
The backend changes [are in this PR](https://github.com/gallery-so/go-gallery/pull/240). 
^ **the linked PR should be merged and deployed first**

## Summary of changes: 
- removed all uses of jwt
- in AuthContext, if the auth state is `UNKNOWN`, try to get the current user. 
    - if a user is returned, set the logged in state
    - if a user is not returned, set the logged out state.
         - if we detect that the user used to be logged in but no user is returned, display a toast saying 'session expired'.
- added a `handleUnauthorized` action that is used in the fetcher that will set the logged out state and potentially show an expired sesh toast if the request returns a 401
- Replace `LoggedInState` that had jwt and userId with simpler Symbol `LOGGED_IN`

example of 401 handling for expired token
Screen Shot 2022-01-27 at 16 42 02


src/components/ManageWallets/ManageWallets.tsx
```diff
@@ -3,7 +3,7 @@ import colors from 'components/core/colors';
 import Spacer from 'components/core/Spacer/Spacer';
 import ErrorText from 'components/core/Text/ErrorText';
 import { BodyMedium, BodyRegular } from 'components/core/Text/Text';
-import { USER_SIGNIN_ADDRESS_LOCAL_STORAGE_KEY } from 'contexts/auth/constants';
+import { USER_SIGNIN_ADDRESS_LOCAL_STORAGE_KEY } from 'constants/storageKeys';
```
kaitoo1, line 6: we had local storage keys in 2 files so this is consolidation

src/contexts/auth/AuthContext.tsx
```diff
@@ -52,97 +53,102 @@ export const useAuthActions = (): AuthActions => {
...
+   * Sets the user state to logged out and clears local storage
    */
-  const setLoggedOut = useCallback(() => {
-    setAuthState(LOGGED_OUT);
-  }, []);
+  const setLoggedOut = useCallback(
+    (isUnauthorized = false) => {
+      if (isUnauthorized) {
+        pushToast(EXPIRED_SESSION_MESSAGE);
+      }
```
Robinnnnn, line 73: 😍

src/contexts/auth/AuthContext.tsx
```diff
@@ -52,97 +53,102 @@ export const useAuthActions = (): AuthActions => {
...
   }, []);
 
-  const { pushToast } = useToastActions();
+  const handleUnauthorized = useCallback(() => {
+    setLoggedOut(true);
+  }, [setLoggedOut]);
+
+  useEffect(() => {
+    setIsLoggedInLocally(authState === LOGGED_IN);
+  }, [authState, setIsLoggedInLocally]);
```
Robinnnnn, line 120: a bit concerned there will be a race condition between this effect and the one below.

according to the line above, `authState` will always begin as `UNKNOWN`:
```js
const [authState, setAuthState] = useState<AuthState>(UNKNOWN);
```

this means this effect could at first call `setIsLoggedInLocally(UNKNOWN === LOGGED_IN)`, which results in `false`, even though the other useEffect may end up determining that the `authState` is `LOGGED_IN`.

the worst case is that the user's previous local state is logged in, but this effect wipes it out and sets it to false, and we're unable to determine whether they were logged in before.

could we just remove this effect and call `setIsLoggedInLocally` in the below effect? (commented there)

src/contexts/auth/AuthContext.tsx
```diff
@@ -52,97 +53,102 @@ export const useAuthActions = (): AuthActions => {
...
-          // const response = await _fetch<ValidateJwtResponse>('/auth/jwt_valid', 'validate jwt');
-          // if (!response.valid) {
-          //   pushToast('Your session is invalid or expired. Please try again.');
-          //   logOut();
-          //   return;
-          // }
-
-          await logIn({ jwt: token, userId }, userSigninAddress);
+        if (authenticatedUser) {
+          setAuthState(LOGGED_IN);
```
Robinnnnn, line 128: here add: `setIsLoggedInLocally(true)`

src/contexts/auth/AuthContext.tsx
```diff
@@ -52,97 +53,102 @@ export const useAuthActions = (): AuthActions => {
...
           return;
         }
 
-        setLoggedOut();
-      }
+        // if no authenticated user is returned but they were logged in previously, show a toast
+        if (isLoggedInLocally) {
+          pushToast(EXPIRED_SESSION_MESSAGE);
+        }
```
Robinnnnn, line 135: this should continue to work because `isLoggedInLocally` will come from persisted state

src/contexts/auth/AuthContext.tsx
```diff
@@ -52,97 +53,102 @@ export const useAuthActions = (): AuthActions => {
...
 
-      if (authState === UNKNOWN) {
-        void loadAuthState();
+        setAuthState(LOGGED_OUT);
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      } catch (_error: unknown) {
+        setAuthState(LOGGED_OUT);
```
Robinnnnn, line 140: here we can `setIsLoggedInLocally(false)`

src/contexts/auth/AuthContext.tsx
kaitoo1: makes sense! yeah totally valid concern, good catch.
I originally moved it out into a separate useEffect because I was seeing the metamask sign message open twice and thought that something in here was causing it to run more than once. I didn't see it anymore after I moved it so I thought that did the trick

But I just tried moving it all back into the same one  and it looks like it's behaving as expected still  💪

Robinnnnn (approved):

PR #538, opened 2022-01-24
Add toggle between grid+list view on mobile

kaitoo1:
GRID
Screen Shot 2022-01-24 at 22 26 39

LIST
Screen Shot 2022-01-24 at 22 26 32

PR #551, opened 2022-01-29
Added Singe Collection Page

jakzaizzat:
### Entry point to this page
Click any collection section in the gallery to go this page

image

### Possible use case

1. Collection with name & description

CleanShot 2022-01-29 at 12 59 20@2x

2. Collection with name

CleanShot 2022-01-29 at 12 59 54@2x

3. Collection without name & description
CleanShot 2022-01-29 at 13 00 14@2x

### Expected Behaviour

- [x]  Single Collection Page
    - [x]  The page renders a single collection, determined by the url.
    - [x]  If the collection belongs to the user viewing the page, render the “Edit Name & Collection” and “Edit Collection” options.
    - [x]  The page has a “Share” button which is displayed for all users.
    - [x]  Clicking an NFT navigates the user to the NFT Detail Page.
    - [x]  Clicking on the username in the header navigates the user back to the Gallery Page.
- [x]  Gallery Page
    - [x]  Clicking on the title of a collection navigates the user to the Single Collection Page
    - [x]  The “...” menu for each collection the Gallery Page has a “Share” option that copies the collection page url to the clipboard when clicked
- [x]  NFT Detail Page
    - [x]  If the user arrived on the NFT Detail Page by clicking through from the Single Collection Page, the “Back to Gallery” button should say “Back to Collection” instead.

### Changes for existing codebase
* Move the Collection Gallery into his own component. Since, we are using it for gallery and collection page. 
* Convert mobileLayout to its own hooks
* Refactor the `useCollectionById` to call collection by id api instead of using authenticated hooks since we need to support public api.


src/components/NftPreview/NftPreview.tsx
```diff
@@ -34,11 +34,16 @@ const LAYOUT_DIMENSIONS: Record<number, number> = {
...
   const navigateToUrl = useNavigateToUrl();
 
+  const username = window.location.pathname.split('/')[1];
+  const storage = globalThis?.sessionStorage;
+
   const handleNftClick = useCallback(
     (event: React.MouseEvent<HTMLElement>) => {
-      navigateToUrl(`${window.location.pathname}/${collectionId}/${nft.id}`, event);
+      event.stopPropagation();
+      if (storage) storage.setItem('prevPage', window.location.pathname);
```
Robinnnnn, line 43: i know we're trying to avoid using the window due to SSR. @tbezman any recs on the best approach here?

src/flows/shared/steps/OrganizeGallery/CollectionRowSettings.tsx
```diff
@@ -21,9 +23,12 @@ type Props = {
 function CollectionRowSettings({ collection, wizard: { push } }: Props & WizardComponentProps) {
   const { showModal } = useModal();
   const { setCollectionIdBeingEdited } = useCollectionWizardActions();
+  const username = useAuthenticatedUsername();
 
   const { id, name, collectors_note, hidden } = collection;
 
+  const collectionUrl = `${window.location.origin}/${username}/${id}`;
```
Robinnnnn, line 30: similar comment as above. looks like this article has some suggestions: https://stackoverflow.com/questions/65199051/how-to-get-page-url-or-hostname-in-nextjs-project

src/hooks/api/collections/useUpdateCollectionInfo.ts
```diff
@@ -46,6 +47,19 @@ export default function useUpdateCollectionInfo() {
...
+      await mutate(
+        getCollectionByIdCacheKey({ id: collectionId }),
+        (value: GetCollectionResponse) => {
+          const newValue = cloneDeep<any>(value);
+          newValue.collection.name = name;
+          newValue.collection.collectors_note = collectors_note;
+          return newValue
+        },
+        false
+      );
```
Robinnnnn, line 62: thank you for dealing with our weird meme lol

src/flows/shared/steps/OrganizeGallery/CollectionRowSettings.tsx
jakzaizzat: Thanks for the link! Let me try this first

src/hooks/api/collections/useUpdateCollectionInfo.ts
jakzaizzat: It works 😂

src/scenes/CollectionGalleryPage/CollectionGalleryHeader.tsx
```diff
@@ -0,0 +1,175 @@
...
+  mobileLayout: DisplayLayout;
+  setMobileLayout: (mobileLayout: DisplayLayout) => void;
+};
+
+function CollectionGalleryHeader({ collection, mobileLayout, setMobileLayout }: Props) {
+  const { showModal } = useModal();
+  const { push } = useRouter();
+  const screenWidth = useBreakpoint();
+  const user = usePossiblyAuthenticatedUser();
+  const username = useMemo(() => window.location.pathname.split('/')[1], []);
```
Robinnnnn, line 34: same comment here. though i'm pretty sure this logic already existed on another page!

src/scenes/CollectionGalleryPage/CollectionGalleryHeader.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  const unescapedCollectorsNote = useMemo(() => unescape(collection.collectors_note || ''), [
+    collection.collectors_note,
+  ]);
```
Robinnnnn, line 38: nit: memoing here is overkill i think

src/scenes/CollectionGalleryPage/CollectionGalleryHeader.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  const authenticatedUserIsOnTheirOwnPage = username.toLowerCase() === user?.username.toLowerCase();
+
+  const collectionUrl = window.location.href;
```
Robinnnnn, line 42: same comment!

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useCallback, useEffect, useMemo } from 'react';
+import usePersistedState from 'hooks/usePersistedState';
+import { useBreakpoint } from 'hooks/useWindowSize';
+import usePrevious from 'hooks/usePrevious';
+import Head from 'next/head';
+import CollectionGallery from './CollectionGallery';
+
+// suggest mobile users to use landscape mode for 3-column experience.
+// use localStorage to prevent showing the toast again.
+function useSuggestion() {
```
Robinnnnn, line 15: we can remove this! no longer used anymore: https://github.com/gallery-so/gallery/pull/538/files#diff-121f304ab5e398e8f28b52e3657b0a12c10d9d0de6ee320a3825a23decbee7e0

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+  ]);
+}
+
+type CollectionGalleryPageProps = {
+  username: string;
+  collectionId: string;
+};
+
+function CollectionGalleryPage({ collectionId, username }: CollectionGalleryPageProps) {
+  useSuggestion();
```
Robinnnnn, line 58: can remove

src/scenes/NftDetailPage/NftDetailPage.tsx
```diff
@@ -19,10 +19,13 @@ type Props = {
...
 
 function NftDetailPage({ nftId }: Props) {
-  const { replace, back } = useRouter();
+  const { replace, back, query } = useRouter();
   const canGoBack = useCanGoBack();
 
   const username = window.location.pathname.split('/')[1];
+  const collectionId = query.collectionId;
+  const isFromCollectionPage =
+    globalThis?.sessionStorage?.getItem('prevPage') === `/${username}/${collectionId}`;
```
Robinnnnn, line 28: same comment... we should brainstorm what the best way is to see where they've come from. we can ask terence next week.

jakzaizzat: Sure! The current way of dealing kind of hacky things..

tbezman: Seems like we could refactor some of the navigation context stuff to store a stack of what routes we've been on so we could actually track this. without local storage.

tbezman: It's not blocking though, we can land this

jakzaizzat: @tbezman I like this idea. If I'm not mistaken, currently we only store only number of stack in the context.

kaitoo1:
I pulled down the branch and it looks great! awesome PR 🔥

Just a couple of things I noticed:
- I can't seem to proceed after clicking "Save Collection" on the edit view. This should take me back to the Collection Page. The Back button works as expected for me. 👍
- It looks like the left alignment of the NFTs is a bit different on the Gallery Page and Collection Page when a collection is small and the NFTs don't take up the whole row. examples:
Gallery Page: 
Screen Shot 2022-01-30 at 21 22 44

Collection Page:
Screen Shot 2022-01-30 at 21 23 53

To resolve this, I think we need an extra wrapping div with `width: 100%` around the `NftGallery` component in `CollectionGallery`:
Screen Shot 2022-01-30 at 21 22 10


Otherwise looks great to me.. once these are resolved we can merge it! We can sync up on next steps tomorrow, but broadly I imagine we'll want to test it out for a bit on dev and see if the feature as a whole feels right and ready to ship, then we can release it. I'll talk to the rest of the team to coordinate.

kaitoo1 (approved):

PR #558, opened 2022-02-01
refactor gallery to use display:grid instead of flex

kaitoo1:
This is a refactor PR to change our Gallery layout to use display: grid instead of display: flex.

This allows column widths to be automatically calculated while preserving the desired behavior of evenly spacing the nfts on rows that are complete(using space-between) while left-aligning rows that are incomplete.
Being able to use space-between to automatically space nfts apart means we don't have to use margin, which means we don't have to then use negative margin to re-align the row with the parent container.

This also fixes an issue where the right side of the collection wasn't flush with the parent container.

before:
Screen Shot 2022-02-01 at 22 38 06

after:
Screen Shot 2022-02-01 at 22 37 48

Confirmed the following:
- works on Brave (chrome) , Safari, Firefox
- works for 1-6 column collections (1 column layouts are displayed at 50% width)
- works on Gallery and Single Collection Page

### DIFF
--- a/src/components/NftGallery/NftGallery.tsx
+++ b/src/components/NftGallery/NftGallery.tsx
@@ -30,7 +30,6 @@ function NftGallery({ collection, mobileLayout }: Props) {
   );
 }
 
-// grid-template-columns: repeat(${({ columns }) => columns}, minmax(auto, 50%));
 const StyledCollectionNfts = styled.div<{ columns: number; mobileLayout: DisplayLayout }>`
   display: grid;
   grid-template-columns: ${({ columns, mobileLayout }) =>
--- a/src/components/NftGallery/NftGallery.tsx
+++ b/src/components/NftGallery/NftGallery.tsx
@@ -1,4 +1,4 @@
-import { DEFAULT_COLUMNS, LAYOUT_GAP_BREAKPOINTS } from 'constants/layout';
+import { DEFAULT_COLUMNS } from 'constants/layout';
 import breakpoints from 'components/core/breakpoints';
 import { DisplayLayout } from 'components/core/enums';
 import NftPreview from 'components/NftPreview/NftPreview';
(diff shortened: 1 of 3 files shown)


src/components/NftGallery/NftGallery.tsx
```diff
@@ -24,36 +24,26 @@ function NftGallery({ collection, mobileLayout }: Props) {
...
     </StyledCollectionNfts>
   );
 }
 
 const StyledCollectionNfts = styled.div<{ columns: number; mobileLayout: DisplayLayout }>`
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: ${({ columns, mobileLayout }) =>
+    mobileLayout === DisplayLayout.LIST ? '1fr' : `repeat(${columns},  minmax(auto, 50%))`};
```
Robinnnnn, line 36: what is this black magic

src/components/NftPreview/NftPreview.tsx
```diff
@@ -119,39 +108,6 @@ const StyledNftPreview = styled.div<{ columns: number }>`
...
-    width: ${({ columns }) =>
-      columns === 1
-        ? `${SINGLE_COLUMN_NFT_WIDTH}px;`
-        : `calc((100% - ${LAYOUT_GAP_BREAKPOINTS.desktop * columns}px) / ${columns});`}
-    margin: ${LAYOUT_GAP_BREAKPOINTS.desktop / 2}px;
-  }
-`;
-
-const StyledNftPreviewList = styled(StyledNftPreview)`
-  width: 100%;
```
Robinnnnn, line 154: 😍😍😍😍

Robinnnnn (approved):

src/components/NftGallery/NftGallery.tsx
kaitoo1: its kinda crazy, im learning display:grid is pretty powerful

PR #561, opened 2022-02-02
Add EnsOrAddress component

holic:
This quickly translates an address to an ENS name (where available) using [a highly cached API I built](https://ensideas.com/about).

The results look like this:
image

This could be optionally coupled with some lightweight frontend caching that stores the last ~day of queries in localStorage, but not strictly necessary - Cloudflare edge cache is very fast (~50ms for me).

I only swapped out the address for one page/component, but happy to do more if you like this approach!


Robinnnnn (approved):
you're the hero we don't deserve

PR #563, opened 2022-02-02
lowercase address before querying ENS name

holic:
Meant to do this in #561. This ensures a consistent cache key in SWR and avoids a redirect/round trip at the API.

Also went ahead and did the more defensive check for `data.address` in case the request fails or returns bad JSON for any reason.

### DIFF
--- a/src/components/EnsOrAddress.tsx
+++ b/src/components/EnsOrAddress.tsx
@@ -9,10 +9,11 @@ type Props = {
 
 export const EnsOrAddress = ({ address }: Props) => {
   const { data } = useSWR(
-    `https://api.ensideas.com/ens/resolve/${encodeURIComponent(address)}`,
+    `https://api.ensideas.com/ens/resolve/${encodeURIComponent(address.toLowerCase())}`,
     fetcher
   );
-  if (data) {
+
+  if (data?.address) {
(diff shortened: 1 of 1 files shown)


kaitoo1 (approved):

src/components/EnsOrAddress.tsx
```diff
@@ -9,10 +9,11 @@ type Props = {
 
 export const EnsOrAddress = ({ address }: Props) => {
   const { data } = useSWR(
-    `https://api.ensideas.com/ens/resolve/${encodeURIComponent(address)}`,
+    `https://api.ensideas.com/ens/resolve/${encodeURIComponent(address.toLowerCase())}`,
```
kaitoo1, line 12: any chance address is null or undefined? might not hurt to  address?.toLowerCase()

holic: `address` comes from props, which is typed! do we have any JS to worry about that may be passing in bad/untyped args?

kaitoo1: no js to worry about 👍
and yeah you're right, we should be good in this case. I brought it up bc we had an error before where we were calling toLowerCase() on something undefined so this gave me flashbacks lol but I think it was a special case. carry on!

PR #570, opened 2022-02-03
Generate OpenGraph images for user, collection, and NFT pages

holic:
image

Based on this spec:
https://www.notion.so/gallery-so/Eng-Social-Link-Previews-fa15977d33534738a9f2a52758954d5b
https://www.figma.com/file/JvnheQ7wnlLy5mVx1F14Xz/fun-stuff?node-id=126%3A2

Uses Vercel's CDN caching headers defined here:
https://vercel.com/docs/concepts/edge-network/caching
This may not be as performant as we'd want/expect, since Vercel flushes the cache every deploy. Might consider putting Cloudflare in front to get more life out of the cached images.


pages/[username]/[collectionId]/[nftId]/index.tsx
```diff
@@ -16,8 +18,34 @@ export default function NftDetailPage({ nftId }: NftDetailPageProps) {
...
 
-export const getServerSideProps: GetServerSideProps<NftDetailPageProps> = async ({ params }) => ({
-  props: {
-    nftId: params?.nftId ? (params.nftId as string) : undefined,
-  },
-});
+export const getServerSideProps: GetServerSideProps<NftDetailPageProps> = async ({ params }) => {
+  const username = params?.username ? (params.username as string) : undefined;
+  const nftId = params?.nftId ? (params.nftId as string) : undefined;
+  return {
```
Comment (Robinnnnn) [inline pages/[username]/[collectionId]/[nftId]/index.tsx:24]: could we temporarily disable this in prod?

we usually use [featureFlag.ts](https://github.com/gallery-so/gallery/blob/main/src/utils/featureFlag.ts) but it relies on `window`... is it possible to detect the origin in `getServerSideProps`?

Comment (Robinnnnn) [reply pages/[username]/[collectionId]/[nftId]/index.tsx:24]: ahh we can use `NEXT_PUBLIC_VERCEL_ENV`

pages/opengraph/collection/[collectionId].tsx
```diff
@@ -0,0 +1,42 @@
+import { useRouter } from 'next/router';
+import { OpenGraphPreview } from 'components/opengraph/OpenGraphPreview';
+import useCollectionById from 'hooks/api/collections/useCollectionById';
+
+export default function OpenGraphCollectionPage() {
```
Comment (Robinnnnn) [inline pages/opengraph/collection/[collectionId].tsx:5]: this setup is so freaking cool

Comment (Robinnnnn) [reply pages/opengraph/collection/[collectionId].tsx:5]: i thought we'd use raw html/css but the fact that it's a plain nextjs route is amazing

pages/opengraph/collection/[collectionId].tsx
```diff
@@ -0,0 +1,42 @@
...
+  const { query } = useRouter();
+  const collection = useCollectionById({ id: query.collectionId as string });
+
+  const width = parseInt(query.width as string) || 600;
+  const height = parseInt(query.height as string) || 300;
+
+  if (!collection) {
+    // TODO: render something nice?
+    throw new Error('no collection found');
```
Comment (Robinnnnn) [inline pages/opengraph/collection/[collectionId].tsx:14]: in these cases we should just fall back to the default image configured in `app`

pages/api/opengraph/image.ts
```diff
@@ -0,0 +1,91 @@
...
+    res.send(imageBuffer);
+  } catch (error: unknown) {
+    // TODO: log this to some error tracking service?
+    console.log('error while generating opengraph image', error);
+    res.json({
+      error: {
+        code: 'UNEXPECTED_ERROR',
+        message: (error as any).toString(),
+      },
+    });
```
Robinnnnn, line 85: what would be the behavior to the client (twitter) if this goes wrong? maybe we still `res.send` successfully with the default image?

pages/opengraph/collection/[collectionId].tsx
Comment (kaitoo1) [reply pages/opengraph/collection/[collectionId].tsx:14]: agreed, we can show what we currently show in our link previews

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
```diff
@@ -5,6 +5,7 @@ import Head from 'next/head';
 import CollectionGallery from './CollectionGallery';
 import useBackButton from 'hooks/useBackButton';
 import ActionText from 'components/core/ActionText/ActionText';
+import { baseUrl } from 'utils/baseUrl';
```
kaitoo1, line 8: nit: unused import?

src/scenes/NftDetailPage/NftDetailPage.tsx
```diff
@@ -13,6 +13,7 @@ import NftDetailText from './NftDetailText';
 import Head from 'next/head';
 import { useRouter } from 'next/router';
 import useBackButton from 'hooks/useBackButton';
+import { baseUrl } from 'utils/baseUrl';
```
kaitoo1, line 16: same, unused import?

pages/[username]/[collectionId]/index.tsx
```diff
@@ -28,9 +30,35 @@ export default function CollectionGallery({ collectionId, username }: Collection
...
+}) => {
+  const username = params?.username ? (params.username as string) : undefined;
+  const collectionId = params?.collectionId ? (params.collectionId as string) : undefined;
+  return {
+    props: {
+      username,
+      collectionId,
+      metaTags: collectionId
+        ? [
+            // TODO: fetch collection for better title?
```
Comment (kaitoo1) [inline pages/[username]/[collectionId]/index.tsx:42]: are these todos for this PR or the future? collection name would be cool for sure

pages/[username]/[collectionId]/[nftId]/index.tsx
Comment (kaitoo1) [reply pages/[username]/[collectionId]/[nftId]/index.tsx:24]: +1 for hiding on prod for now

pages/api/opengraph/image.ts
```diff
@@ -0,0 +1,91 @@
...
+    await page.waitForSelector('#opengraph-image');
+    const element = await page.$('#opengraph-image');
+
+    if (!element) {
+      throw new Error('No #opengraph-image element found');
+    }
+
+    const imageBuffer = await element.screenshot({ type: 'png' });
+
+    // s-maxage:
```
kaitoo1, line 62: thanks for the comments!

src/components/opengraph/OpenGraphPreview.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const OpenGraphPreview = ({ title, description, imageUrls }: Props) => (
+  <>
+    <div className="container">
+      <div className="gallery">
+        {imageUrls.map((url) => (
+          <div key={url} className="image" style={{ backgroundImage: `url("${url}")` }} />
+        ))}
+      </div>
+      <div className="byline">
+        <div className="username">{unescape(title)}</div>
```
kaitoo1, line 19: just a heads up that titles are optional for Collections so we might not always have a title here. 

that shouldn't affect too much, one place I could see that matter is the separator? not sure if we want to show the separator if we don't have a title. 
that's more of a design decision though so I can call it out on discord.

kaitoo1:
awesome PR @holic !
the preview image generation + open graph component were super cool to see, TIL so thank you for that 👍
and nice work figuring out the meta tag issues

it looks like our questions are mostly around error cases and what that would look like on twitter for example.
otherwise looks great, Im excited for this one!

pages/[username]/[collectionId]/index.tsx
Comment (holic) [reply pages/[username]/[collectionId]/index.tsx:42]: > are these todos for this PR or the future? collection name would be cool for sure

Mostly for future! The reason I can't do this right now is we're inside `getServerSideProps`, which doesn't do any backend queries yet. We'd need to query for the collection name if we wanted to use that in the title.

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
holic: might consider wiring up CI and ensuring lint passes on the PR before merging so ya'll don't have to be the human linters!

pages/api/opengraph/image.ts
holic: added a fallback above!

src/components/opengraph/OpenGraphPreview.tsx
holic: since we're blocking on a design pass, I'll leave this for the design work

holic:
- Added a fallback redirect to the current opengraph preview image when the image generation attempt fails
- Put the new opengraph tags behind a feature flag

src/components/core/enums.ts
```diff
@@ -21,4 +21,5 @@ export enum DisplayLayout {
 export enum FeatureFlag {
   GENERAL_MEMBERSHIP_MINT = 'GENERAL_MEMBERSHIP_MINT',
   SINGLE_COLLECTION = 'SINGLE_COLLECTION',
+  OPENGRAPH_IMAGES = 'OPENGRAPH_IMAGES',
 }
```
holic, line 25: Feels a little clunky that I have to import this enum separately from the `isFeatureFlagEnabled`, might be worth either moving this enum to `featureFlags.ts` or allowing the `isFeatureFlagEnabled` to accept a string key and lean on TS to ensure that it's a valid enum key.

Robinnnnn: yeahh agreed, string key probably cleanest

pages/api/opengraph/image.ts
Robinnnnn: +1

pages/api/opengraph/image.ts
Robinnnnn: incredible!

src/utils/isProduction.ts
```diff
@@ -1,3 +1,5 @@
 export default function isProduction() {
-  return typeof window !== 'undefined' && window.location.origin === 'https://gallery.so';
+  return (
+    process.env.NEXT_PUBLIC_VERCEL_ENV === 'production' || process.env.NODE_ENV === 'production'
```
Robinnnnn, line 3: hmm doesn't a vercel deploy mark `NODE_ENV` as `production`? even if it's to dev?

Robinnnnn: at build time

src/utils/openGraphMetaTags.ts
```diff
@@ -0,0 +1,34 @@
+import { MetaTagProps } from 'pages/_app';
+import { baseUrl } from './baseUrl';
+
+type Params = {
+  title: string;
+  previewPath: string;
+};
+
+export const openGraphMetaTags = ({
```
Robinnnnn, line 9: CLEAN

Robinnnnn (approved):

src/utils/isProduction.ts
holic: you're right! fixed https://github.com/gallery-so/gallery/pull/570/commits/df6763520534ed00628d6623952b45c14c2cff03

Robinnnnn (approved):

pages/[username]/[collectionId]/index.tsx
Comment (kaitoo1) [reply pages/[username]/[collectionId]/index.tsx]: sounds gooddd

PR #571, opened 2022-02-03
Single Collection Follow up 2

kaitoo1:
- track historyStack in GalleryNavigationProvider so that we can easily access the previous page path without using storage
- use useHistoryStack to not use back() on Collection Page if we came from the Edit Collection Page
- replace existing usage of storage to get previous page with useHistoryStack
- put `isHovering` state back on UserGalleryCollection so that we can close the drop down on mouse out
- wrap long collection name for mobile
- track share event

### DIFF
--- a/src/contexts/navigation/GalleryNavigationProvider.tsx
+++ b/src/contexts/navigation/GalleryNavigationProvider.tsx
@@ -98,10 +98,10 @@ export function GalleryNavigationProvider({ children }: Props) {
       };
 
       setHistoryStackLength(window.history.state.idx ?? 0);
-      setHistoryStack((stack) => {
-        stack.pop();
-        return stack;
-      });
+      // setHistoryStack((stack) => {
+      //   stack.pop();
+      //   return stack;
+      // });
     }
 
     window.addEventListener('popstate', handlePopState);
--- a/src/contexts/navigation/GalleryNavigationProvider.tsx
+++ b/src/contexts/navigation/GalleryNavigationProvider.tsx
@@ -23,6 +23,7 @@ type Props = {
 export function GalleryNavigationProvider({ children }: Props) {
   const [historyStackLength, setHistoryStackLength] = useState(0);
   const { asPath } = useRouter();
+  // This is a flattened list of all the paths the user has visited. Nothing gets popped off, unlike the browser history.
   const [historyStack, setHistoryStack] = useState<string[]>([]);
 
   useEffect(() => {
--- a/src/contexts/navigation/GalleryNavigationProvider.tsx
+++ b/src/contexts/navigation/GalleryNavigationProvider.tsx
@@ -1,7 +1,8 @@
 import { createContext, ReactNode, useContext, useEffect, useMemo, useState } from 'react';
 import { FADE_TIME_MS } from 'components/FadeTransitioner/FadeTransitioner';
+import { useRouter } from 'next/router';
 
-type GalleryNavigationContextType = { historyStackLength: number };
+type GalleryNavigationContextType = { historyStackLength: number; historyStack: string[] };
 
 const GalleryNavigationContext = createContext<GalleryNavigationContextType | undefined>(undefined);
 
@@ -21,6 +22,12 @@ type Props = {
 
 export function GalleryNavigationProvider({ children }: Props) {
   const [historyStackLength, setHistoryStackLength] = useState(0);
+  const { asPath } = useRouter();
+  const [historyStack, setHistoryStack] = useState<string[]>([]);
+
+  useEffect(() => {
+    setHistoryStack((stack) => [...stack, asPath]);
+  }, [asPath]);
 
   useEffect(() => {
     const originalPushState = window.history.pushState;
@@ -91,6 +98,10 @@ export function GalleryNavigationProvider({ children }: Props) {
       };
 
       setHistoryStackLength(window.history.state.idx ?? 0);
+      setHistoryStack((stack) => {
+        stack.pop();
+        return stack;
+      });
     }
 
     window.addEventListener('popstate', handlePopState);
@@ -104,8 +115,9 @@ export function GalleryNavigationProvider({ children }: Props) {
   const value = useMemo<GalleryNavigationContextType>(
     () => ({
       historyStackLength,
+      historyStack,
     }),
-    [historyStackLength]
+    [historyStackLength, historyStack]
   );
 
   return (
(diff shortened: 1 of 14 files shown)


src/contexts/navigation/GalleryNavigationProvider.tsx
```diff
@@ -21,6 +22,12 @@ type Props = {
 
 export function GalleryNavigationProvider({ children }: Props) {
   const [historyStackLength, setHistoryStackLength] = useState(0);
+  const { asPath } = useRouter();
+  const [historyStack, setHistoryStack] = useState<string[]>([]);
+
+  useEffect(() => {
+    setHistoryStack((stack) => [...stack, asPath]);
+  }, [asPath]);
```
Robinnnnn, line 30: will this keep pushing to the stack even when clicking "back"? should we only add to the stack in the `pushState` function below?

src/scenes/NftDetailPage/NftDetailPage.tsx
```diff
@@ -13,6 +13,7 @@ import NftDetailText from './NftDetailText';
 import Head from 'next/head';
 import { useRouter } from 'next/router';
 import useBackButton from 'hooks/useBackButton';
+import { useHistoryStack } from 'contexts/navigation/GalleryNavigationProvider';
```
Robinnnnn, line 16: seems like this import isnt' used

src/scenes/CollectionGalleryPage/CollectionGalleryHeader.tsx
```diff
@@ -43,7 +42,11 @@ function CollectionGalleryHeader({ collection, mobileLayout, setMobileLayout }:
     [collection.collectors_note]
   );
 
-  const authenticatedUserIsOnTheirOwnPage = username.toLowerCase() === user?.username.toLowerCase();
+  const handleShareClick = useCallback(() => {
+    Mixpanel.track('Share Collection', { path: `/${username}/${collection.id}` });
+  }, [collection.id, username]);
```
Robinnnnn, line 47: 😍

src/hooks/useBackButton.tsx
```diff
@@ -17,7 +22,10 @@ export default function useBackButton({ username }: Props) {
         return;
       }
 
-      if (canGoBack) {
+      // If the user has a different page in their history such as the Edit Collection Page, don't use back()
+      const dontGoBack = historyStack[historyStack.length - 2]?.includes('/edit?collectionId=');
```
Robinnnnn, line 26: nit: would prefer this variable defined outside of the callback, similar to `canGoBack`. the reason is because this forrces `historyStack` to be a dependency of this callback, which is an array that's almost guaranteed to change more frequently than a simple true/false.

e.g.
```js
const previousRoute = historyStack[historyStack.length - 2]
const backDisabled = previousRoute?.includes('/edit?collectionId=');

const handleBackClick = useCallback(() => {
  ...
  if (!canGoBack || backDisabled) { <----- changed the logic a bit, not sure if better or worse
    void replace(`/${username}`);
    return;
  }
  back();
}, [...backDisabled...])
```

src/hooks/useBackButton.tsx
```diff
@@ -29,7 +37,7 @@ export default function useBackButton({ username }: Props) {
         void replace(`/${username}`);
```
Robinnnnn, line 37: this should be controlled by the arguments coming into this hook right? i.e. if you're on the NFT Detail Page, and you came from the Collection Page, the back button should take you to your collection page (not user profile)

src/contexts/navigation/GalleryNavigationProvider.tsx
```diff
@@ -21,6 +22,12 @@ type Props = {
 
 export function GalleryNavigationProvider({ children }: Props) {
   const [historyStackLength, setHistoryStackLength] = useState(0);
```
Robinnnnn, line 24: now that we have `historyStack`, i think we can toss `historyStackLength`, since it's essentially a redundant piece of state.

down below we use `historyStackLength` to determine `canGoBack`:
image

but we can just update this to:
```
export function useCanGoBack() {
  const { historyStack } = useGalleryNavigationContext();

  return historyStack.length > 0;
}
```

src/contexts/navigation/GalleryNavigationProvider.tsx
Robinnnnn: i think we should remove this effect and replace with two things:
1. add `setHistoryStack((stack) => [...stack, asPath]);` in the `pushState` function below
2. add `setHistoryStack((stack) => stack.slice(0, -1));` in the `popState` function below

kaitoo1: Definitely, I started down this road at first with the intention of ideally tracking the browser history 1:1 but it kept getting out of sync if I navigated around and clicked the back and forward buttons a bunch. When it got out of sync, the "Back to Gallery" button would navigate me to all sorts of weird places.

Then I thought that since all I wanted to do was know what url I was on previously, I could instead just have a linear list of urls I've been to. 
So lets say I navigate A -> B then hit the back button.
Browser history would just show [A] because it added B when I went to B, but then popped it when I clicked back.
My list would show [A,B,A] and that's just a direct, dumb representation of where you've been. 

My reasoning was that
- This list isn't going to be used to navigate forwards or backwards in history, but instead just to see what pages the user came from.
- Since we're not trying to recreate the browser history 1:1, this has less risk of getting out of sync and causing weird behavior

I can rename this and add a comment as well to clarify if the above makes sense to you

src/hooks/useBackButton.tsx
kaitoo1: oh yeah totally!

kaitoo1: done ✅

src/scenes/NftDetailPage/NftDetailPage.tsx
kaitoo1: removed ✅

src/hooks/useBackButton.tsx
kaitoo1: I believe this is ok atm because this else block is a fallback only if canGoBack is falsy or back is disabled (came from edit route). When clicking this back button from the NFT Detail Page after they arrived from the Collection Page, `canGoBack` is true so `back()` will take the user to the correct page.

If they arrive at the NFT Detail Page directly so they have no history, then they this will be called and they will be taken to the Gallery Page.

PR #572, opened 2022-02-03
bug repro: tags in page-level <Head> not rendering to server-sent HTML

holic:


holic:
Found it! It's because we're deferring rendering on the server in `<SafeHydrate>`, likely to avoid errors with using suspense on the server.

I can see a few options to resolve this:
1. upgrade to react 18,  configure next with experimental `concurrentFeatures: true` flag
2. turn off suspense for swr and move all usages to loading indicators
3. hack around this by loading the data needed inside `getServerSideProps` for each page, pass in via custom set of props that can append the meta tags in `_app` (this would increase load time for pages and be kind of a heavy lift)

1 seems like the lightest lift, but has some unknowns (e.g. will it actually work? will react 18 break any assumptions in the app or libs we depend on)

PR #574, opened 2022-02-04
specify which elements we allow in markdown

kaitoo1:
Changes:
This PR updates the markdown component to specify which elements we allow via markdown. 

For now we will allow users to use the following elements in their bio and collection+nft collectors notes: `['a', 'strong', 'em', 'ol', 'ul', 'li', 'p']`

### DIFF
--- a/src/components/core/Markdown/Markdown.tsx
+++ b/src/components/core/Markdown/Markdown.tsx
@@ -19,6 +19,8 @@ export default function Markdown({ text }: Props) {
             <>{children}</>
           ),
       }}
+      allowedElements={['a', 'strong', 'em', 'ol', 'ul', 'li', 'p']}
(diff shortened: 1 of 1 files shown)


src/components/core/Markdown/Markdown.tsx
```diff
@@ -19,6 +19,8 @@ export default function Markdown({ text }: Props) {
             <>{children}</>
           ),
       }}
+      allowedElements={['a', 'strong', 'em', 'ol', 'ul', 'li', 'p']}
+      unwrapDisallowed
```
kaitoo1, line 23: this allows disallowed markdown to be rendered as text instead of ignored completely

Robinnnnn (approved):

PR #586, opened 2022-02-04
Set up support for meta tags rendered via getServerSideProps

holic:
I tried moving around `<SafeHydrate>` with no luck and also tried to upgrade to React 18 with no luck. This is the other idea I had in mind to get meta tags rendering in the server-sent HTML on a per-page level.


tbezman (approved):
Looks good! Where did the SafeHydrate / React 18 stuff fall apart? Gonna need to start thinking about it for the long term approach

Robinnnnn:
LGTM, but should we have a dev-only flag for the username route? if we merge + deploy this we'll end up with no link preview for users right

Robinnnnn (approved):

holic:
> Looks good! Where did the SafeHydrate / React 18 stuff fall apart? Gonna need to start thinking about it for the long term approach

The current component tree looks something like this
```
<App> (_app.tsx)
  <SafeHydrate>
    <SWRConfig> (in AppProvider.tsx)
      <NextPageComponent> (pages/*.tsx)
        <GalleryRoute>
          ...
```

I tried to move `<SafeHydrate>` down into `<GalleryRoute>` to get it out of the way of `<NextPageComponent>`:

```
<App> (_app.tsx)
  <SWRConfig> (in AppProvider.tsx)
    <NextPageComponent> (pages/*.tsx)
      <GalleryRoute>
        <SafeHydrate>
          ...
```

But this means that `<SWRConfig>` (that turns on suspense) now lives outside of `<SafeHydrate>`, which causes the "can't do suspense with server side rendering" errors.

Another approach could be moving the `<AppProvider>` down into `<GalleryRoute>`, but there's a handful of pages that do fetching, etc. before rendering `<GalleryRoute>` so we'd need to rethink/refactor those a bit.

For React 18, the page seemed to get stuck in a loading state and the server process spit out some errors (with no stacktraces) with something along the lines of `document is not defined`. Again, probably more refactoring needed - I think a bunch of the frontend currently expects/assumes it's only rendering on client side.

holic:
> LGTM, but should we have a dev-only flag for the username route? if we merge + deploy this we'll end up with no link preview for users right

I wasn't intending on merging this, just wanted to show the diff and get feedback before bundling it with the social previews PR.

Robinnnnn:
> I wasn't intending on merging this, just wanted to show the diff and get feedback before bundling it with the social previews PR.

amazing, thanks for doing all the exploratory work here – super down for this as a short term solution

holic:
@Robinnnnn I went ahead and removed the meta tag modifications that overlap with the other PR and reduced this PR to just adding support for meta tags via getServerSideProps. Should be safe to merge this as-is and add rebase+add meta tags in my other PR.

PR #587, opened 2022-02-06
Fix fade transition with suspense jank

tbezman:
There was some pretty gross jank when transitioning to a route which suspended while waiting for data.

The root cause here was the `FullScreenLoader` wasn't getting the same transition styles as the actual route, so it just looked really sudden / jarring.

**Before / After**

On the right, you're not seeing the fallback since the API resolves faster than the transition finishes. I tested it out with a component which suspends forever and the functionality looks right.

https://user-images.githubusercontent.com/6754223/152666156-01b6d6fb-87f0-449b-91df-66c036103f2f.mp4


Robinnnnn (approved):
🔥🔥🔥🔥🔥🔥

PR #591, opened 2022-02-09
Add collector's note functionality

connorrothschild:
# Collector's Note Functionality

This PR introduces the ability for users to add collector's notes to their NFTs. It lives behind a feature flag enabled on dev and disabled on prod. [You can add notes to your assets, browse, etc. on the preview](https://gallery-git-connor-collectors-note-rebase-gallery-so.vercel.app/). Any testing related to visual changes to assets, overall feel of note experience, etc. would be helpful. 

## At a glance

Assets with collector's notes look like this: 

image

The asset container is designed such that it still takes up the vast majority of the vertical space in the screen; the collector's note, other than the first line, is hidden "beyond the fold."

This means that long collector's notes do not force the asset page to resize in any way, but instead the entire note (other than the first line) appear **under** the main content. To access that lengthy note, the page visitor will need to scroll down:

image

Here are four views of a NFT according to three different users:

1. User visits NFT *they own* that **has a note**
2. User visits NFT *they own* that **does not have a note**
3. User visits NFT *they do not own* that **has a note**
4. User visits NFT *they do not own* that **does not have a note**

<div style="display: flex; flex-wrap: wrap;">
image
image
image
image
</div>

## Visual differences from current sites

Due to a reconfiguration of how elements were positioned, this PR has a slight effect on the visual appearance of assets, whether they have a note or not. Here is an example of before (left) and after (right), for an asset *without* a note:

<div style="display: flex; flex-wrap: wrap;">
image
image
</div>

We had [formerly](https://github.com/gallery-so/gallery/pull/591/files#diff-3588cdcee5c92a9ff85f423ed5717ad78e40cf9311a32cdf6469c5bfdc64476eL109-L111) (e.g. in the left image) been applying a fixed width and height to the *asset container* on desktop to be equal to the asset height:

```js
 @media only screen and ${breakpoints.desktop} {
    width: ${({ maxHeight }) => maxHeight}px;
    height: ${({ maxHeight }) => maxHeight}px;
}
```

It is unclear why we were doing this (and it also created some dead horizontal space because this 1/1 aspect ratio was applied to the container, not the asset itself); it needed to be removed to allow for collector's notes. 

[Now](https://github.com/gallery-so/gallery/pull/591/files#diff-3588cdcee5c92a9ff85f423ed5717ad78e40cf9311a32cdf6469c5bfdc64476eR129-R132), we size an asset according to its width (`100%`) which then sets the height to whatever it needs to be to fit within the parent container. We apply a max height based on the 1. viewport height (`85vh`), 2. footer height (`GLOBAL_FOOTER_HEIGHT`), and 3. some additional spacing so that the very top of the collector's note is visible (`46px`, which is roughly equal to the line height of "Collector's Note", the first line of text, and padding):

```js
@media only screen and ${breakpoints.tablet} {
    width: 100%;
    ${({ footerHeight }) => `max-height: calc(85vh - 46px - ${footerHeight}px)`};
}
```

On vertically small screens, this means that assets no longer exceed the page bounds but instead stay within ~85% of the viewport. Before/after:

<div style="display: flex; flex-wrap: wrap;">
image
image
</div>

I think this is preferable but feel free to push back.

## Dynamic input size

The size of the textarea is based on the user's input. As the user inputs more, the textarea height will progressively increase, and the page will scroll down:

https://user-images.githubusercontent.com/13339581/153255776-c2cd4c7a-91de-4f28-97eb-3c923dc70aae.mov

### Reducing textarea height

Making the `textarea` height dynamic to user input was easy when the user *added* content; it was difficult to reduce the size of the `textarea` on delete. (This is due to the note height being equal to the element's `scrollHeight`, which does not decrease based on input.) This means that if—for some reason—a user had a long note and deleted some of it, there would be dead space in the editor:

https://user-images.githubusercontent.com/13339581/153256183-b9c0574a-d413-4f3b-b9fd-d58e6f2e1bd7.mov

This is fixed on page reload and flagged with a TODO. We currently *do* reset the height if the user deletes **all content**—but this is an imperfect solution.

### Initial height of editor

For long notes, when a user fresh-reloads or newly visits a page, they have to edit the textarea at least once to trigger its full height resize. This is because we resize on the textarea `onChange` event. So, if a user has a long collector's note and want to edit it (after visiting on a fresh page reload), it will initially be ~150px until they change a letter:

https://user-images.githubusercontent.com/13339581/153256539-476459d6-3ba1-4c2e-85a7-bf5d8734cd50.mov

This is flagged with a TODO.


src/hooks/api/nfts/useUpdateNft.ts
```diff
@@ -0,0 +1,26 @@
...
+      await updateNft<UpdateNftInfoResponse, UpdateNftInfoRequest>(
+        `/nfts/update`,
+        'update nft info',
+        {
+          id: nftId,
+          collectors_note: collectorsNote,
+        }
+      );
+
+      await mutate(getNftCacheKey({ id: nftId }));
```
Robinnnnn, line 22: 😍

src/scenes/NftDetailPage/NftDetailAsset.tsx
```diff
@@ -53,6 +54,8 @@ function NftDetailAssetComponent({ nft, maxHeight }: NftDetailAssetComponentProp
 
 type Props = {
   nft: Nft;
+  userOwnsAsset: boolean;
```
Robinnnnn, line 57: supernit: would prefer being a bit more explicit here - technically all assets are owned by users. maybe something like `authenticatedUserOwnsAsset`

src/scenes/NftDetailPage/NftDetailAsset.tsx
```diff
@@ -75,40 +78,58 @@ function NftDetailAsset({ nft }: Props) {
...
   ${({ shouldEnforceSquareAspectRatio }) =>
     shouldEnforceSquareAspectRatio ? 'aspect-ratio: 1' : ''};
 
-  @media only screen and ${breakpoints.desktop} {
-    width: ${({ maxHeight }) => maxHeight}px;
-    height: ${({ maxHeight }) => maxHeight}px;
+  @media only screen and ${breakpoints.tablet} {
+    width: 100%;
+    ${({ isCollectorsNoteEnabled, footerHeight }) =>
+      isCollectorsNoteEnabled ? `max-height: calc(85vh - 46px - ${footerHeight}px)` : ''};
```
Robinnnnn, line 132: curious, where did the `46px` come from?

src/scenes/NftDetailPage/NftDetailNote.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+  const scrollDown = () => {
+    if (collectorsNoteRef.current) {
+      collectorsNoteRef.current.scrollIntoView({
+        block: 'end',
+        inline: 'nearest',
+        behavior: 'smooth',
+      });
+    }
+  };
```
Robinnnnn, line 46: let's lock this in a `useCallback`!

src/scenes/NftDetailPage/NftDetailNote.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+  const handleEditCollectorsNote = useCallback(() => {
+    setIsEditing(true);
+
+    // TODO Expand note to full height first time it is opened
+    // Currently user has to type first to expand it fully
```
Robinnnnn, line 52: is this still needed?

src/scenes/NftDetailPage/NftDetailNote.tsx
```diff
@@ -0,0 +1,241 @@
...
+  // If the user hits cmd + ctrl enter, submit the note
+  const handleKeyDown = useCallback(
+    async (event) => {
+      if (event.key === 'Enter' && event.metaKey) {
+        event.preventDefault();
+        await handleSubmitCollectorsNote();
+      }
+    },
+    [handleSubmitCollectorsNote]
+  );
```
Robinnnnn, line 90: 😍

src/scenes/NftDetailPage/NftDetailNote.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const handleNoteChange = useCallback((event: React.ChangeEvent<HTMLTextAreaElement>) => {
+    setCollectorsNote(event.target?.value);
+    setNoteHeight(event.target?.scrollHeight);
+
+    // On clear, reset note height (textarea scrollHeight does not decrease on its own, it only increases)
+    // So we use this hard reset if the user deletes all content. Could have more elegant solution
+    // TODO Reduce size to text content on any delete
+    if (event.target?.value === '') {
+      setNoteHeight(MIN_NOTE_HEIGHT);
+    }
```
Robinnnnn, line 101: love this solution

src/scenes/NftDetailPage/NftDetailNote.tsx
```diff
@@ -0,0 +1,241 @@
...
+  }, []);
+
+  return (
+    <StyledContainer ref={collectorsNoteRef} tabIndex={0} onKeyDown={handleKeyDown}>
+      <Spacer height={18} />
+
+      <StyledTitleAndButtonContainer>
+        {(hasCollectorsNote || isEditing) && <BodyRegular>Collector&rsquo;s Note</BodyRegular>}
+        {userOwnsAsset && !hasCollectorsNote && !isEditing && (
+          <TextButton text={`+ Add Collector's Note`} onClick={handleEditCollectorsNote} />
```
Robinnnnn, line 117: supernit: any reason to use back ticks here? can just do `text="+ Add Collector's Note"`

src/scenes/NftDetailPage/NftDetailPage.tsx
```diff
@@ -28,6 +29,9 @@ function NftDetailPage({ nftId }: Props) {
   const isFromCollectionPage =
     globalThis?.sessionStorage?.getItem('prevPage') === `/${username}/${collectionId}`;
 
+  const authenticatedUsername = useAuthenticatedUsername();
+  const userOwnsAsset = authenticatedUsername === username;
```
Robinnnnn, line 33: ah so this isn't documented well, but unfortunately `useAuthenticatedUsername` will throw an error if you're not logged in. this is bc it calls `useAuthenticatedUser` under the hood. i tried going to an NFT detail page while logged out (in incognito), and was indeed able to trigger the error.

image

there is a more lax hook you can use from the same file: `usePossiblyAuthenticatedUser`. so you can update like this:

```js
  const authenticatedUser = usePossiblyAuthenticatedUser(); // returns a user object OR or `undefined` if not logged in
  const userOwnsAsset = authenticatedUser?.username === username;
```

src/scenes/NftDetailPage/NftDetailAsset.tsx
connorrothschild: This is roughly equal to the height of the "Collector's Note" title + the first line of text + padding. It was kind of some back-and-forth editing to find the magic number. 

On second thought, we might only want to subtract the 46px **if the asset has a note** or **if the user owns the current asset** (e.g. has the `Add Collector's Note` button).

src/scenes/NftDetailPage/NftDetailNote.tsx
connorrothschild: Yes, this is only relevant when the user reloads the page or visits an existing long note and tries to edit it. (I can clarify in the TODO comment as well.) There's a video of the problem this in the edited PR under "Remaining issues"

Robinnnnn:
first of all, this is a **phenomenal** PR. thank you for addressing comments and laying everything out in such detail 🔥🔥🔥

next steps:
- since this is safe behind a feature flag, let's just get this merged
- the yarn.lock had another 1-liner conflict which i just fixed directly in the github editor lol 

re: remaining issues:
- **reducing textarea height**: this is a super small issue and shouldn't block release. i say we add this as a task on the task board under P4 priority (can add to "triaged")
- **initial height of editor**: i believe this can be fixed using a hack with `useRef` where we place a ref on the textarea, and once we detect that it mounted we trigger `onChange`. again def not a blocker for this PR but we can find some time to play around with it

kaitoo1 (changes requested):

kaitoo1:
Awesome PR!
Thanks for taking the time to consider all the possible layouts and working with the rest of the team to figure out the best way to display the note.
I took it for a spin on the preview url and overall the note editor and viewing functionality looks great 🔥 Adding and editing a note was smooth, and details like error handling, responsiveness, and auto-scroll when editing makes it feel polished.

I did notice a few things that I think are important to address before we merge this PR. I tried to add as much information as possible so it's easier to repro and test.

**regression of how we display NFT media assets**
I think these are all related - most likely related to changing the width and height of `NftDetailAsset` ?

iframes are smaller
https://gallery-git-connor-collectors-note-rebase-gallery-so.vercel.app/ayybee_gallery/22TbT2v8aawTYpzMS3nNr4vHoM9/22Rir2KWF7WJapZAMzAz5FRQN6x
CleanShot 2022-02-10 at 19 15 22

images keep growing with window size (no max size)
https://gallery-git-connor-collectors-note-rebase-gallery-so.vercel.app/kaito/1yyRVZsVsiuXPT7RBGpPVZ8J7Ap/1yyROx12WjUPNktvapZfsmBcxD4
CleanShot 2022-02-10 at 19 16 17

horizontal images get stretched into a square
https://gallery-git-connor-collectors-note-rebase-gallery-so.vercel.app/pixelpete/23U56ohE49qoV9nACUFXwXpZip9/23U38IwBVWvuZIfIqcu1NBMYFuv
CleanShot 2022-02-10 at 21 13 18


svgs + glb models don't appear (I think they're in the dom but are missing dimensions)
https://gallery-git-connor-collectors-note-rebase-gallery-so.vercel.app/kaito/24Go5g7DtmqtkebEf9TL9QK5XkU/1yyROru2wMZ9LekSqV61MVlExWJ
CleanShot 2022-02-10 at 19 11 06
CleanShot 2022-02-10 at 19 09 34

The next one might be related as well 

**regression: Shimmer Loader doesn't appear until right before image is loaded**

video: (left window is prod, right window is preview link)
https://drive.google.com/file/d/13mwqXwdKZ5tHIhaeLhUFROpVK6iXAxJ0/view?usp=sharing

**User can trigger the editor to appear on other people's NFTs, and their own while signed out**
repro steps:
- go to any NFT with a collectors note you should not be able to edit (other people's NFT, or your own while signed out)
- double click the note. the editor will appear
- there isn't a save button but the keyboard shortcut (command+shift/ctrl+enter ) will submit and call the update endpoint. 

I left comments on the NftDetailNote component with suggestions on how we could simplify its logic to help with situations like this 
video:
https://drive.google.com/file/d/1uHnqRAgGRPMwusqKV1qS3PBfP-TeVhpc/view?usp=sharing

**The note is visible behind the footer**
Screen Shot 2022-02-10 at 21 16 22

connorrothschild:
@kaitoo1 Thank you for such a comprehensive review 🙏 all visual changes are now accounted for, to the best of my knowledge. It was mostly an issue with `shouldEnforceSquareAspectRatio` and applying `max-height`/`max-width` throughout. Early on in pursuing the note, I had taken out some CSS that ended up being very important (its since been readded).

I've tested all of the examples you linked to, as well as some others I thought would be relevant. Here's toggling between prod/current, for three different screen sizes:

https://user-images.githubusercontent.com/13339581/153492176-39fb8ae5-0da4-4dae-a88c-5042465c4f85.mov

https://user-images.githubusercontent.com/13339581/153492392-b2701601-dc7f-4af5-839e-c2a22c679acc.mov

https://user-images.githubusercontent.com/13339581/153493522-9dbc17da-4616-4406-a93a-59dcc6d74273.mov

We might play around with design of the mobile note (slight border under the note? something to that effect) but now the layout is near-identical to prod on all screen sizes, I believe.

---

I'm going to take a quick break and then look into the more structural comments re: 1. where the note component lives and 2. putting logic in parent/children to simplify the components

connorrothschild:
@kaitoo1 Some updates re: your great feedback!
 
[This commit](https://github.com/gallery-so/gallery/pull/591/commits/1dc5ad8d038e8a265addfcaab178f1d2a4ede38b) moves `NftDetailNote` one component up, into `NftDetailPage` rather than `NftDetailAsset`

[This commit](https://github.com/gallery-so/gallery/pull/591/commits/0f7d1480c0468775b8c15c4b4461d6745714ca4c) is a somewhat involved refactor that splits up `NftDetailNote` into `NoteViewer` and `NoteEditor` components, rendering each depending on whether the logged in user owns the asset. It also restructures the markup within the editor so it is rendered based on the state more intentionally (e.g. grouping together all elements that are rendered if the user is editing, has a note, etc).

It felt reallllyyyy good to rewrite the code in this way, it feels much safer etc. Good advice! (And let me know if you would change anything else; this was in the direction you were requesting but might still be improvable!)

kaitoo1 (approved):

kaitoo1:
@connorrothschild  
thanks so much for addressing all the feedback! the changes look great and the `NftDetailNote` component is already a lot easier to navigate and wrap my head around. love the `NoteViewer` and `NoteEditor` components. 

I'm glad you enjoyed restructuring it. the same approach can be applied to any component so I hope that helps in the future too. It'll also help us avoid any gotchas moving forward as we keep building on top of this feature.

Thanks for the videos as well, looks like all the nfts are displayed as expected too now 🙌

LGTM nice work! Collectors notes are going to add a whole new dimension to this page, I'm excited to release it

I believe @Robinnnnn is compiling some follow up polish items but those can be addressed as follow up items so this PR is good to merge

Robinnnnn:
🥇 🥇 🥇 🥇

PR #596, opened 2022-02-10
Update allowlist

Robinnnnn:
- add addresses
- use `Set` instead of array
- prepare to use external `.json` file

### DIFF
--- a/src/scenes/MembershipMintPage/GeneralCardAllowlist.ts
+++ b/src/scenes/MembershipMintPage/GeneralCardAllowlist.ts
@@ -1,6 +1,6 @@
 import isProduction from 'utils/isProduction';
 
-export const ALLOWLIST_DEV = [];
+const ALLOWLIST_DEV: string[] = [];
 
 const ALLOW_LIST_LEGENDA = [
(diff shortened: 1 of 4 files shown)


src/scenes/MembershipMintPage/GeneralCardAllowlist.ts
```diff
@@ -2836,12 +2836,18 @@ const ALLOW_LIST_CUSTOM = [
...
-  ...ALLOW_LIST_LEGENDA,
-  ...ALLOW_LIST_POOLSUITE,
-  ...ALLOW_LIST_CUSTOM,
-].map((s) => s.toLowerCase());
+const ALLOWLIST_PROD = [...ALLOW_LIST_LEGENDA, ...ALLOW_LIST_POOLSUITE, ...ALLOW_LIST_CUSTOM].map(
+  (s) => s.toLowerCase()
+);
 
-export const getAllowlist = () => (isProduction() ? ALLOWLIST_PROD : ALLOWLIST_DEV);
+export const getLocalAllowlist = () => new Set(isProduction() ? ALLOWLIST_PROD : ALLOWLIST_DEV);
```
Robinnnnn, line 2853: converting to `Set` here for better performance

PR #604, opened 2022-02-16
Replace instances of overflow: scroll with overflow: auto

connorrothschild:
This PR removes extraneous scrollbars by replacing `overflow: scroll` with `overflow: auto`. It affects three elements:

### 1. Sidebar on collection editing page (before/after):

<div style="display: flex;">

</div>

### 2. Staging area

<div style="display: flex;">

</div>

### 3. NFT detail text

<div style="display: flex;">

</div>

---

These scrollbars will still appear if **there is overflow**, but in events where content does not overflow no scrollbars appear.

For those interested, us Mac OSX were likely being tricked because trackpads automatically hide scrollbars. This is also why plugging in an external mouse would cause the scrollbars to reappear: https://kilianvalkhof.com/2021/css-html/you-want-overflow-auto-not-overflow-scroll/

### DIFF
--- a/src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
@@ -129,7 +129,7 @@ const StyledStagingArea = styled.div`
 
   padding: 100px 80px;
 
-  overflow: scroll;
+  overflow: auto;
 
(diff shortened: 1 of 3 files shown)


Robinnnnn (approved):

kaitoo1:
good find!

kaitoo1 (approved):

PR #610, opened 2022-02-19
Adjust styling of NFT preview text on small screens

connorrothschild:
This PR adjusts styling of asset preview text on small screens. It fixes the bug where collection titles/NFT names were cut off on mobile/small screens. Before/after:

<div style="display: flex;">
image
image
</div>

---

On screens under our `tablet` breakpoint, we 
1. reduce font size and line height, 
2. hide overflow, and 
3. apply a `text-overflow` property of `ellipsis` which leads to the `...` truncation when overflow begins. 

We use `-webkit-line-clamp` so that the NFT title will be truncated after one line of text, and the collection title will begin truncation after 2 lines of text (as you can see in the screenshot above). We can adjust these numbers, although generally 3 lines of text is the upper limit if we want to preserve the current design of the Nft preview text only taking up a max of ~56px.

You can read more about line clamping overflow [here](https://css-tricks.com/almanac/properties/l/line-clamp/). It has [full browser support](https://caniuse.com/?search=line-clamp) other than IE.

Also, I added a slight `text-shadow` to preview text, no matter the screen size. This makes titles more readable, especially on white assets, and makes the text feel elevated from the background:

<div style="display: flex;">
image
image
</div>

Let me know if you think we should reduce the intensity of the shadow!

---

## Note on browser support 

I've verified that this looks good on desktop Chrome, FF, and Safari. There is a difference in how Chrome and FF render truncated text though -- Chrome, by default, will only show the `...` if there is a break `within` a word; whereas FF will render it no matter what (even breaking after a word, e.g. a space). Firefox/Chrome:

<div style="display: flex;">
image
image
</div>

Of these two, the first (Firefox) seems preferable. I don't like that Chrome ends the string with no visual representation of truncation, so it just looks like the title is wrong.

To mitigate this, I'm applying `word-wrap: break-word` and `word-break: break-all` to the div only on mobile screens. I don't think we want to apply this everywhere because it will break words arbitrarily with no hyphens, making it look a bit goofy:

image

By applying only on small screens, we make sure that it only does these odd word breaks when absolutely necessary. There is a bit of a tradeoff here no matter what; we either choose to allow for strings to be cut off on Chrome with no visual representation (no `...`) or we break *within* words.

### DIFF
--- a/src/components/NftPreview/NftPreviewLabel.tsx
+++ b/src/components/NftPreview/NftPreviewLabel.tsx
@@ -40,7 +40,7 @@ export const StyledNftPreviewLabel = styled.div`
 
 const StyledBodyRegular = styled(BodyRegular)<{ lines: number }>`
   margin: 0;
-  text-shadow: 1px 1px 3px rgba(0, 0, 0, 1);
+  text-shadow: 1px 1px 3px rgba(0, 0, 0, .5);
   display: -webkit-box;
   -webkit-box-orient: vertical;
   -webkit-line-clamp: ${({ lines }) => lines};
--- a/src/components/NftPreview/NftPreviewLabel.tsx
+++ b/src/components/NftPreview/NftPreviewLabel.tsx
@@ -39,28 +39,36 @@ export const StyledNftPreviewLabel = styled.div`
 `;
 
 const StyledBodyRegular = styled(BodyRegular)<{ lines: number }>`
+  word-wrap: break-word;
+  word-break: break-all;
+
   margin: 0;
-  text-shadow: 1px 1px 3px rgba(0, 0, 0, .5);
+  text-shadow: 1px 1px 3px rgba(0, 0, 0, 0.7);
   display: -webkit-box;
   -webkit-box-orient: vertical;
+  line-clamp: ${({ lines }) => lines};
   -webkit-line-clamp: ${({ lines }) => lines};
   overflow: hidden;
   font-size: 12px;
   line-height: 16px;
   text-overflow: ellipsis;
-  // word-break: break-word;
-  // word-break: break-all;
 
-    @media only screen and ${breakpoints.tablet} {
-      font-size: 14px;
-      line-height: 20px;
-      display: unset;
-      -webkit-box-orient: unset;
-      -webkit-line-clamp: unset;
-      overflow: visible;
-      text-overflow: unset;
-    }
+  @media only screen and ${breakpoints.mobileLarge} {
+    word-wrap: unset;
+    word-break: unset;
   }
+
+  @media only screen and ${breakpoints.tablet} {
+    font-size: 14px;
+    line-height: 20px;
+    display: unset;
+    -webkit-box-orient: unset;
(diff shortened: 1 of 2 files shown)


src/components/NftPreview/NftPreviewLabel.tsx
```diff
@@ -27,14 +32,43 @@ export const StyledNftPreviewLabel = styled.div`
...
-const StyledBodyRegular = styled(BodyRegular)`
+const StyledBodyRegular = styled(BodyRegular)<{ lines: number }>`
+  word-wrap: break-word;
+  word-break: break-all;
+
   margin: 0;
+  text-shadow: 1px 1px 3px rgba(0, 0, 0, 0.7);
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  line-clamp: ${({ lines }) => lines};
```
kaitoo1, line 49: TIL! very cool

src/components/NftPreview/NftPreviewLabel.tsx
```diff
@@ -27,14 +32,43 @@ export const StyledNftPreviewLabel = styled.div`
...
+  -webkit-line-clamp: ${({ lines }) => lines};
+  overflow: hidden;
```
Robinnnnn, line 51: do you know why we're using `overflow: hidden`? looks like this only gets applied on mobile, but it causes a weird box-in effect on the text shadow:
Screen Shot 2022-02-22 at 12 43 44 AM

removing it fixes the issue. looks like desktop doesn't get `overflow: hidden` applied
Screen Shot 2022-02-22 at 12 44 22 AM

src/components/NftPreview/NftPreviewLabel.tsx
```diff
@@ -27,14 +32,43 @@ export const StyledNftPreviewLabel = styled.div`
   padding: 8px;
   z-index: 10;
   justify-content: end;
-  height: 100%;
+  // height: 100%;
```
Robinnnnn, line 35: toss?

kaitoo1:
nice this looks great - I like how the text shadow makes it a bit easier to read, good idea

kaitoo1:
+1 on robin's comments, otherwise LGTM

connorrothschild:
@Robinnnnn, confirmed that need `overflow: hidden` or else the text will overlap with itself/next lines. Pushed one solution:

Applied `padding-right` and then a negative `margin-right` to each text element. Before/after:

image

image

This fixes the `text-shadow` overflow effect and leaves the rest mostly unchanged. The only other difference now is that the truncated `...` will sometimes overflow (as you can see in the image on # 52 above)

But I think this actually makes sense since the truncation is suggesting there is more beyond the visible text. 

Thoughts on this solution? I am noticing that as we design around very small previews (e.g. each of the above are 74px), more and more weird edge cases and interactions come up

Robinnnnn:
@connorrothschild love the hack! this is perfect, gets us most of the way there!

Robinnnnn (approved):

PR #611, opened 2022-02-20
Ensure number of columns on Edit Page displays correctly despite window width

connorrothschild:
This adjusts the sizes of images when editing collections. It is designed to show the correct number of user-specified columns for users on a desktop device (>= 1100px). Here is a before and after of each number of user-selected columns:

image

---

All that's happening here is a reduction in sizes for images in each of the `n-`column layouts. 

The only other notable edit is that I have replaced the `margin: 24px` applied to each image with a `margin: 24px 0 0 24px`. (You can see how the assets in `NEW` above are closer together.) This 
1. allows the images to fill more of the available space
2. prevents rendering of dead space to the right of assets that are on the last column of each row (e.g. below):

<div style="display: flex;">
image
image
</div>

This does reduce the margin between elements, but I personally don't think that losing that space between assets is an issue? Happy to undo this change based on the team's thoughts.

---

This is an imperfect and intermediate solution only meant to solve the issue of columns rendering incorrectly for desktop users. Screens under 1100px will still break columns (although this may be a desired behavior?), and there might be a more complete solution in the future. For now, this fixes the user-reported issue and solves layout issues for users who edit their collection on desktop (most likely the vast majority of them).

### DIFF
--- a/src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
@@ -38,11 +38,11 @@ const DND_WIDTHS: Record<number, number> = {
 // Width of draggable image for each Column # setting
 const IMAGE_SIZES: Record<number, number> = {
   1: 400,
-  2: 320,
-  3: 280,
-  4: 207,
-  5: 156,
-  6: 122,
+  2: 330,
+  3: 212,
+  4: 153,
+  5: 122,
+  6: 91,
 };
 
 const defaultDropAnimationConfig: DropAnimation = {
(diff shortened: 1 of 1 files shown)


Robinnnnn:
hey connor, thanks so much for looking at this! unfortunately it looks like the new widths are quite small (before/after pics below).

after discussing with @kaitoo1 it seems that the best way to handle this would be to make a new breakpoint just for this screen, but that's kind of hacky / over-optimizing an area that we will likely refactor down the line.

since we're planning on beefing up this page later on, let's ignore this bug for now, since it's so rarely supported!

image

image

connorrothschild:
Sounds good to me!

PR #614, opened 2022-02-23
Add support for blank space blocks

kaitoo1:
## Summary:
Add support for blank space blocks so that users can add blank spaces within their collections via the editor.

## some notable changes:
- changed the editor context to use the new `StagingItem` type to represent any item that is in the dnd staging area. This can be either an `EditModeNft` or `WhitespaceBlock`.
- added a `collectionLayout` util with helper methods including one to compute the list of whitespace positions that will be sent to the backend when saving a collection
- updated sidebar logic to stage whitespace blocks in the editor according to the whitespace list when loading an existing collection

## checklist

**Editor**
- [x] "Add blank space" button in sidebar adds blank space to editor
- [x] User can drag blank space just as they can an NFT, and position is preserved when changing column #
- [x] Visual confirmation when a blank space is added, and when user hovers over or drags a blank space
- [x] The blank space positions submitted to the backend are the expected value both for creating + updating a collection
- [x] The list of nft ids submitted to analytics when creating a collection is not affected
- [x] The "Deselect All" button removes blank spaces in addition to NFTs
- [x] Loading the editor for an existing collection with blank spaces populates the editor accordingly

**Gallery + Collection view**
- [x] Blank space is rendered in the expected positions for each collection
- [x] Blank space is not rendered in list view on mobile
- [x] Blank space takes up correct dimensions if the entire row is blank and doesn't include NFTs


src/contexts/collectionEditor/CollectionEditorContext.tsx
```diff
@@ -81,59 +82,56 @@ type Props = { children: ReactNode };
...
-
-      return [...previous, ...nftsNotYetStaged];
-    });
+  const stageNfts = useCallback((nfts: StagingItem[]) => {
+    setStagedItemsState((previous) => [...previous, ...nfts]);
   }, []);
 
   const unstageNfts = useCallback((ids: string[]) => {
-    const idsMap = Object.fromEntries(ids.map((key) => [key, true]));
+    setStagedItemsState((previous) =>
```
kaitoo1, line 120: simplified filtering logic to remove unnecessary object mapping

src/contexts/collectionEditor/CollectionEditorContext.tsx
```diff
@@ -1,28 +1,28 @@
 import { createContext, memo, ReactNode, useCallback, useContext, useMemo, useState } from 'react';
 import { arrayMove } from '@dnd-kit/sortable';
 import { DragEndEvent } from '@dnd-kit/core';
-import { EditModeNft } from 'flows/shared/steps/OrganizeCollection/types';
+import { EditModeNft, StagingItem } from 'flows/shared/steps/OrganizeCollection/types';
```
kaitoo1, line 4: most of the changes in this file are to change the `EditModeNft` type to `StagingItem`

src/flows/shared/steps/OrganizeCollection/CollectionCreateOrEditForm.tsx
```diff
@@ -17,14 +17,15 @@ import { Collection, CollectionLayout } from 'types/Collection';
 import useAuthenticatedGallery from 'hooks/api/galleries/useAuthenticatedGallery';
 import useCreateCollection from 'hooks/api/collections/useCreateCollection';
 import Mixpanel from 'utils/mixpanel';
-import { Nft } from 'types/Nft';
+import { StagingItem } from './types';
```
kaitoo1, line 20: same here, most of the changes in this file are to change the EditModeNft type to StagingItem

src/flows/shared/steps/OrganizeCollection/Editor/SortableStagedBlankBlock.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { CSS } from '@dnd-kit/utilities';
+import UnstageButton from './UnstageButton';
+
+type Props = {
+  id: string;
+  size: number;
+};
+
+export default function SortableStagedBlankBlock({ id, size }: Props) {
+  const { attributes, listeners, isDragging, setNodeRef, transform, transition } = useSortable({
```
kaitoo1, line 16: lines 16-27 to get the dnd related props is duplicated across this file and `SortableStagedNft`, but I decided it was ok because it's a pain to prop drill these from a shared component, and it doesn't make sense to use a context because it's just two files and the properties are unique for every item in the dnd.

kaitoo1: we can revisit if we add another kind of sortable staged item in the future

src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
```diff
@@ -19,11 +19,12 @@ import { Heading } from 'components/core/Text/Text';
 import {
   useCollectionEditorActions,
   useCollectionMetadataState,
-  useStagedNftsState,
+  useStagedItemsState,
```
kaitoo1, line 22: the changes in this file are mostly changing `stagedNfts` to `stagedItems`

src/flows/shared/steps/OrganizeCollection/Editor/UnstageButton.tsx
```diff
@@ -3,19 +3,18 @@ import styled from 'styled-components';
 import { BodyRegular } from 'components/core/Text/Text';
 import colors from 'components/core/colors';
 import { useCollectionEditorActions } from 'contexts/collectionEditor/CollectionEditorContext';
-import { EditModeNft } from '../types';
 
 type Props = {
-  editModeNft: EditModeNft;
+  id: string; // id of staged item to remove
```
kaitoo1, line 8: simplified to passing in just the id, we didn't need the whole `editModeNft`

src/components/NftGallery/NftGallery.tsx
```diff
@@ -6,12 +6,38 @@ import { useMemo } from 'react';
...
+function insertWhitespaceBlocks(nfts: Array<Nft | WhitespaceBlock>, whitespaceList: number[]) {
+  const result = [...nfts];
+  // Insert whitespace blocks into the list of items to stage according to the saved whitespace indexes.
+  // Offset the index to insert at by the number of whitespaces already added
+  whitespaceList.forEach((index, offset) =>
+    result.splice(index + offset, 0, { id: `blank-${generate12DigitId()}`, isWhitespace: true })
+  );
+
+  return result;
+}
```
Robinnnnn, line 39: nit: this function should be good to unit test

src/flows/shared/steps/OrganizeCollection/CollectionCreateOrEditForm.tsx
```diff
@@ -106,17 +107,17 @@ function CollectionCreateOrEditForm({
       }
 
       // Collection is being created
-      if (!collectionId && nfts && layout) {
+      if (!collectionId && stagedItems && layout) {
         Mixpanel.track('Create collection', {
           added_name: title.length > 0,
           added_description: description.length > 0,
-          nft_ids: nfts.map((nft) => nft.id),
+          nft_ids: removeWhitespacesFromStagedItems(stagedItems).map(({ id }) => id),
```
Robinnnnn, line 114: great catch!

src/flows/shared/steps/OrganizeCollection/Editor/CollectionEditor.tsx
```diff
@@ -30,8 +31,22 @@ function convertNftsToEditModeNfts(nfts: Nft[], isSelected = false): EditModeNft
...
+): StagingItem[] {
+  const result = [...stagingItems];
+  // Insert whitespace blocks into the list of items to stage according to the saved whitespace indexes.
+  // Offset the index to insert at by the number of whitespaces already added
+  whitespaceList.forEach((index, offset) =>
+    result.splice(index + offset, 0, { id: `blank-${generate12DigitId()}` })
+  );
+
+  return result;
+}
```
Robinnnnn, line 46: should we abstract this? logic used elsewhere

src/flows/shared/steps/OrganizeCollection/Editor/SortableStagedWhitespace.tsx
```diff
@@ -0,0 +1,127 @@
...
+  useEffect(() => {
+    if (showFadeinAnimation.current) {
+      setTimeout(() => {
+        showFadeinAnimation.current = false;
+      }, fadeinAnimationDuration);
+    }
+  }, []);
+
+  return (
+    <StyledSortableWhitespace
```
Robinnnnn, line 43: ser you're a magician ⚡️

src/utils/collectionLayout.test.ts
```diff
@@ -0,0 +1,34 @@
...
+      { id: 'blank-3' },
+      generateTestNft(),
+      generateTestNft(),
+      generateTestNft(),
+      { id: 'blank-4' },
+      generateTestNft(),
+      { id: 'blank-5' },
+    ];
+    const whitespaceList = getWhitespacePositionsFromStagedItems(stagedItems);
+    expect(whitespaceList).toEqual([0, 0, 1, 4, 5]);
```
Robinnnnn, line 32: 😍😍😍😍😍😍😍

src/utils/collectionLayout.ts
```diff
@@ -0,0 +1,41 @@
...
+      result.push(nftIndex);
+    }
+  });
+
+  return result;
+}
+
+export function generate12DigitId() {
+  return Math.round(Math.random() * 1000000000000);
+}
```
Robinnnnn, line 30: nit: could we just use something like uuid? https://www.npmjs.com/package/uuid

kaitoo1: we could - this felt sufficient since we just need an key we can use to rendering stage+unstange the white space blocks, so it doesn't have to be a proper id. happy to add it if we think that's better

src/utils/collectionLayout.ts
```diff
@@ -0,0 +1,41 @@
...
+
+// filter whitespaces from stagedItems and map each EditModeNft -> Nft
+export function removeWhitespacesFromStagedItems(stagedItems: StagingItem[]) {
+  return stagedItems.reduce((filtered: Nft[], item) => {
```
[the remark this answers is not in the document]
kaitoo1: I basically wanted to do a .filter then a .map, where I'm filtering out the whitespaces, then mapping the list of `EditModeNfts` to a list of `EditModeNft.nft` s

Using `reduce` let me do that in one loop instead of 2

I agree with staying away from reduce unless it's necessary.. it's not exactly necessary here but there is a practical benefit over .filter().reduce() in this case which I think justifies it 😅

src/components/NftGallery/NftGallery.tsx
kaitoo1: good call, added to test file

src/flows/shared/steps/OrganizeCollection/Editor/CollectionEditor.tsx
kaitoo1: done

src/utils/collectionLayout.ts
Robinnnnn: for sure! this def isn't blocking, but i presume we'll have more cases to do ID generation in the future, both in the product and in tests, and uuid is generally widely accepted + safe. it's also technically less likely to produce a collision than this func

src/utils/collectionLayout.ts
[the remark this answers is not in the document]
Robinnnnn: good point! i'd argue that O(2n) is just as chill as O(n) – what we really want to avoid is quadratic n^2. and since this func isn't running _all the time_ (only when you click save), .filter + .map would be totally chill

PR #615, opened 2022-02-25
Add collection navigation functionality

connorrothschild:
This PR will add the ability for user's to navigate left/right to previous/next NFTs in a given collection by clicking the left/right arrows. 

Here is a preview of navigation on desktop/mobile:

image

On hover, the 'Next'/'Prev' text will appear. This is only true on screens larger than `420px` (our `mobileLarge` breakpoint).

image

---

Behind the scenes, the PR relies mostly on existing code in [`NavigationHandle.tsx`](https://github.com/gallery-so/gallery/pull/615/files#diff-a806797ab0d24bdf3836fad629525300af92d535db33aeeda60eb8235c61fbcb). 

The most notable edit to that component was how clicking an arrow would render navigation; the draft had this click triggering a `push`, but I found that this interfered with the ability to click `Back to Gallery/Collection` (because it updated the history). [Now](https://github.com/gallery-so/gallery/pull/615/files#diff-a806797ab0d24bdf3836fad629525300af92d535db33aeeda60eb8235c61fbcbR34), we use `replace` rather than push so history is replaced, and the user clicking `Back to Gallery/Collection` will still go back to the intended route. If we need to record each page in history and need `push`, we'll need to handle `useBackButton` differently for routes navigated to via navigation arrows.

---

## Known issues

Navigating away from an NFT detail page via 'Back to Gallery'/'Back to Collection' creates a very brief layout shift. This is because the arrows are removed as soon as the back button is clicked, and before the page transition is complete:

https://user-images.githubusercontent.com/13339581/155770232-97f48e08-9486-4111-ae61-7d0b372e3f18.mov

This only applies on screens where the presence of navigation arrows affects layouts (e.g. above `mobile` breakpoint and below `desktop` breakpoints. I assume this occurs because the flow is 
1. Navigation begins
2. `prevNftId` and `nextNftId` are `null` because there is no *current* NFT and no current `collectionId`
3. `<NavigationHandle />` is therefore not rendered
4. Navigation ends

In order to fix I think we would have to defer the removal of `collectionId` until after the page transition has completed, or something to that effect.


src/scenes/NftDetailPage/NavigationHandle.tsx
```diff
@@ -20,23 +21,26 @@ const HOVER_TEXT = new Map<number, string>([
...
 
-  const { push } = useRouter();
+  const { replace } = useRouter();
   const handleOnClick = useCallback(() => {
     // TODO(Terence): Figure out how to get this state across since Next doesn't support navigation state
     // void push(nftId, { state: { collection: [] } });
-    void push({ pathname: nftId });
-  }, [nftId, push]);
+    // void push({ pathname: nftId });
+    void replace(`/${username}/${collectionId}/${nftId}`);
```
connorrothschild, line 39: Is replacing history acceptable here, or is `push` needed? 

If so, we will need to be creative in how to handle `useBackButton` via clicking 'Back to Gallery' which by default will just go back in the user's history.

src/scenes/NftDetailPage/NftDetailNote.tsx
```diff
@@ -259,7 +254,16 @@ const StyledTextAreaWithCharCount = styled(TextAreaWithCharCount)<TextAreaProps>
 
   p {
     right: 0;
-    bottom: ${({ footerHeight }) => footerHeight}px;
+    bottom: 0;
+  }
+
+  // We only apply padding to account for footer, which is not fixed on mobile
```
connorrothschild, line 260: Unrelated to this fix, a cleanup bugfix that addressed this:

image

image

Can open a separate PR if we want but it is very minor lol

src/scenes/NftDetailPage/NftDetailPage.tsx
```diff
@@ -27,12 +31,21 @@ type Props = {
...
+  const collectionId = query.collectionId as string;
   // TODO: Should refactor to utilize navigation context instead of session storage
   const isFromCollectionPage =
     globalThis?.sessionStorage?.getItem('prevPage') === `/${username}/${collectionId}`;
 
+  const collectionNfts = useCollectionById({ id: collectionId })?.nfts ?? [];
+  const collectionIndexOfNft = collectionNfts?.findIndex((nft) => nft.id === nftId) ?? 0;
+
+  const prevNftId = collectionNfts[collectionIndexOfNft - 1]?.id ?? '';
+  const nextNftId = collectionNfts[collectionIndexOfNft + 1]?.id ?? '';
```
connorrothschild, line 47: 1. Checks to find collection, or if null returns an empty array
2. Finds index of current NFT, or returns 0
3. Assigns prev/next ID to be the surrounding elements if they exist, or empty string if not
4. Empty string prevents render of arrows

src/scenes/NftDetailPage/NavigationHandle.tsx
Robinnnnn: oooo good question, i think we should use `replace` for that reason

src/scenes/NftDetailPage/NftDetailNote.tsx
Robinnnnn: amazing, great catch!

src/scenes/NftDetailPage/NftDetailPage.tsx
```diff
@@ -27,12 +31,21 @@ type Props = {
...
+
   const username = window.location.pathname.split('/')[1];
-  const collectionId = query.collectionId;
+  const collectionId = query.collectionId as string;
   // TODO: Should refactor to utilize navigation context instead of session storage
   const isFromCollectionPage =
     globalThis?.sessionStorage?.getItem('prevPage') === `/${username}/${collectionId}`;
 
+  const collectionNfts = useCollectionById({ id: collectionId })?.nfts ?? [];
+  const collectionIndexOfNft = collectionNfts?.findIndex((nft) => nft.id === nftId) ?? 0;
```
Robinnnnn, line 44: i think we should `useMemo` here since `findIndex` takes O(n)

src/scenes/NftDetailPage/NftDetailPage.tsx
Robinnnnn: nit: instead of defaulting to empty string, i think it would be more explicit to default to `null`

Robinnnnn:
> Navigating away from an NFT detail page via 'Back to Gallery'/'Back to Collection' creates a very brief layout shift

we could also prevent this by always rendering "invisible" arrows that take up the space of arrows

Robinnnnn:
i think we want to align the arrows with the corner elements in the navbar - as seen in the figma. this will likely require absolute positioning

Screen Shot 2022-02-25 at 6 35 27 PM

image

Robinnnnn:
these arrows should be the same shape + color.

i love the arrow you're using in the NFT detail page, so we should use that in the navbar. i also noticed your arrow is black, but should be light grey (same as text / navbar)

Screen Shot 2022-02-25 at 6 37 45 PM

connorrothschild:
> these arrows should be the same shape + color.
> 
> i love the arrow you're using in the NFT detail page, so we should use that in the navbar. i also noticed your arrow is black, but should be light grey (same as text / navbar)
> 
> Screen Shot 2022-02-25 at 6 37 45 PM

Fixed the color! 

And this is interesting, they are the same on my device:




I wonder if there's some font rendering issue behind the scenes

connorrothschild:
With the layout changes, there's a new layout shift that occurs on navigation, where the content shifts up briefly. I'll investigate this tomorrow

https://user-images.githubusercontent.com/13339581/155828800-1300ec93-d627-4833-b073-f81bc716a842.mov

src/scenes/NftDetailPage/NftDetailPage.tsx
Robinnnnn: yupp that's the perfect way to memo!

you could also use optional chaining before square brackets as well. here's an example (renamed `collectinIndexOfNft` => `nftIndex`):
```js
const prevNftId = collectionNfts?.[nftIndex - 1]?.id ?? null
const nextNftId = collectionNfts?.[nftIndex + 1]?.id ?? null
```

you could also do something more exhaustive like the following, esp. since it would be a red flag for an NFT to not belong to a collection:
```js
const collectionNfts = useCollectionById({ id: collectionId })

const { prevNftId, nextNftId } = useMemo(() => {
  if (!collectionNfts) {
    // TODO: send error to sentry. technically all NFTs should belong to a collection.
    return {
      prevNftId: null,
      nextNftId: null,
    }
  }

  const nftIndex = collectionNfts.findIndex((nft) => nft.id === nftId)
  if (nftIndex === -1) {
    // TODO: send error to sentry. technically the NFT should exist within the collection.
    return {
      prevNftId: null,
      nextNftId: null,
    }
  }
  return {
    prevNftId: collectionNfts[nftIndex - 1]?.id ?? null,
    nextNftId: collectionNfts[nftIndex + 1]?.id ?? null,
  }
}, [collectionNfts])
```

Robinnnnn:
@connorrothschild woah that's weird that the arrows appear the same for you! it looks like the Prev and Next arrows don't have the `Helvetica Neue` font applied.

the navbar uses a `TextButton` component, so i think we just need to wrap the arrow / prev text in the same [`TextButton`](https://github.com/gallery-so/gallery/blob/main/src/components/core/Button/TextButton.tsx). this imports ActionText, which imports ButtonText, which ultimately inherits the Helvetica font family.

also, the arrows still don't look to have the same padding as the navbar – as in the arrows are further to the edge of the screen than the nav!

Screen Shot 2022-02-27 at 11 06 10 PM

finally, according to the figma, it looks like the next/prev are the same font size as the navbar. but for me the next/prev look bigger than the nav font size. i presume this will be fixed also by using TextButton

connorrothschild:
@Robinnnnn Ah yes, I totally neglected to update the styling since I started from an existing file (kinda just went on autopilot from there haha). Thanks for flagging, I'll update that to be stylistically in line with the Figma/Nav early tomorrow 🙏

connorrothschild:
@Robinnnnn Success 🙏 [preview](https://gallery-git-connor-collection-navigation-gallery-so.vercel.app/) and screenshots (desktop/mobile):

<div style="display: flex">
Screen Shot 2022-02-28 at 10 52 10 AM
Screen Shot 2022-02-28 at 10 58 08 AM
</div>

---

As a note, I disabled the `:hover` effect where the text/arrow becomes black, because the text appearing seems to be enough of a visual cue? 

The Figma *does not* have arrows getting darker on hover, but our PRD *does* mention it. Do you have a preference? (With/without hover effect):
<div style="display: flex;">
image
image
</div>

Robinnnnn:
@connorrothschild looks incredible- just sync'd with Jess and let's make the whole thing appear black on hover, including the text!

also sorry about the merge conflict! added a new hook so it's easier to detect when the screen is mobile mode

src/scenes/NftDetailPage/NftDetailPage.tsx
```diff
@@ -28,12 +30,41 @@ type Props = {
 function NftDetailPage({ nftId }: Props) {
   const { query } = useRouter();
 
+  const screenWidth = useBreakpoint();
+  const isMobileScreen = screenWidth === size.mobileLarge || screenWidth === size.mobile;
```
kaitoo1, line 34: we should be able to use the new hook robin added `useIsMobileWindowWidth`

@Robinnnnn I see `useIsMobileWindowWidth ` returns true for `size.mobile` only, could we expand it to include `size.mobileLarge` as well?

src/scenes/NftDetailPage/NavigationHandle.tsx
```diff
@@ -20,63 +20,94 @@ const HOVER_TEXT = new Map<number, string>([
...
+        <StyledHoverText>
+          <ActionText>{hoverText}</ActionText>
         </StyledHoverText>
       </StyledTextWrapper>
     </StyledNavigationHandle>
   );
 }
 
-const StyledTextWrapper = styled.div<{ direction: Directions }>`
+const StyledTextWrapper = styled(ActionText)<{ direction: Directions }>`
```
kaitoo1, line 54: We should avoid turning `StyledTextWrapper` into a p tag (`ActionText`) if we can because `StyledTextWrapper` has children that are block elements (the hover text + spacer divs).

Technically p tags shouldn't contain block elements like `div`s or other `p`s cus its invalid html:
https://stackoverflow.com/a/8696078


if the intention was to style the arrow, maybe we could do this instead: 
`<ActionText>{arrow}</ActionText>`

src/components/core/Page/GlobalNavbar/GlobalNavbar.tsx
```diff
@@ -22,6 +22,7 @@ const StyledGlobalNavbar = styled.div`
   display: flex;
   justify-content: flex-end;
 
+  position: relative;
```
kaitoo1, line 25: what's this for?
I see the commit message you added but I don't fully understand 😅
 `position GlobalNavBar and StyledBacklink so that collection navigation arrows cannot be clicked while in header`

kaitoo1:
Niceee!!
Thanks for the detailed PR description again.

Looks good, I just noticed a small issue for interaction with modals:
When I open a modal on the NFT detail page, the arrows appear above the "faded" background - they should probably appear behind so that they're not visible/interactive when a modal is open. Might be a matter of just finding the right z index.
navigation arrow modal



Otherwise looks good! I left some design feedback about the mobile view on discord. Desktop looks great though 👍

src/components/core/Page/GlobalNavbar/GlobalNavbar.tsx
connorrothschild: Before adding `position: relative`, the `  z-index: 1;` applied to `StyledGlobalNavbar` didn't do anything. `position: relative` creates a stacking context.

Because `z-index` wasn't applied to the navbar, that meant we could hover over "Back to Gallery" or "Sign In" (or whatever else is directly above Prev/Next arrows and unintentionally trigger the hover event on the arrows:

Here, I'm hovering over sign in but its instead registering a hover event on the arrow below:

image
 
The `position: relative` is just forcing the intended stacking context we wanted via `z-index: 1`

src/scenes/NftDetailPage/NftDetailPage.tsx
Robinnnnn: @kaitoo1 i think it should be a separate hook or an optional param in the hook bc the places i refactored were only looking for `size.mobile` specifically, not including mobileLarge

src/hooks/useWindowSize.ts
```diff
@@ -50,3 +50,8 @@ export function useBreakpoint(): size {
 export function useIsMobileWindowWidth() {
   return useBreakpoint() === size.mobile;
 }
+
+export function useIsMobileOrMobileLargeWindowWidth() {
+  const breakpoint = useBreakpoint();
+  return breakpoint === size.mobileLarge || breakpoint === size.mobile;
+}
```
connorrothschild, line 57: This is a long name but in line with above and accurate? can change

src/scenes/NftDetailPage/NavigationHandle.tsx
```diff
@@ -1,48 +1,63 @@
...
 
-function NavigationHandle({ direction, nftId }: Props) {
-  const arrow = useMemo(() => ARROWS.get(direction) ?? '', [direction]);
+function NavigationHandle({ direction, username, collectionId, nftId }: Props) {
+  const isMobileOrMobileLarge = useIsMobileOrMobileLargeWindowWidth();
+
+  const arrow = useMemo(
+    () => (isMobileOrMobileLarge ? MOBILE_ARROWS.get(direction) : ARROWS.get(direction) ?? ''),
+    [isMobileOrMobileLarge, direction]
+  );
```
connorrothschild, line 42: Might need review. I assume this is the best way to handle dynamic rendering (icons on mobileLarge and down, text on tablet and higher)

src/scenes/NftDetailPage/NavigationHandle.tsx
```diff
@@ -53,30 +68,67 @@ const StyledTextWrapper = styled.div<{ direction: Directions }>`
   display: flex;
   margin: auto;
   flex-direction: ${({ direction }) => (direction ? 'row-reverse' : 'row')};
-`;
+  position: absolute;
+  z-index: 100;
+  color: ${colors.gray50};
+  padding: 16px;
+  margin: 0 -16px -16px -16px;
```
connorrothschild, line 75: This adds a larger tap area to the arrow, but does not change its visual position thanks to negative margin

src/scenes/NftDetailPage/NavigationHandle.tsx
Robinnnnn: looks good!

Robinnnnn (approved):

PR #616, opened 2022-02-26
Authenticated analytics tracking

Robinnnnn:
**Main Changes**
- added `AnalyticsContext`
- implemented `useTrack` hook to use moving forward
- implemented `useAuthenticatedUserId` hook that gets the ID from auth context, **not** localStorage
- migrated auth-related tracking utils to `contexts/analytics/authUtil`

**Testing**
- clicked around app, didn't break
- confirmed data coming in on mixpanel

**Remaining TODOs**
- [x] address [setLoggedIn](https://github.com/gallery-so/gallery/pull/616/files#diff-653b47b011f110177d42dfb1d6480dba125054324aec8df56efa493a2a0c7b43R102)


pages/_app.tsx
```diff
@@ -36,7 +36,7 @@ const App: FC<{
   const relayCache = pageProps.__relayCache as RecordMap | undefined;
 
   useEffect(() => {
-    Mixpanel.track('Page view', { path: asPath });
+    _track('Page view', { path: asPath });
```
Robinnnnn, line 39: this is the only case where we use `_track` directly, since `_app.tsx` is above the AnalyticsContext, and page view tracking isn't imperative to track with authentication

src/components/core/Button/TextButton.tsx
```diff
@@ -28,7 +28,7 @@ function TextButton({
       onClick={onClick}
       underlineOnHover={underlineOnHover}
       disabled={disabled}
-      data-testId={dataTestId}
+      data-testid={dataTestId}
```
Robinnnnn, line 31: this typo was causing errors in the console. everywhere else we're using lowercased `data-testid`

src/utils/mixpanel.ts
```diff
@@ -1,92 +0,0 @@
-import mixpanel from 'mixpanel-browser';
```
Robinnnnn, line 1: this file was migrated to a combination of `contexts/analytics/authUtil` and `AnalyticsContext`

src/contexts/navigation/GalleryNavigationProvider.tsx
```diff
@@ -26,9 +27,15 @@ export function GalleryNavigationProvider({ children }: Props) {
...
 
+  const track = useTrack();
+
   useEffect(() => {
+    // history
     setHistoryStack((stack) => [...stack, asPath]);
-  }, [asPath]);
+
+    // analytics
+    track('Page view', { path: asPath });
```
Robinnnnn, line 37: moved from _app.tsx

PR #618, opened 2022-02-27
Edit collection from Main Gallery

jakzaizzat:
## Changes

Allow users to edit their collection through main gallery / homepage. 

CleanShot 2022-02-28 at 07 49 34

### DIFF
--- a/src/components/CopyToClipboard/CopyToClipboard.tsx
+++ b/src/components/CopyToClipboard/CopyToClipboard.tsx
@@ -32,7 +32,7 @@ export default function CopyToClipboard({ textToCopy, children }: Props) {
     <Container onClick={handleCopyToClipboard}>
       {isToastMounted && (
         <Toast isActive={isActive}>
-          <Caption color={colors.white}>Copied</Caption>
+          <Caption color={colors.white}>Link Copied</Caption>
         </Toast>
       )}
       {children}
@@ -59,6 +59,7 @@ const Toast = styled.div<{ isActive: boolean }>`
   position: absolute;
   background: black;
   padding: 4px 8px;
+  width: fit-content;
   pointer-events: none;
 
   animation: ${({ isActive }) => css`
--- a/src/flows/shared/components/WizardFooter/WizardFooter.tsx
+++ b/src/flows/shared/components/WizardFooter/WizardFooter.tsx
@@ -13,7 +13,6 @@ import { GalleryWizardProps } from 'flows/shared/types';
 import isPromise from 'utils/isPromise';
 import { useRouter } from 'next/router';
 import { useAuthenticatedUsername } from 'hooks/api/users/useUser';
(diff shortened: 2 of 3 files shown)


src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -32,7 +32,7 @@ export default function CopyToClipboard({ textToCopy, children }: Props) {
     <Container onClick={handleCopyToClipboard}>
       {isToastMounted && (
         <Toast isActive={isActive}>
-          <Caption color={colors.white}>Copied</Caption>
+          <Caption color={colors.white}>Link Copied</Caption>
```
jakzaizzat, line 35: Since we don't have a wallet address in the dropdown, I changed directly the text in the `CopyToClipboard` component. 

Later on, we can make it as a `props` if we have more use cases for this component

src/flows/shared/components/WizardFooter/WizardFooter.tsx
```diff
@@ -58,6 +62,11 @@ function WizardFooter({
         setIsLoading(false);
       }
 
+      if (isFromGalleryPage) {
+        back();
```
jakzaizzat, line 66: Need to use `back` instead of `push` to maintain scroll position when the user finished edit the collection

src/flows/shared/components/WizardFooter/WizardFooter.tsx
```diff
@@ -70,6 +79,11 @@ function WizardFooter({
   }, [collectionId, next, onNext, push, username]);
 
   const handlePreviousClick = useCallback(() => {
+    if (isFromGalleryPage) {
+      back();
```
jakzaizzat, line 83: Same as above

src/components/CopyToClipboard/CopyToClipboard.tsx
Robinnnnn: nice, +1 on making this based on props!

src/flows/shared/components/WizardFooter/WizardFooter.tsx
Robinnnnn: great catch 🔥

Robinnnnn (approved):
LGTM! tested out the functionality 🔥

jakzaizzat:
@kaitoo1 Thanks for the good catch! 

Updated the changes. Seems like we can simplify the changes by using `collectionId` as a flag. 

1. If there is `collectionId` in the query : back to the previous page with window back to retain scroll position
2. If not, we assume its coming from the wizard page

kaitoo1 (approved):

kaitoo1:
@jakzaizzat thanks for addressing, that logic makes sense!

LGTM

jakzaizzat:
@kaitoo1 Sorry. Missed out this scenario 🤦🏻‍♂️. Updated the changes

PR #623, opened 2022-03-01
Imgix + SSR proof of concept

kaitoo1:
This PR introduces using Imgix to get resized + scaled versions of images to prepare us for the indexer switch.

To serve as proof of concept and agree with the team on the implementation pattern before making the change across the whole app, I got it working with the Nft Detail Page only. We don't have to merge this, I can use it as the foundation for the rest of the conversion.

changes:
- updated the Nft Detail Page route to retrieve the nft from the backend in server side props
- added an `getImageSrcSet` util function to generate a signed srcset of the image's source url using imgix's library and token. imgix requires the url to be signed when using their web proxy to resize any public image. the token is not exposed to the browser because we are using it in serversideprops only and it does not have the `NEXT_PUBLIC_` env var prefix
- updated `ImageWithLoading` to take an optional srcset prop which will render the Imgix-provided image. The srcset includes two versions of the image, for 1x and 2x resolution to support normal resolution screens as well as high resolution ones such as retina displays. Using `srcset` allows us to tell the `img` tag what versions of the image it can use for different settings such as screen resolution or size. More info about `srcset` and `size` attributes: https://medium.com/@woutervanderzee/responsive-images-with-srcset-and-sizes-fc434845e948

considerations for implementing this across the app
- on other pages such as the Gallery and Collection pages, the srcset will include multiple versions of the image at varying widths. Imgix supports this. 
- we can do this page by page before we switch to the indexer. However, we must convert all pages where we use nft images before we switch to our indexer.
- this means we'll have to re-enable SSR for the members page, which we had turned off because it felt a bit slow.

### DIFF
--- a/pages/[username]/[collectionId]/[nftId]/index.tsx
+++ b/pages/[username]/[collectionId]/[nftId]/index.tsx
@@ -4,26 +4,42 @@ import { GetServerSideProps } from 'next';
 import GalleryRedirect from 'scenes/_Router/GalleryRedirect';
 import { MetaTagProps } from 'pages/_app';
 import { openGraphMetaTags } from 'utils/openGraphMetaTags';
+import { Nft } from 'types/Nft';
+import { getImageSrcSet } from 'utils/imageSrcSet';
 
 type NftDetailPageProps = MetaTagProps & {
-  nftId?: string;
+  nft?: Nft;
 };
 
-export default function NftDetailPage({ nftId }: NftDetailPageProps) {
-  if (!nftId) {
+export default function NftDetailPage({ nft }: NftDetailPageProps) {
+  if (!nft) {
     // Something went horribly wrong
     return <GalleryRedirect to="/" />;
   }
 
-  return <GalleryRoute element={<NftDetailPageScene nftId={nftId} />} footerIsFixed />;
+  return <GalleryRoute element={<NftDetailPageScene nft={nft} />} footerIsFixed />;
 }
 
 export const getServerSideProps: GetServerSideProps<NftDetailPageProps> = async ({ params }) => {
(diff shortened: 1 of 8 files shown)


pages/[username]/[collectionId]/[nftId]/index.tsx
```diff
@@ -4,26 +4,42 @@ import { GetServerSideProps } from 'next';
...
+
+  if (!data?.nft) {
+    return {
+      notFound: true,
+    };
+  }
+
+  // generate srcset for image resizing
+  const imageSrcSet = getImageSrcSet(data.nft.image_original_url, 600);
+
```
Comment (Robinnnnn) [inline pages/[username]/[collectionId]/[nftId]/index.tsx:39]: general Q: would we normally place this fetch in `getServerSideProps`? is that the correct way to handle in nextjs?

src/utils/imageSrcSet.ts
```diff
@@ -0,0 +1,29 @@
+import ImgixClient from '@imgix/js-core';
+
+export type ImageSrcSet = {
+  srcSet: string;
+  src: string;
+};
+
+const IMGIX_DOMAIN = 'assets.gallery.so';
```
Robinnnnn, line 8: note to self: still gotta set up SSL for this. seems like it works otherwise

pages/[username]/[collectionId]/[nftId]/index.tsx
Comment (kaitoo1) [reply pages/[username]/[collectionId]/[nftId]/index.tsx:39]: yup according to the docs! https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props#fetching-data-on-the-client-side

leaving here for others

Robinnnnn:
@kaitoo1 gonna close this, thanks for investigating!

PR #640, opened 2022-03-10
configure sentry tunnel with new proxy

kaitoo1:
This PR enables the tunnel option during Sentry.init to to use our Sentry proxy.

official docs:
https://docs.sentry.io/platforms/javascript/troubleshooting/#using-the-tunnel-option

bug successfully submitted on brave:
Screen Shot 2022-03-10 at 18 02 03

bug received in sentry:
Screen Shot 2022-03-10 at 17 58 42


Robinnnnn (approved):
so sorry i didn't review this earlier! thanks for crushing this kaito!

PR #643, opened 2022-03-16
Set user id using mixpanel identify function

kaitoo1:
We want to set the user's id as the `distinct_id` field in mixpanel when possible. This is done by calling mixpanel.identify() when the user logs in, or on page load and they are already logged in. 

Includes a bonus change to add Sentry logging when there's a Mixpanel error.


Robinnnnn (approved):
awesome, thanks for setting this up!

PR #648, opened 2022-03-19
Collectors note polish items

connorrothschild:
This PR fixes bugs related to collector's note functionality. In particular, it fixes two interrelated bugs related to the height of `textarea` elements not properly responding to user input.

Now, the collector's note `textarea` will 1. expand to meet its full height whenever text is added, 2. reduce to meet its new height whenever text is *removed*, and 3. take up the full height on a fresh page load.

https://user-images.githubusercontent.com/13339581/159140541-302778b3-add1-4ce4-a0a7-a16577a8d908.mov

Behind the scenes, this actually abstracts the height logic **out** of the collectors note component, and creates a [new](https://github.com/gallery-so/gallery/compare/connor/collectors-note-polish?expand=1#diff-e2b5bbd1c1e6855e61b21e0dcd0343a3d6fe6a03da6b6df34952292da9163667R74-R129) reusable component `AutoResizingTextAreaWithCharCount`. As the name implies, this resizes on its own based on inputs and can be used elsewhere as needed!

[Preview link 🔗](https://gallery-git-connor-collectors-note-polish-gallery-so.vercel.app/)


src/components/core/TextArea/TextArea.tsx
```diff
@@ -10,16 +10,14 @@ type TextAreaProps = {
...
 
-export function TextArea({
-  className,
-  onChange = noop,
-  placeholder,
-  defaultValue,
-  autoFocus = false,
-}: TextAreaProps) {
-  return (
+export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
```
Robinnnnn, line 16: nice use of forwardRef! 🔥 

so we still have a vanilla `TextAreaWithCharCount` component that uses this `TextArea` component. this doesn't pass in a ref – guessing that's ok?

image

src/components/core/TextArea/TextArea.tsx
```diff
@@ -70,11 +71,68 @@ export function TextAreaWithCharCount({
...
+  const [parentHeight, setParentHeight] = useState(DEFAULT_TEXTAREA_HEIGHT);
+
+  useEffect(() => {
+    if (textAreaRef.current) {
+      setParentHeight(`${textAreaRef.current.scrollHeight}px`);
+      setTextAreaHeight(`${textAreaRef.current.scrollHeight}px`);
+    }
+  }, [textAreaProps.defaultValue]);
+
+  const onChangeHandler = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
```
Robinnnnn, line 94: nit: let's wrap this in `useCallback` and name the function `handleChange` !

src/components/core/TextArea/TextArea.tsx
connorrothschild: It works (e.g. doesn't break anything) but I did come across [this rule](https://deepscan.io/docs/rules/react-useless-forward-ref) so likely not best practice.

It seems like this is our [best approach](https://stackoverflow.com/a/69127914):

```jsx
export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
  (
    { className, onChange = noop, placeholder, defaultValue, autoFocus = false, textAreaHeight },
    ref
  ) => {
    const _ref = useRef<HTMLTextAreaElement>(null);
    const textAreaRef = ref || _ref;
...
```

Alternatively, we could make a separate component that doesn't use forwardRef but is otherwise the same... Although the above feels better to me.

connorrothschild:
@Robinnnnn both comments addressed 👍

src/components/core/TextArea/TextArea.tsx
Robinnnnn: +1 your suggestion feels ideal. thanks for digging!

Robinnnnn (approved):

src/components/core/TextArea/TextArea.tsx
```diff
@@ -70,11 +76,68 @@ export function TextAreaWithCharCount({
...
+  ...textAreaProps
+}: TextAreaWithCharCountProps) {
+  const textAreaRef = useRef<HTMLTextAreaElement>(null);
+
+  const DEFAULT_TEXTAREA_HEIGHT = 'auto';
+
+  const [textAreaHeight, setTextAreaHeight] = useState(DEFAULT_TEXTAREA_HEIGHT);
+  const [parentHeight, setParentHeight] = useState(DEFAULT_TEXTAREA_HEIGHT);
+
+  useEffect(() => {
```
kaitoo1, line 92: it looks like both the `useEffect` and `handleChange` runs when the note is updated, leading to the textAreaHeight being set twice with different values, `DEFAULT_TEXTAREA_HEIGHT` and `textAreaRef.current.scrollHeight`. `setParentHeight` also gets called twice with the same value.
Not sure if that's intentional, I wonder if it can be simplified?

connorrothschild: @kaitoo1 I think the duplication of `useEffect` and `handleChange` is necessary, but we can def look into it further.

AFAIK `scrollHeight` will never decrease on text delete. Because of this, anytime the text changes, we have to run `setTextAreaHeight(DEFAULT_TEXTAREA_HEIGHT)` (or set it to zero), in case content is deleted and we want to shrink the text input. 

We could do something like `event.target.value.length < textAreaProps.defaultValue.length ? setTextAreaHeight(DEFAULT_TEXTAREA_HEIGHT) : null` so we only 'reset to zero' if the user deletes text?

---

Per your comment, I thought something like this would work (combining the 'reset' and the 'increase' into one callback):

```jsx
  const handleChange = useCallback((event: React.ChangeEvent<HTMLTextAreaElement>) => {
    // This is needed to reduce height of textarea on text delete
    setTextAreaHeight(DEFAULT_TEXTAREA_HEIGHT);

    if (textAreaRef.current) {
      setTextAreaHeight(`${textAreaRef.current.scrollHeight}px`);
      setParentHeight(`${textAreaRef.current.scrollHeight}px`);
    }

    if (textAreaProps.onChange) {
      textAreaProps.onChange(event);
    }
  }, []);
```

But it doesn't reduce textarea size on delete, I assume because the two `setTextAreaHeight`s are being run simultaneously (and the later one is taking priority). 

So I wrapped the ```
setTextAreaHeight(`${textAreaRef.current.scrollHeight}px`);
``` in a `setTimeout(() => {...}, 0)` but this led to some weird jitters on textarea increase 😔

https://user-images.githubusercontent.com/13339581/159530608-d75abb0c-74a0-4c06-9c48-e46711c216cb.mov

I think the reason the current solution works and is needed is because the `useEffect` runs after `handleChange` but before the next rerender. For reference, I was using [this article](https://medium.com/@lucasalgus/creating-a-custom-auto-resize-textarea-component-for-your-react-web-application-6959c0ad68bc_) to build the component, which explains that.

---

I think the duplication of ```setParentHeight(`${textAreaRef.current.scrollHeight}px`);``` is also needed, AFAIK, for the same reason.  If we remove the `setParentHeight` from the `useEffect`, we get the following effect when we delete a block of text:

https://user-images.githubusercontent.com/13339581/159534831-1a5fd552-45b2-4e91-8714-079e24637e28.mov

I think the current solution, separating into a `useEffect` and `handleChange`, works because the flow is something like this?
1. User submits note 
2. `textAreaProps.onChange(event)` is run
3. `textAreaProps.defaultValue` updates
4. This triggers `useEffect`
5. If text was deleted, scrollHeight can be reduced
---

Sorry for the long comment, kinda thinking out loud as I try to work through the logic here. Def open to other feedback and ideas toward improving this, if you can think of a better way to handle!

File: src/components/core/TextArea/TextArea.tsx [hunk 4]
```diff
@@ -70,11 +76,68 @@ export function TextAreaWithCharCount({
   );
 }
 
+export function AutoResizingTextAreaWithCharCount({
+  className,
+  currentCharCount,
```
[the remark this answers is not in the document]
connorrothschild: You're right, we don't need any of the above because they are passed in via `...textAreaProps`!

src/components/core/TextArea/TextArea.tsx
kaitoo1: @connorrothschild thanks so much for the detailed explanation and spending time trying out some alternatives! 
It makes sense to me now that when we delete text, we need to first reset the height so that we can use scrollHeight again to set the new, potentially reduced height. The way you set that up with the `onChange` followed by a `useEffect` is a nice solution to achieve this.

I think a comment would be great here explaining how this works since it's a bit unintuitive (at least for me) and it's tricky to discern why it's set up this way. example comment suggestion:
Screen Shot 2022-03-24 at 0 31 45

Given that it works as expected and having both `useEffect` and `onChange` is necessary, I think it's good to merge after just adding the comment!


------
Optional optimization explorations below
------
I did try some experimenting to see if we could eliminate any unnecessary calls for `setTextAreaHeight` or `setParentHeight` since I was curious, and these are **optional optimizations** so don't feel like you have to try the stuff below - just throwing it here if you're curious.


It looks like we can remove `setParentHeight` from the `onChange` without any adverse effects 👀 (we do need it in the `useEffect` like you said) From what I see the desired behavior is maintained.. maybe you'll notice otherwise? attached a video and screenshot of the code below. 


https://user-images.githubusercontent.com/80802871/159735208-ed798dc5-e1d2-40ba-97e1-ce496403342f.mp4


Screen Shot 2022-03-23 at 23 59 59



A further possible improvement is to only 'reset to zero' when we delete text like you suggested. That way, when we add text we only call `setTextAreaHeight` and `setParentHeight` once each. Again it looks like the desired behavior is maintained, attached video and screenshot of code below

https://user-images.githubusercontent.com/80802871/159737827-95122b89-c8d1-45ae-be85-4bc06ffc7378.mp4

Screen Shot 2022-03-23 at 23 54 01

connorrothschild: @kaitoo1  

Sorry for the delay and thanks for the great thoughts here! On optimization #1, it looks like there is a visual difference but only on Firefox — I guess its good we use a variety of browsers on the team :) 

without duplication (aka including in `onChange`), we see a more severe 'jump' when the scrollDown triggers, but *with* the duplication it is less severe:

https://user-images.githubusercontent.com/13339581/159789138-12550c94-f693-488f-b74a-65d96ed08858.mov

https://user-images.githubusercontent.com/13339581/159789143-7ff1a8df-c2d7-4044-a63d-49becc0d05f2.mov
 
It doesn't look like this occurs on Chrome. But for our few FF users, I think we keep the duplication 👍

--- 

I've been playing around with optimization 2 and it turns out we can't use `textAreaProps.defaultValue` for tracking because this only changes on submit. This means that if a user lengthens their collectors note (but doesn't submit), and then deletes it, the textarea won't resize unless their text length is shorter than the **already-submitted note**, rather than the one they are currently writing. Notice how below the textarea doesn't resize until we pass 17, which is the original textarea length:

https://user-images.githubusercontent.com/13339581/159791634-3e8ed183-3dff-4c1e-92a8-4eadea8d69e9.mov

I think that this is because `handleChange` is wrapped in a `useCallback`. If we instead made it a regular function, the numbers would update properly and we could do the exact optimization you described:

https://user-images.githubusercontent.com/13339581/159794200-d7ec9815-3865-4760-b0a1-20e0d09d8a50.mov

But I'm not sure if that optimization is worth ditching `useCallback`? Will defer to the perf experts on this one 🤔

cc @Robinnnnn

kaitoo1: @connorrothschild good catch on cross browser behavior for optimization 1.

For Optimization 2 - I'm a bit confused I might be missing something - if we use `event.target.value` to track the old and new values, it seems like it shouldn't matter if `textAreaProps.defaultValue` changes or not.
It seems to work for me using the callback:
Screen Shot 2022-03-24 at 14 07 34

https://user-images.githubusercontent.com/80802871/159846386-b07a85c9-c75f-424d-b286-f9300c445f52.mp4

connorrothschild: Sorry for the delay, I never saw the email notifying that there was another comment here!

@kaitoo1 You're totally right, my issue was I forgot the `oldText.current = event.target.value`. A bit of a facepalm moment for me 🤦‍♂️

Just updated to reflect the outcomes of this conversation! Think this is good to merge, I'll bring it up at weekly kickoff

kaitoo1: no problem thanks for checking it out! 
ah ok that makes sense, that's good news that it works 🙌
LGTM

kaitoo1 (approved):

PR #650, opened 2022-03-22
Always reset the walletconnect provider to allow multiple subsequent connections

kaitoo1:
This PR fixes a regression where the user was not able to connect a wallet using WalletConnect more than once. This meant they could not add another wallet to their account using WalletConnect if they had signed in with WalletConnect.

The fix is to reset the wallet connect provider every time the user clicks the WC option. Previously, we had only reset it if the user had encountered an error or otherwise aborted the WC flow previously.

This previously worked with us just removing the WC data from local storage every time the user clicked the WC option, but it appears something changed.
The solutions are from https://github.com/NoahZinsmeister/web3-react/issues/217

In the medium term, it looks like the beta of the next version of web3-react is out and the author suggests upgrading for a better dev experience. https://github.com/NoahZinsmeister/web3-react/tree/main#web3-react-beta

### DIFF
--- a/src/hooks/api/users/useRemoveUserAddress.ts
+++ b/src/hooks/api/users/useRemoveUserAddress.ts
@@ -37,7 +37,7 @@ export default function useRemoveUserAddress() {
       await mutate(
         getUserCacheKey({ id: user.id }),
         (user: User) => {
-          const addresses = user.addresses.filter((address) => address !== addressToRemove);
+          const addresses = user?.addresses.filter((address) => address !== addressToRemove);
           return { ...user, addresses };
         },
         false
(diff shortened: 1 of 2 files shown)


src/hooks/api/users/useRemoveUserAddress.ts
```diff
@@ -37,7 +37,7 @@ export default function useRemoveUserAddress() {
       await mutate(
         getUserCacheKey({ id: user.id }),
         (user: User) => {
-          const addresses = user.addresses.filter((address) => address !== addressToRemove);
+          const addresses = user?.addresses.filter((address) => address !== addressToRemove);
```
kaitoo1, line 40: small fix for a bug I saw when removing addresses from my user where user was undefined.
it otherwise didn't seem to affect behavior and we removing swr and these mutations soon so I don't think further investigation is necessary

Robinnnnn (approved):
incredible find and fix, thank you kaito! 🔥

PR #651, opened 2022-03-23
Fix jest config to support next compiler

tbezman:
I forgot to update our jest config when we upgraded to the Rust compiler. This fixes it!


Robinnnnn (commented):
😍

PR #652, opened 2022-03-23
Remove dead code

tbezman:
I used `ts-prune` to remove lingering code that we don't need anymore. There are more things to remove potentially but I only removed the ones that we're definitely not using and might make future refactors more difficult.


kaitoo1 (approved):

kaitoo1:
sweet, thanks for cleaning these up!

PR #654, opened 2022-03-23
capture web3error and error separately during wallet actions

kaitoo1:
This tries to address the following issue:
https://sentry.io/organizations/usegallery/issues/3122657620/events/ce4b0f3370a04bdcb13a559a69111561/?project=6187637

We aren't capturing web3 errors correctly because they are not Error types. The suggested solution is to just capture the message. If we find that isn't sufficient we can dig deeper on how to capture the full error.


Robinnnnn (approved):
great find, thank you kaito!

PR #655, opened 2022-03-24
Fix so wallet row is unmounted after disconnecting the wallet

kaitoo1:
**Summary**
This fixes a bug where upon disconnecting a wallet from a user account, the wallet row remained on screen. The expected behavior is that the wallet row is unmounted so that there is visual feedback that the wallet was removed.

As a bonus I also added a notification that says `Wallet ${removedAddress} has been removed.` upon successful removal, since we already have a notification upon successful addition.


Robinnnnn (approved):
🔥🔥🔥

PR #656, opened 2022-03-24
show username under Owned By on NFT Detail Page

kaitoo1:
## Summary

This PR changes the detail text on the NFT Detail Page to show username under "Owned By" instead of the owner's address.

(this means we won't show the .eth for user who have a .eth for "Owned By". We will still show .eth where applicable for "Created By")

Screen Shot 2022-03-24 at 21 42 06


Robinnnnn (approved):
cleannn thank you kaito! will merge

PR #677, opened 2022-04-05
[graphql] NFT detail page

Robinnnnn:


src/scenes/NftDetailPage/NftDetailView.tsx
```diff
@@ -0,0 +1,183 @@
+import { captureException } from '@sentry/nextjs';
```
kaitoo1, line 1: NftDetailPage was doing a lot, so I split it into 2 parts and moved the rendering logic into this new component, `NftDetailView`.

`NftDetailPage` fetches the nft + collection and redirects to Not Found if either are not found, and otherwise renders the page itself which includes `NftDetailView`.

`NftDetailView` renders the content of the page (image, text, collectors note, etc), and is only rendered if the nft and collection exist.

kaitoo1 (approved):

PR #678, opened 2022-04-06
Fix lintoooor

Robinnnnn:
## Fix linter

Turns out our linter has been broken for quite some time. The biggest risk here were some rules of hooks not getting caught. It also explains why we didn't catch #661 

## Added as a CI step to Github Actions

✅ Successful cache restore to prevent yarn re-installs 

image

✅ Fails job if lint fails

image

### DIFF
--- a/.github/workflows/code-lintooor.yml
+++ b/.github/workflows/code-lintooor.yml
@@ -1,0 +1,26 @@
+name: ci
+on: [push]
+jobs:
+  code-lintooor:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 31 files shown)


.github/workflows/code-lintooor.yml
```diff
@@ -0,0 +1,26 @@
+name: ci
```
Robinnnnn, line 1: seems like there's a bug in github actions where this worked for a hot sec, and is no longer running at all. i vote we merge this and hope that it eventually begins running

PR #679, opened 2022-04-06
Full width on landscape

Robinnnnn:
full width if single column + landscape (before/after)

image

60% width if single column + NOT landscape. current behavior is 50% (before/after)

image

### DIFF
--- a/src/components/NftGallery/NftGallery.tsx
+++ b/src/components/NftGallery/NftGallery.tsx
@@ -59,7 +59,7 @@ function NftGallery({ collection, mobileLayout }: Props) {
 const StyledCollectionNfts = styled.div<{ columns: number; mobileLayout: DisplayLayout }>`
   display: grid;
   grid-template-columns: ${({ columns, mobileLayout }) =>
-    mobileLayout === DisplayLayout.LIST ? '1fr' : `repeat(${columns},  minmax(auto, 50%))`};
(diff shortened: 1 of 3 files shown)


src/components/NftGallery/NftGallery.tsx
```diff
@@ -59,7 +59,7 @@ function NftGallery({ collection, mobileLayout }: Props) {
 const StyledCollectionNfts = styled.div<{ columns: number; mobileLayout: DisplayLayout }>`
   display: grid;
   grid-template-columns: ${({ columns, mobileLayout }) =>
-    mobileLayout === DisplayLayout.LIST ? '1fr' : `repeat(${columns},  minmax(auto, 50%))`};
+    mobileLayout === DisplayLayout.LIST ? '1fr' : `repeat(${columns},  minmax(auto, 100%))`};
```
Robinnnnn, line 62: always have the child fit to 100% of its space; the child's container will determine how wide it gets based on # of columns

src/contexts/shimmer/ShimmerContext.tsx
```diff
@@ -102,6 +102,8 @@ const Container = styled.div`
   display: flex;
   justify-content: center;
   align-items: center;
+
+  overflow: hidden;
```
Robinnnnn, line 106: fixes a bug with the shimmer

PR #706, opened 2022-04-20
Keyboard navigation functionality

connorrothschild:
This introduces keyboard navigation functionality in the following contexts:

- **NFT detail page**: navigation via `left`/`right`; back navigation via `escape`/`backspace`
- **Gallery editor**: back navigation via `escape`/`backspace`
- **Main gallery page**: edit main gallery via `e`
- **Collection page**: edit collection via `e`

The bulk of the work is done in the [`useKeyDown` hook](https://github.com/gallery-so/gallery/compare/connor/collection-keyboard-navigation#diff-b0a02934299f69e06089fac5417af9313e636eab9b6f32d035fe0531d38cba35), so this could use the greatest attention. It's basically [this hook](https://usehooks.com/useKeyPress/) with a few lines added to ensure the document body is in focus before triggering keyboard navigation. (This prevents a user from navigating while typing in modals, notes, etc.).

---

Two remaining questions, which I'll flag in comments below: 
1. When on a **collection page**, do we want escape/backspace to trigger navigation back to `/{username}`? This was not in the spec but feels consistent with the rest of our keyboard navigation. (I currently have it included but can remove.)
2. When a user leaves `/edit`, do we want to go back to the previous state in history? Or always go to /{username}?


src/hooks/useBackButton.tsx
```diff
@@ -19,8 +19,8 @@ export default function useBackButton({ username }: Props) {
   const backDisabled = previousRoute?.includes('/edit?collectionId=');
 
   const handleBackClick = useCallback(
-    (event: React.MouseEvent<HTMLElement>) => {
-      if (event.metaKey) {
+    (event?: React.MouseEvent<HTMLElement>) => {
+      if (event?.metaKey) {
```
connorrothschild, line 23: This enables the use of `useBackButton` without providing an event (it is traditionally triggered via click but this frees us up to use the same functionality elsewhere)

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
```diff
@@ -32,6 +34,28 @@ function CollectionGalleryPage({ collectionId, username, queryRef }: CollectionG
     });
   }, [username, collectionId, track]);
 
+  const ePress = useKeyDown('e');
+
+  // FIXME: Do we also want the escape/backspace keypress to trigger navigation to /{username}? Not in spec
```
connorrothschild, line 39: Flagging: This is not in spec but it feels in line with existing navigation functionality. So I've added it below

src/scenes/Modals/ConfirmLeaveModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useCallback } from 'react';
+import styled from 'styled-components';
+// import { useAuthenticatedUsername } from 'hooks/api/users/useUser';
+import useBackButton from 'hooks/useBackButton';
+
+function ConfirmLeaveModal() {
+  const { hideModal } = useModal();
+
+  // FIXME: Do we want escape on /edit to go back to the previous state in history? Or simply always go to /{username}?
+  const username = 'connorr'; // useAuthenticatedUsername();
```
connorrothschild, line 15: The new graphQL work goes a bit over my head and I need some help identifying the best way to get the user's username using this new configuration (we have removed `useAuthenticatedUsername` since I started work on this

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
```diff
@@ -32,6 +34,29 @@ function CollectionGalleryPage({ collectionId, username, queryRef }: CollectionG
...
+  }, [push, username]);
+
+  useEffect(() => {
+    if (ePress) {
+      navigateToEdit();
+    }
+    // FIXME: Do we also want the escape/backspace keypress to trigger navigation to /{username}? Not in spec
+    if (escapePress || backspacePress) {
+      navigateToUserGallery();
+    }
```
connorrothschild, line 57: Flagging

src/flows/shared/steps/OrganizeCollection/Editor/CollectionEditor.tsx
```diff
@@ -196,6 +199,15 @@ function CollectionEditor({ viewerRef }: Props) {
...
   const shouldDisplayEditor = stagedNfts.length > 0;
 
+  const escapePress = useKeyDown('Escape');
+  const { showModal } = useModal();
+
+  useEffect(() => {
+    if (escapePress) {
+      showModal(<ConfirmLeaveModal />);
+    }
+  }, [escapePress, showModal]);
```
Robinnnnn, line 209: love this!

src/hooks/useKeyDown.tsx
```diff
@@ -0,0 +1,43 @@
...
+      }
+    },
+    [targetKey]
+  );
+
+  // Add event listeners
+  useEffect(() => {
+    window.addEventListener('keydown', downHandler);
+    window.addEventListener('keyup', upHandler);
+    console.log('clicking ');
```
Robinnnnn, line 31: toss

src/hooks/useKeyDown.tsx
```diff
@@ -0,0 +1,43 @@
...
+    // Remove event listeners on cleanup
+    return () => {
+      window.removeEventListener('keydown', downHandler);
+      window.removeEventListener('keyup', upHandler);
+    };
+  }, [downHandler, upHandler]);
+
+  // If the user is not currently focused on the body, return
+  const activeEl = document.activeElement;
+  if (activeEl?.tagName !== 'BODY') return;
```
Robinnnnn, line 41: let's return `false` here, since in the case afterwards `keyPressed` will be a boolean! will make for consistent return types

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
```diff
@@ -32,6 +34,29 @@ function CollectionGalleryPage({ collectionId, username, queryRef }: CollectionG
...
   }, [username, collectionId, track]);
 
+  const ePress = useKeyDown('e');
+  const escapePress = useKeyDown('Escape');
+  const backspacePress = useKeyDown('Backspace');
+
+  const { push } = useRouter();
+
+  const navigateToEdit = useCallback(() => {
+    void push(`/edit?collectionId=${collectionId}`);
```
Robinnnnn, line 44: let's only navigate to the collection edit page for the given ID if the collection belongs to the user. when i'm on a collection page that i _don't_ own, it should not append the URL query param. 

current behavior:
- go to collection that i don't own: `/mikey/22PDnzYR9Xgj2lRRHrD3j7dNpVz`
- click "e"
- get navigated to `/edit?collectionId=22PDnzYR9Xgj2lRRHrD3j7dNpVz`, even though it's not my collection

desired behavior: clicking "e" should just navigate to my own edit page (i.e. /edit)

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
Robinnnnn: let's do it!

Robinnnnn: i think we can toss `backspacePress` though. just escape is fine

src/scenes/Modals/ConfirmLeaveModal.tsx
Robinnnnn: haha foshooo let's schedule time to pair on this!

src/scenes/NftDetailPage/NftDetailView.tsx
```diff
@@ -87,6 +90,36 @@ export default function NftDetailView({
...
+      navigateToId(prevNftId);
+    }
+    if (escapePress || backspacePress) {
+      handleBackClick();
+    }
+    // Placing prevNftId and nextNftId in the dependency list would trigger back again, if the user navigated via collection
+    // E.g. nft detail page (back) -> collection page (back) -> gallery. Is there a better solution here?
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [nextPress, prevPress, escapePress, backspacePress, handleBackClick]);
+
```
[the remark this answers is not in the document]
Robinnnnn: so thinking about this architecture more, i think we should have the hook receive a callback as an argument to trigger, rather than having an effect in the current component handle it.

example:
```js
// current
const nextPress = useKeyDown('ArrowRight');
const doThing = useCallback(() => { ... })
useEffect(() => {
  if (nextPress) doThing()
}, [nextPress, doThing])

// proposed
const doThing = useCallback(() => { ... })
useKeyboardShortcut('ArrowRight', doThing)
```

this would help encapsulate the side-effects into separate hooks, and the component doesn't need to worry about setting up an effect to handle business logic.

a full refactor might look something like this:

```js
const handleNextPress = useCallback(() => {
    if (nextNftId) navigateToId(nextNftId)
}, [nextNftId])

const handlePrevPress = useCallback(() => {
    if (prevNftId) navigateToId(prevNftId)
}, [prevNftId])

const handleBackClick = useBackButton({ username });

useKeyboardShortcut('ArrowRight', handleNextPress)
useKeyboardShortcut('ArrowLeft', handlePrevPress)
useKeyboardShortcut('Escape', handleBackClick)
```

i _believe this should fix your issue with random actions firing unintentionally. you'd need to update the internals of useKeyboardShortcut a bit

Robinnnnn:
> This keeps erroring during testing on: `should redirect to collection page when click the sign in button`. Seems like this is an issue with the test as it appeared in other PRs too?

sorry but do you mind clarifying this issue? would love to repro!

> When on a collection page, do we want escape/backspace to trigger navigation back to /{username}? This was not in the spec but feels consistent with the rest of our keyboard navigation. (I currently have it included but can remove.)

yup, i love this!

> When a user leaves /edit, do we want to go back to the previous state in history? Or always go to /{username}?

previous state in history! the same behavior when the user clicks "back" in the footer from the /edit page. i believe that's what you have currently 👍

connorrothschild:
> > This keeps erroring during testing on: `should redirect to collection page when click the sign in button`. Seems like this is an issue with the test as it appeared in other PRs too?
> 
> sorry but do you mind clarifying this issue? would love to repro!

Oops sorry @Robinnnnn I meant during the e2e checks. Not like in dev environment or anything

image

I think this error was occurring on the other most recent PRs too, so I figured its a nonissue?

src/hooks/useKeyDown.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    if (keyPressed) {
+      callbackFn();
+      // After executing, reset keyPressed to false so that the user can click the same key again on the same page
+      // Only currently relevant if user is on /edit -> escapes -> shows modal -> escapes -> clicks escape again
+      // This also prevents route-based navigation from duplicating
+      setKeyPressed(false);
+    }
+  }, [keyPressed, callbackFn]);
+}
```
connorrothschild, line 39: @Robinnnnn I fixed the error we were talking about at the end of our call by setting `setKeyPressed` to false right after execution of the callback function. 

This allows us to include `callbackFn` in the dependency array now (so maybe I was wrong about the reasons for the loop, but this fixes 👍)

Robinnnnn:
@connorrothschild yeahhh the e2e tests have been flaky, you can ignore!

src/hooks/useKeyDown.tsx
Robinnnnn: NIC!!!

Robinnnnn (approved):

Robinnnnn:
couple more bugs!
- when on a profile **while logged out**, like /robin, clicking "E" takes me back to the homepage
- when on a collection **while logged out**, clicking "E" takes me back to the homepage
to detect whether someone is logged out, you should be able to check whether the `query.viewer.user.username` is explicitly null

Robinnnnn (approved):

connorrothschild:
The last commit solves an issue where showing/hiding a modal via the escape key caused a jitter issue, because both `AnimatedModal.tsx` and `CollectionEditor.tsx` show and/or hide the same modal via the escape key.

This led to, when the user clicked escape to **hide** a modal, it flashing up and down briefly: 

https://user-images.githubusercontent.com/13339581/164312306-d17d44b4-2a07-48e5-a875-8dd5b781a016.mov

By adding a 150ms timeout, we get the following:

https://user-images.githubusercontent.com/13339581/164312506-4fcbe876-e3bf-4928-ac5b-81d1b6b26d34.mov

The difference for the user should be imperceptible.

Robinnnnn:
Perfect, thanks @connorrothschild!!

PR #707, opened 2022-04-20
fix image ratio on safari

kaitoo1:
this PR fixes image ratio bug for safari

before
Screen Shot 2022-04-20 at 12 03 27

after
Screen Shot 2022-04-20 at 12 04 18


Robinnnnn (approved):
😍

PR #709, opened 2022-04-21
Fix cypress metamask

jakzaizzat:
I found out that the reason cypress failed lately was because of different `username` of the test account. 

Updated from `gentlemanbeggar` to `silverlake`

**Other changes**
* Move all the static data to fixture
* Add fallback method if the metamask window didn't popup


Robinnnnn (commented):
username is not always guaranteed to be `silverlake`! discussing in discord

jakzaizzat:
@Robinnnnn Removed the fixtures. Now we compare the url slug and `h1` tag to check the username.

Robinnnnn (approved):

PR #714, opened 2022-04-23
Fix weird keyboard navigation bug

connorrothschild:
This fixes a bug where if you type a backspace in a textarea and then click right or left arrow key the app navigates to `/{username}` rather than the next/previous NFT id. This was because it was registering the `Backspace` key and "queuing" it up for when the next key press was registered. 

https://user-images.githubusercontent.com/13339581/164838867-8fc6a483-f05c-4c10-a914-ae4149453831.mov

Fixed by resetting `keyPressed` to false when key is released. Also, moved the `activeEl` declaration outside of the useEffect so that we listen for changes in focused element.

### DIFF
--- a/src/components/core/TextArea/TextArea.tsx
+++ b/src/components/core/TextArea/TextArea.tsx
@@ -27,6 +27,7 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
         placeholder={placeholder}
         defaultValue={defaultValue}
         onChange={onChange}
+        onKeyUp={(e) => e.stopPropagation()} // To prevent keyboard navigation from triggering while in textarea
         autoFocus={autoFocus}
         autoComplete="off"
         autoCorrect="off"
--- a/src/hooks/useKeyDown.tsx
+++ b/src/hooks/useKeyDown.tsx
@@ -22,11 +22,12 @@ export default function useKeyDown(targetKey: string, callbackFn: () => void) {
     };
   }, [pressHandler]);
 
+  const activeEl = document.activeElement;
+
   // useEffect ensures that the callback function is called once, which is relevant when navigating
   useEffect(() => {
     // If the user is not currently focused on the body, return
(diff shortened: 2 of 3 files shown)


connorrothschild:
Also, while you're here:

Should we be more explicit and only disable accidental navigation if the user is focused on a `textarea`? Currently we require the `body` to be focused. This means that if a user clicked a button ("Additional Information" etc) they also could not navigate until they click on the body again.

I just couldn't think if there were other focused elements that might trigger accidental navigation... So figured this was a bit safer.

Robinnnnn (commented):
> Should we be more explicit and only disable accidental navigation if the user is focused on a textarea? Currently we require the body to be focused. This means that if a user clicked a button ("Additional Information" etc) they also could not navigate until they click on the body again.

great observation. i definitely want to avoid the scenario where clicking on "additional information" would kill keyboard navigation.

any way we could use [`preventDefault`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) and/or [`stopPropagation`](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation) on the textarea component? this may be able to stop the keystroke from being picked up by useKeyDown

src/components/core/TextArea/TextArea.tsx
```diff
@@ -27,6 +27,7 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
         placeholder={placeholder}
         defaultValue={defaultValue}
         onChange={onChange}
+        onKeyUp={(e) => e.stopPropagation()} // To prevent keyboard navigation from triggering while in textarea
```
connorrothschild, line 30: @Robinnnnn Now including this in `<TextArea>` and we no longer need any check for focus in `useKeyDown`

Robinnnnn: niceee that's sick!

Robinnnnn (approved):

PR #715, opened 2022-04-24
Prevent accidental image selection when editing collections/galleries

connorrothschild:
This prevents accidental highlighting of NFT preview images while in editing contexts, by adding `user-select: none` to three places:

### Gallery editor

Here we add `user-select: none` and `pointer-events: none` to the individual images so that the user does not accidentally drag the **image** when they intend to drag the **row**.

https://user-images.githubusercontent.com/13339581/164994892-fa334480-57dd-46cf-a34e-5512ef77765a.mov

### Collection sidebar

Here we add `user-select: none` to individual images so that a user who tries to drag images (thinking dragging is a functionality here, which it is not) do not see highlighting of the other NFTs:

https://user-images.githubusercontent.com/13339581/164995045-e38e6a48-76a7-4849-b924-ccc0565cee02.mov

### Collection editor main view

Finally we add a `user-select: none` to each individual image so that the user does not accidentally select the inner text while trying to reorganize NFT order:

https://user-images.githubusercontent.com/13339581/164995181-60c94036-d0fe-4239-804b-5a9c47c411a7.mov

### DIFF
--- a/src/flows/shared/steps/OrganizeCollection/Editor/SortableStagedNft.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Editor/SortableStagedNft.tsx
@@ -87,6 +87,8 @@ export const StyledSortableNft = styled.div`
   cursor: grab;
   margin: 24px;
 
+  user-select: none;
+
   &:hover ${StyledUnstageButton} {
     opacity: 1;
   }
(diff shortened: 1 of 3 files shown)


Robinnnnn (approved):
😍😍😍

kaitoo1:
nice!
I still see the the sidebar icons highlighted on chrome, I wonder if `user-select:none` can be applied to the parent instead?

Screen Shot 2022-04-26 at 19 55 20

PR #718, opened 2022-04-27
GAL-18 Added Poster Page

jakzaizzat:
## Updates

* Refactor the banner component to support new UI
* Refactor the toast component to support new design
* Added a new banner for the poster event
* New pages `/object006`
* Added to FF


src/hooks/useTimer.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from 'react';
+
+// TODO: Abstract this as a props
+const mintDate = new Date('2022-05-23T00:00:00');
```
jakzaizzat, line 4: Lmk if we wanna move this `date` somewhere else or any changes in the format

pages/object006.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { FeatureFlag } from 'components/core/enums';
+import GalleryRedirect from 'scenes/_Router/GalleryRedirect';
+import { graphql, useLazyLoadQuery } from 'react-relay';
+import { object006Query } from '__generated__/object006Query.graphql';
+import GalleryAuthenticatedRoute from 'scenes/_Router/GalleryAuthenticatedRoute';
+
+export default function Poster() {
+  const query = useLazyLoadQuery<object006Query>(
+    graphql`
+      query object006Query {
```
Robinnnnn, line 12: 🔥

src/components/Banner/Banner.tsx
```diff
@@ -31,40 +41,82 @@ export default function Banner({ queryRef, text, requireAuth = false }: Props) {
 
   const isAuthenticated = Boolean(query.viewer?.user?.id);
 
-  const [dismissed, setDismissed] = usePersistedState(
-    JAN_18_MIGRATION_BANNER_DISMISSED_LOCAL_STORAGE_KEY,
-    false
-  );
+  const [dismissed, setDismissed] = usePersistedState(LOCAL_STORAGE_KEY || '', false);
```
Robinnnnn, line 44: do we need to reassign to the const on L26? why not just keep this as `localStorageKey`

src/scenes/_Router/GalleryRoute.tsx
```diff
@@ -86,7 +96,21 @@ export default function GalleryRoute({
...
+    return isFeatureEnabled(FeatureFlag.POSTER_PAGE) ? (
+      <Banner
+        title={countdownTimer}
+        text="Thank you for being a member of Gallery. Celebrate our new brand with us by signing our 2022 Community Poster that we will mint as an NFT."
+        queryRef={query}
+        localStorageKey="gallery-poster-banner"
+        requireAuth
+      >
+        <NavLink to="/object006">Sign Poster</NavLink>
+      </Banner>
```
Robinnnnn, line 109: i don't think the "Sign Poster" should necessarily be passed as `children`, since it describes the action component instead of the main text! how about something like this:

```js
      <Banner
        title={countdownTimer}
        text="Thank you for being a member of Gallery. Celebrate our new brand with us by signing our 2022 Community Poster that we will mint as an NFT."
        queryRef={query}
        localStorageKey="gallery-poster-banner"
        requireAuth
        actionComponent={<NavLink to="/object006">Sign Poster</NavLink>}
      />
```

then in `Banner.tsx` you can just render `{actionComponent}`

src/components/Banner/Banner.tsx
```diff
@@ -31,40 +41,82 @@ export default function Banner({ queryRef, text, requireAuth = false }: Props) {
...
-      <BaseM color={colors.white}>{text}</BaseM>
-    </StyledBanner>
+    <StyledContainer>
+      <StyledBanner>
+        <StyledContent>
+          {title && <StyledTitle>{title}</StyledTitle>}
+          <StyledText color={colors.offBlack}>{text}</StyledText>
+        </StyledContent>
+        <StyledAction>
+          <div onClick={hideBanner}>{children}</div>
```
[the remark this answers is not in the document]
Robinnnnn: left some thoughts on this here: https://github.com/gallery-so/gallery/pull/718/files#r859430816

in terms of `hideBanner`, do we need to put it in its own div? could we just add it to `StyledAction`?

e.g.

```js
<StyledAction onClick={hideBanner}>
  {actionComponent}
</StyledAction>
```

src/hooks/useTimer.ts
```diff
@@ -0,0 +1,41 @@
...
+  const hasEnded = countDown <= 0;
+
+  return {
+    days,
+    hours,
+    minutes,
+    seconds,
+    hasEnded,
+  };
+}
```
Robinnnnn, line 41: nit: doesn't block this PR, but maybe we can write some unit tests for this? wanna make sure things still look good if there's just 1 minute left (e.g. `00:00:01:00`)

src/hooks/useTimer.ts
```diff
@@ -0,0 +1,41 @@
...
+  const currentTime = Date.now();
+
+  const [countDown, setCountDown] = useState(mintDate.getTime() - currentTime);
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setCountDown(mintDate.getTime() - Date.now());
+    }, 1000);
+
+    return () => clearInterval(interval);
```
Robinnnnn, line 16: nice clear 😍

src/scenes/PosterPage/PosterPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+import StyledBackLink from 'components/NavbarBackLink/NavbarBackLink';
+import { useIsMobileWindowWidth } from 'hooks/useWindowSize';
+import { useToastActions } from 'contexts/toast/ToastContext';
+import useTimer from 'hooks/useTimer';
+
+export default function PosterPage() {
+  const isMobile = useIsMobileWindowWidth();
+  const { pushToast } = useToastActions();
+
+  const FIGMA_URL = 'https://www.figma.com/file/YrjZkebwPMIPr3aFx8o2w1/Untitled?node-id=1%3A2';
```
Robinnnnn, line 19: that's the difference between this URL and the figmaEmbedUrl?

src/hooks/useTimer.ts
Robinnnnn: also nit: judging by how the values are used, we could return:

```js
  return {
    timestamp: `${days}:${hours}:${minutes}:${seconds}`,
    hasEnded,
  };
```

instead of 

```js
  return {
    days,
    hours,
    minutes,
    seconds,
    hasEnded,
  };
```

src/scenes/PosterPage/PosterPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+    // TODO: Replace with hook
+    window.history.back();
+  };
+
+  const handleSignPoster = () => {
+    pushToast(
+      'Thank you for participating in the (Object 006) 2022 Community Poster event.',
+      () => {},
+      true
+    );
```
Robinnnnn, line 35: nit: the variables being passed in here are a bit hard to understand, we could update it to accept an object instead:

```js
pushToast({
  message: 'Thank you for participating in the (Object 006) 2022 Community Poster event.',
  onDismiss: () => {},
  autoClose: true,
})
```

would need to refactor other usages though

src/scenes/_Router/GalleryRoute.tsx
```diff
@@ -86,7 +96,21 @@ export default function GalleryRoute({
...
   }, [footer, footerVisibleOutOfView, footerVisibleWithinView, footerIsFixed, isMobile]);
 
-  const banner = useMemo(() => <Banner text="" queryRef={query} />, [query]);
+  const banner = useMemo(() => {
+    return isFeatureEnabled(FeatureFlag.POSTER_PAGE) ? (
+      <Banner
+        title={countdownTimer}
+        text="Thank you for being a member of Gallery. Celebrate our new brand with us by signing our 2022 Community Poster that we will mint as an NFT."
+        queryRef={query}
+        localStorageKey="gallery-poster-banner"
```
kaitoo1, line 105: tiny nit: let's use underscores for the local storage key `gallery_poster_banner` for consistency with other storage keys. the other keys are defined in `storageKeys.ts` so we can do the same for this too

src/components/Banner/Banner.tsx
kaitoo1, line 58: instead of disabling the banner when they click "Sign Poster" in the banner, let's wait to disable it until they click the "Sign Poster" button on the actual page (that opens Figma)

that way the user still has a way of coming back to this page if they decide to do it later

src/scenes/PosterPage/PosterPage.tsx
jakzaizzat: A little bit different. In the embed Figma, there is `https://www.figma.com/embed?embed_host=share&url=${FIGMA_URL}` in the front. 


I created a separate component for Embed Figma `<PosterFigmaFrame>` that accepts the `FIGMA_URL` as a prop. So, we can only manage it in one place.

src/components/Banner/Banner.tsx
Robinnnnn: such a good idea

Robinnnnn: a simple fix would be to just remove the hideBanner handler from the action component, and simply have the user dismiss it by clicking the "X"

```js
        <StyledAction>
          {actionComponent}
          <StyledClose onClick={hideBanner}>&#x2715;</StyledClose>
        </StyledAction>
```

jakzaizzat: @Robinnnnn Great idea! Seems like a simple fix. Will implement this way first. 

I've been thinking on how to close the `banner` from outside the `banner` component itself 🤔 . Seems like over-engineer the component lol

src/scenes/PosterPage/PosterPage.tsx
jakzaizzat: Nice! This one much more readable. I can do follow up PR on this since I might need to check other places too

src/scenes/_Router/GalleryRoute.tsx
jakzaizzat: Beautiful 😍. Thanks for the suggestion robin!

src/hooks/useTimer.ts
jakzaizzat: Definitely! I can write it in the next PR

src/scenes/PosterPage/PosterPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <InteractiveLink href="/">Gallery</InteractiveLink>
+        </div>
+        <StyledParagraph>
+          <BaseM>
+            Thank you for being a member of Gallery. Celebrate our new brand with us by signing our
+            poster.
+          </BaseM>
+          <BaseM>
+            The final product will be available to mint as a commemorative token for early believers
+            in our mission and product
```
kaitoo1, line 55:
```suggestion
            in our mission and product.
```

src/scenes/_Router/GalleryRoute.tsx
```diff
@@ -86,7 +100,22 @@ export default function GalleryRoute({
     }
   }, [footer, footerVisibleOutOfView, footerVisibleWithinView, footerIsFixed, isMobile]);
 
-  const banner = useMemo(() => <Banner text="" queryRef={query} />, [query]);
+  const banner = useMemo(() => {
+    const hideBannerPages = ['/auth'];
```
kaitoo1, line 104: nice, I think we can also hide the banner on the poster page. `/members/poster`

src/components/Banner/Banner.tsx
Robinnnnn: oh yeah... we'd need some global context like the ToastProvider haha

src/scenes/_Router/GalleryRoute.tsx
```diff
@@ -57,6 +66,12 @@ export default function GalleryRoute({
     queryRef
   );
 
+  const { timestamp } = useTimer();
+
+  const countdownTimer = useMemo(() => {
+    return `Ends in ${timestamp}`;
+  }, [timestamp]);
```
Robinnnnn, line 73: <no action required, just an observation!>

it feels a bit weird defining this in the `GalleryRoute.tsx` file as the banner won't be rendered on every screen (especially before the poster page is live). so we'll essentially have this `countdownTimer` running in the background for every GalleryRoute that's used.

overall i'm not worried because this will be a temporary feature, and this will not cause unnecessary re-renders. so performance will be fine.

one way to address this in the future may be to define a separate PosterBanner that encapsulates the countdown logic:
```js
// GalleryRoute.tsx
const bannerComponent = useMemo(() => {
  return isFeatureEnabled(FeatureFlag.POSTER_PAGE)
    ? <PosterBanner queryRef={query} />
    : <Banner text="" queryRef={query} />
}, [])

...

// PosterBanner.tsx
const PosterBanner = ({ queryRef }) => {
  const { timestamp } = useTimer();

  const countdownTimer = useMemo(() => {
    return `Ends in ${timestamp}`;
  }, [timestamp]);
  ...
}
```

Robinnnnn (approved):

src/scenes/_Router/GalleryRoute.tsx
jakzaizzat: Agree with you robin. It felt weird let this function running in the background although we didn't used it. 

Thanks for the suggestion!

Sorry robin. I had to do it  Just abstract this to his own component...

Robinnnnn (approved):

kaitoo1 (approved):

PR #729, opened 2022-05-02
[GAL-125] Fix tiny svgs

Robinnnnn:
| Before | After |
| ----- | ----- |
| image | image |

- [x] works on mobile
- [x] duaneking's gallery still loads okay (he has multiple SVG collections)

### DIFF
--- a/src/components/NftPreview/NftPreview.tsx
+++ b/src/components/NftPreview/NftPreview.tsx
@@ -12,6 +12,7 @@ import Gradient from 'components/core/Gradient/Gradient';
 import styled from 'styled-components';
 import NftPreviewLabel from './NftPreviewLabel';
 import { getBackgroundColorOverrideForContract } from 'utils/nft';
+import getVideoOrImageUrlForNftPreview from 'utils/graphql/getVideoOrImageUrlForNftPreview';
 
 type Props = {
   galleryNftRef: NftPreviewFragment$key;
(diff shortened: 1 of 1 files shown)


src/components/NftPreview/NftPreview.tsx
```diff
@@ -99,13 +101,33 @@ function NftPreview({ galleryNftRef }: Props) {
...
+    //
+    // HOWEVER, stretching an svg to 100% when column size = 1 results
+    // in the preview label appearing stretched beneath the image, since
+    // we cap the max height to 80vh when column = 1; so this is disabled
+    // in those cases for now.
+    if (columns > 1 && result?.urls?.large?.endsWith('.svg')) {
+      return '100%';
+    }
+    return 'auto';
+  }, [columns, result?.urls?.large]);
```
Robinnnnn, line 118: this causes a regression in 1-column collections with SVGs in terms of the floating label: 
image

in other words,
1) IF `octal` decides to display his bird in single-column view, it would still appear very small.
2) BUT fixing this isn't straightforward and it would cause a regression in other galleries

this PR compromises by fixing SVG appearance in 1+ column layout views

kaitoo1: nice! good compromise. maybe medium term we can just separate out the logic for rendering single column nfts specifically since that seems to be the case that's giving us the most edge cases. that should fix svg sizes for single-column view once and for all

kaitoo1 (approved):

PR #731, opened 2022-05-03
address poster page design feedback

kaitoo1:
- remove gallery link underneath title
- adjust layout to resemble nft detail page layout
- add footer on desktop

Before (desktop)
Screen Shot 2022-05-03 at 14 33 12

After (desktop)
Screen Shot 2022-05-03 at 14 34 02

Before (mobile)
Screen Shot 2022-05-03 at 14 34 32

After (mobile)
Screen Shot 2022-05-03 at 14 34 24

### DIFF
--- a/src/scenes/_Router/GalleryRoute.tsx
+++ b/src/scenes/_Router/GalleryRoute.tsx
@@ -70,8 +70,8 @@ export default function GalleryRoute({
       return <GlobalNavbar queryRef={query} />;
     }
 
-    return <Filler tallVariant={isMobile} />;
-  }, [navbar, isMobile, query]);
+    return <Filler tallVariant={isMobile && footer} />;
+  }, [navbar, isMobile, footer, query]);
 
   const footerComponent = useMemo(() => {
     if (!footer) {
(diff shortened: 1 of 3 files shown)


src/scenes/_Router/GalleryRoute.tsx
```diff
@@ -70,8 +70,8 @@ export default function GalleryRoute({
       return <GlobalNavbar queryRef={query} />;
     }
 
-    return <Filler tallVariant={isMobile} />;
-  }, [navbar, isMobile, query]);
+    return <Filler tallVariant={isMobile && footer} />;
```
kaitoo1, line 73: it looked like `tallVariant` was used to account for when the footer was taller, so I don't think it needs to be applied if there is no footer.

It was making the filler unnecessarily taller when there was no footer.
@Robinnnnn lmk if it might affect anything else that I might've missed!

Robinnnnn: this should be fine- i wanna refactor all of this anyway! thanks kaito

Robinnnnn (approved):

PR #742, opened 2022-05-06
Add await to relayFetchFunction

Robinnnnn:
Helps fix an issue in tracing where a span for a request would last a very short amount of time, rather than the length of the request.

More details in comments!

### DIFF
--- a/src/contexts/relay/relayFetchFunction.ts
+++ b/src/contexts/relay/relayFetchFunction.ts
@@ -13,7 +13,7 @@ export const relayFetchFunction: FetchFunction = async (request, variables) => {
   const transaction = initSentryTracing(request);
   // ---------- end pre-request hooks
 
-  const response = _fetch<GraphQLResponse>(`${baseurl}/glry/graphql/query`, {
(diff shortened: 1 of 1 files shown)


src/contexts/relay/relayFetchFunction.ts
```diff
@@ -13,7 +13,7 @@ export const relayFetchFunction: FetchFunction = async (request, variables) => {
   const transaction = initSentryTracing(request);
   // ---------- end pre-request hooks
 
-  const response = _fetch<GraphQLResponse>(`${baseurl}/glry/graphql/query`, {
+  const response = await _fetch<GraphQLResponse>(`${baseurl}/glry/graphql/query`, {
```
Robinnnnn, line 16: without the `await`, the sentry teardown function on L27 below would be invoked immediately, sending a transaction to sentry even before the request finished. props to @radazen for catching this!

i made sure to test that the app continued to work and handled errors properly. it seems like relay doesn't care whether what's returned from this function is a promise vs. a full-fledged response.

PR #743, opened 2022-05-08
[GAL-89] [GAL-90] Global Layout Refactor

Robinnnnn:
#### ⚠️ NOTE: This PR also includes large changes that implement the **NFT Detail Modal**. That diff is more clearly laid out in #763, featuring a recorded walkthrough of the changes. 

# Global Layout Refactor
image

### **TL;DR**

- Navbar is now position fixed and always mounted
- Navbar fades as you scroll down, but can be accessed via hover
- Footer is now out of view ("below the fold") on all screens, unless the user scrolls down to reveal it
- Greatly improved developer experience when setting up routes

### Wins
- The nav / footer components are much more resilient to change. We can alter their heights without worrying we broke a page we forgot about
- No need to worry about responsiveness. Before, we used to have obscene logic that adjusted the page height depending on the browser width, because being on mobile affected the height of the footer, which meant the height of the navbar had to adjust accordingly. Today, developers need to give 0 consideration for mobile/tablet.
- The overly complex `GalleryRoute.tsx` and `Page.tsx` have been deprecated

### Caveats & Fast-follows
- The animation complexity has been shifted elsewhere, but into places developers should rarely touch (vs. the original `GalleryRoute` and `Page`, which were commonly used). Every file has been thoroughly commented.
- The navbar now makes its own query on pageload, which can (and should) be combined with other queries on the page. This will be handled as a follow-up (`GAL-153`)
- The new banner appears over the navbar (instead of above it), but we'll likely want to make additional design tweaks (`GAL-154`)
- On mobile devices, scrolling up should reveal the navbar (`GAL-155`)

### ⭐️ GlobalLayoutContext
**This is the main change.** Recommend reading through the comments in `GalleryLayoutContext.tsx` to see how it works. At a high-level:
- The navbar is now rendered above the app in this new context. This allows it to stay in view as the user navigates between pages.
- All the logic that determines when and how the navbar fades in and out of view is encapsulated in this context
- The context is aware of when the app is suspending due to a route transition
- Unlike the navbar, the footer is rendered deeper down the app, alongside the route contents. This is because 1) we don't need any fancy fading for the footer, 2) its existence above the app caused jerkiness during transitions.

### Page [Deprecated]
- The main `Page` component is gone, and all of its sub-components (footer, nav, etc.) have been moved under `/GlobalLayoutContext`

### GalleryRoute
- Totally stripped down to simply dictate whether the navbar, banner, and footer should be visible or not
- It relies on `useEffect` to communicate with the upper context whether to toggle the navbar. I thought the async nature of this might cause problems, but it works great

### Route transitions
- `FadeTransitioner`, the component that manages fading the entire page in/out, has been simplified. It also exports constants to share transition animations with the navbar, since they need to work in unison
- `FullPageLoaderWithLayoutTransitionSupport` is our decorated Capital G loader. This communicates to the layout context when the app is suspending due to a route transition. It also knows when to appear over or under the navbar depending on context

### AuthProvider
- We no longer render `children` until auth state has been confirmed. Without this, the context was causing double-renders on page load, causing a vicious bug related to navbar fading.

### Dropdown
- Add `shouldCloseOnMenuItemClick` prop that allows user to close the dropdown upon clicking on an option. Without this, the dropdown would remain open across route changes.

## Tests
- [x] Visually test every single page and flow on desktop, tablet, and mobile
- [x] Visually test "tall" galleries and community pages that extend beyond the full viewport, as well as "short" ones
- [x] Ensure fade works when navigating back to a previously loaded page
- [x] Ensure fade works when navigating to a new page with unloaded content that throws to suspense
- [x] On touchscreens, ensure users can tap near the top of the device without triggering nav
- [x] Ensure `GalleryRedirect` still works
- [x] Ensure footer doesn't flash between transitions
- [x] Global loader should cover nav if it wasn't previously visible, but keep them in view if they were previously visible
- [x] Ensure user can click on elements "beneath" the navbar
- [x] Ensure user can't click "Sign In" when the navbar is invisible
- [x] Ensure dropdown closes on click if navbar stays in view


src/contexts/auth/AuthContext.tsx
```diff
@@ -236,17 +236,17 @@ const AuthProvider = memo(({ children }: Props) => {
...
-            </AuthStateContext.Provider>
-          </Suspense>
-        </ErrorBoundary>
-      </div>
+      <ErrorBoundary>
+        <Suspense fallback={<FullPageLoader />}>
+          <AuthStateContext.Provider value={authState}>
+            <AuthActionsContext.Provider value={authActions}>
+              <Web3WalletProvider>
+                {shouldDisplayUniversalLoader ? null : children}
```
Robinnnnn, line 244: main change is waiting to render children until auth state is confirmed. More details in PR description

src/contexts/boundary/ErrorBoundary.tsx
```diff
@@ -27,22 +26,30 @@ class ErrorBoundary extends Component {
...
   }
 }
 
+const StyledErrorBoundary = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction: column;
+  height: 100vh;
+`;
```
Robinnnnn, line 51: this will be a common appearance. without the `Page` component, views will now style themselves

src/contexts/navigation/GalleryNavigationProvider.tsx
```diff
@@ -21,6 +21,14 @@ type Props = {
   children: ReactNode;
 };
 
+const SCROLL_DELAY_TIME_MS = 80;
+const SCROLL_TRIGGER_TIME_MS =
+  FADE_TRANSITION_TIME_MS +
+  // this adds an arbirarily small timeframe to ensure the fade transition defined in
+  // `FadeTransitioner.tsx` has completed; without this, in rare cases, the user may
+  // see a flash of scrolling while transitioning
+  +SCROLL_DELAY_TIME_MS;
```
Robinnnnn, line 30: after navigating around the site a million times, i realized this was needed

src/scenes/CollectionGalleryPage/CollectionGalleryPage.tsx
```diff
@@ -76,19 +70,17 @@ function CollectionGalleryPage({ collectionId, username, queryRef }: CollectionG
...
         <title>{headTitle}</title>
       </Head>
-      <Page>
-        <StyledCollectionGalleryWrapper>
-          <StyledPositionedBackLink>
-            <ActionText onClick={handleBackClick}>← Back to Gallery</ActionText>
-          </StyledPositionedBackLink>
-          <CollectionGallery queryRef={query} />
-        </StyledCollectionGalleryWrapper>
-      </Page>
```
Robinnnnn, line 86: deprecated this after discussing with jess.
1. this would appear beneath the new navbar, and a fix would be non-trivial
2. the component seems a bit redundant since the user can just click back on their browser, or use the breadcrumbs on the collection title

src/scenes/Nuke/Nuke.tsx
```diff
@@ -17,16 +16,24 @@ function Nuke() {
...
       <Spacer height={32} />
       <GalleryLink to="/">
         <StyledButton text="Take me home" />
       </GalleryLink>
-    </Page>
+    </StyledNuke>
   );
 }
 
+const StyledNuke = styled.div`
```
kaitoo1, line 29: super nit - I see former `Page`s being styled with a few varying naming patterns now: `StyledHomePage`, `StyledNuke`, `StyledPage` - maybe we can use the same pattern for all of these to make it easy to quickly see which styled components are styled like pages?

src/contexts/globalLayout/GlobalLayoutContext.tsx
```diff
@@ -0,0 +1,371 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
kaitoo1, line 1: this context is brilliant 🙌🙌 
great job figuring this out! love the idea of handling everything in one place

src/contexts/globalLayout/GlobalLayoutContext.tsx
```diff
@@ -0,0 +1,371 @@
...
+        ...GlobalLayoutContextNavbarFragment
+      }
+    `,
+    {}
+  );
+
+  // whether the global banner is visible
+  const [isBannerVisible, setBannerVisible] = useState(false);
+
+  /**
```
kaitoo1, line 90: OP comments

src/contexts/globalLayout/GlobalLayoutContext.tsx
```diff
@@ -0,0 +1,371 @@
...
+  // touchscreen devices will have different behavior
+  const isTouchscreen = useRef(isTouchscreenDevice());
+
+  /**
+   * There is a race condition that arises if a scroll-based fade is triggered immediately after
+   * a route-based fade, causing the nav to appear prematurely. We use a timestamp to determine
+   * whether an animation was recently triggered by route change. (For some reason, this bug
+   * couldn't be fixed with throttling)
+   */
+  const lastFadeTriggeredByRouteTimestampRef = useRef(0);
```
kaitoo1, line 152: 🧠

src/contexts/globalLayout/GlobalLayoutContext.tsx
```diff
@@ -0,0 +1,371 @@
...
+    if (isPageInSuspenseState && isNavbarEnabled) {
+      return debounced;
+    }
+    return throttled;
+  }, [isNavbarEnabled, isPageInSuspenseState, throttled, debounced]);
+  const wasNavbarVisible = usePrevious(isNavbarVisible) ?? false;
+
+  // whether the route wants the navbar disabled. this is an override where scrolling / hovering
+  // will not make the navbar appear
+  const forcedHiddenByRouteRef = useRef(false);
```
kaitoo1, line 141: how come these are refs? I see the comments on L96-98 but I don't fully get it 🤓

I'm guessing we want to update some values without causing a re-render or other effects? but unsure what happens if we don't use ref

src/contexts/globalLayout/GlobalLayoutContext.tsx
```diff
@@ -0,0 +1,371 @@
...
+
+  /**
+   * navbar-related states.
+   *
+   * - `isNavbarEnabled` is the raw state that's set by "dumb" components / interactions
+   *   that simply wanna display / hide the nav, according to its internal logic
+   * - we then need to `debounce` or `throttle` this state because many interactions could be
+   *   fighting to set this state
+   * - the difference between debouncing vs. throttling comes down to whether we need to
+   *   delay the navbar fade animation vs. display it immediately, respectively.
```
kaitoo1, line 123: 😲😲 
the debounce vs throttle logic is super cool

kaitoo1: and by cool I mean fucking brilliant

kaitoo1 (approved):
holy F!!

gonna be honest this PR was a pleasure to read lol. especially GlobalLayoutContext - it's an absolute machine in terms of what it accomplishes and the cases it handles, but it's organized so well it looks simple. Enjoyed reading all the comments too, they made reviewing this PR like 100x easier.
Given all the different forces that affect how the navbar acts, amazing attention to detail and thoughtfulness about handling all the cases you encountered!!!

left a couple of questions for my understanding's sake and also a minor comment about naming convention

LFGGGG

src/contexts/globalLayout/GlobalLayoutContext.tsx
Robinnnnn: so i've been leaning more towards using `ref` in these complex scenarios to avoid side-effects.

**TL;DR – use `state` only when you need to**

let's flip the question: when would you use `state`? state is useful when you want to:
1. trigger a re-render if the value changes (e.g. button disabled => enabled after X change)
2. trigger a side-effect if the value changes (e.g. begin fetching data after X change)

so using `state` is very deliberate and comes with the intention of updating your react view, side effects.

when would you use a `ref`? when you explicitly _don't_ want to trigger re-renders, but still depend on that value when evaluating something.

knowing this, let's think about the purpose of `forcedHiddenByRouteRef`. what does this piece of information do? it tells us whether the route should be completely invisible, despite other factors that try to make it visible.

let's see where `forcedHiddenByRouteRef` is used:
image
image

it's used in these callbacks. when a callback is fired, e.g. `handleFadeNavbarOnScroll`, it looks up whether the navbar should be hidden all the time. if so, it opts out of displaying the navbar, even if the user scrolled to a position where it should be visible.

why is this not in `state`? because it doesn't have to be!
1. the value is being looked up whenever the callback is fired
2. the value is NOT being utilized in the render, or useEffect

in fact, if it _was_ in state, it would de-stabilize our callback and potentially cause unwanted side-effects.

```js
// if we use state
const handleScroll = useCallback(() => {
  if (forcedHiddenByRoute) { ... }
}, [forcedHiddenByRoute]) <-- a new callback would be defined
                              every time this value changes,
                              which is bad if we're putting this
                              callback inside of a document event
                              listener

// if we use ref
const handleScroll = useCallback(() => {
  if (forcedHiddenByRouteRef.current) { ... }
}, []) <-- the callback can remain pure, which is critical
           for event listeners
```

src/scenes/Nuke/Nuke.tsx
Robinnnnn: ooo yeah, great call! it almost makes me want to standardize this into a `Page` component that all views can use 😂 😂 😂

src/contexts/globalLayout/GlobalLayoutContext.tsx
Robinnnnn: ok here's a commit on another PR that's directly relevant! using `state` caused a bug / race condition so i had to switch to using ref: https://github.com/gallery-so/gallery/pull/755/commits/f813a13aa6148443da3722b14b6fef64910e5859

kaitoo1 (approved):

PR #744, opened 2022-05-09
[GAL-148] Fix tracing for logins

Robinnnnn:
Tracing used to be broken at the final login step because we fetch user info imperatively via Queries.

**Before**
Screen Shot 2022-05-09 at 12 09 34 PM

**After**
image


src/contexts/auth/AuthContext.tsx
```diff
@@ -161,18 +162,32 @@ const AuthProvider = memo(({ children }: Props) => {
   const handleLogin = useCallback(
     async (userId: string, address: string) => {
       try {
-        // TODO__GRAPHQL: explicitly check error types (instead of try/catch)
```
Robinnnnn, line 164: bonus: handled this TODO

kaitoo1 (approved):

PR #745, opened 2022-05-10
Add unit test for timer hooks

jakzaizzat:
Added test scenario to cover `useTimer` hooks

1. Make sure its render the timestamp
2. If its the past date, it should return `hasEnded` false
3. if one hour left, it should correct timestamp and not ended
4. if one minute left, it should correct timestamp and not ended

### DIFF
--- a/src/constants/poster.ts
+++ b/src/constants/poster.ts
@@ -1,0 +1,1 @@
+export const mintDate = '2022-05-23T00:00:00';
--- a/src/hooks/useTimer.test.tsx
+++ b/src/hooks/useTimer.test.tsx
@@ -1,14 +1,6 @@
 import { renderHook } from '@testing-library/react-hooks';
 import useTimer from './useTimer';
 
-function convertToSeconds(timestamp: string): number {
-  const [days, hours, minutes, seconds] = timestamp.split(':');
-
-  return (
-    Number(days) * 24 * 60 * 60 + Number(hours) * 60 * 60 + Number(minutes) * 60 + Number(seconds)
-  );
-}
-
 describe('useTimer hooks', () => {
   test('render timestamp', () => {
     const now = new Date();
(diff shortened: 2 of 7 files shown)


src/hooks/useTimer.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const date = new Date(new Date().getTime() + minute * 60000).toString();
+    const { result } = renderHook(() => useTimer(date));
+    const { timestamp, hasEnded } = result.current;
+
+    const totalSeconds = convertToSeconds(timestamp);
+
+    expect(totalSeconds).toBeLessThanOrEqual(minute * 60);
+    expect(hasEnded).toBe(false);
+  });
+});
```
Robinnnnn, line 55: 🔥

src/hooks/useTimer.ts
```diff
@@ -1,10 +1,8 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
-// TODO: Abstract this as a props
-const mintDate = new Date('2022-05-23T00:00:00');
-
-export default function useTimer() {
+export default function useTimer(date: string) {
```
Robinnnnn, line 3: question: is this function deterministic?

in other words, if i pass in a hard-coded `date`, will it always return the same `timestamp` and `hasEnded` output? my guess is no, because this hook calls `Date.now()`, which means it depends on when the test runs...

i wonder if this would be easier to test if we passed in both `mintDate` and `currentTime`:
```js
useTimer(mintDate, currentTime) {}
```

then we know what the initial output should be, and directly compare strings, like:
```js
expect(timestamp).toEqual("00:12:34:56")
// simulate moving forward 1 second
expect(timestamp).toEqual("00:12:34:55")
// simulate moving forward 24 hours
expect(hasEnded).toEqual(true)
```

src/hooks/useTimer.ts
```diff
@@ -1,10 +1,8 @@
...
   const currentTime = Date.now();
+  const mintDate = useMemo(() => new Date(date), [date]);
```
kaitoo1, line 5: since this hook is a generic timer hook, let's rename `mintDate` to something more generic too - maybe `endDate`?

src/constants/poster.ts
```diff
@@ -0,0 +1 @@
+export const mintDate = '2022-05-23T00:00:00';
```
kaitoo1, line 1: nit: can we rename this to all caps `MINT_DATE` ? it's a small pattern we prefer that makes it easy to identify consts where they're used!

kaitoo1:
Nice @jakzaizzat ! thanks for following up, great to make the timer reusable 🚀
Agree with @Robinnnnn 's comments, and left 2 minor comments about variables. Otherwise looks good!

In terms of merging/releasing, we're planning to launch the Poster Page Tues Morning (5/10) US morning. It's soon and we won't block the release with this PR so we can plan to merge this whenever its ready.
When we do, let's just make sure the MINT_DATE const matches the value that'll be on main. (it'll most likely be `2022-05-19T00:00:00`)

src/hooks/useTimer.ts
jakzaizzat: Agree. Changed to `endDated` 🙌🏻

src/constants/poster.ts
jakzaizzat: Sure! Should consistent with other constant

src/hooks/useTimer.ts
jakzaizzat: Make sense 🤔 Definetely much easier to test this way. 

I've refactor the `useTimer` hooks to support `currentTime` as a second params (optional). If its not available, will set to current date

Thanks for the string comparison idea! Felt much confident with it 😆

Robinnnnn (approved):
woooo! thanks for addressing, jakz!

PR #746, opened 2022-05-10
Kaito/community page previews

kaitoo1:


src/components/TokenHolderList/TokenHolderList.tsx
```diff
@@ -0,0 +1,80 @@
+import { useMemo } from 'react';
```
kaitoo1, line 1: this file was only renamed, it used to be for the Membership Directory. Now it can be used to render a list of `TokenHolder`s and therefore shared between Membership Directory + Community Page

src/components/TokenHolderList/TokenHolderList.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+// Get which side of the tokenHolder name to show the preview on
+// 1st and 2nd column should be right, 3rd and 4th column should be left
+function getPreviewDirection(index: number) {
+  return index % 4 < 2 ? Directions.RIGHT : Directions.LEFT;
+}
+
+type Props = {
+  title: string;
+  tokenHoldersRef: MemberListTierFragment$data['owners'] | CommunityPageViewFragment$data['owners'];
```
kaitoo1, line 20: This component is used by both the Community Page and Membership Directory Page, which have query fragments on different types: `Community` and `MembershipTier`. Both of these types have an `owners` field, which is a list of `TokenHolder`s. 

Passing in this common field, `owners` rather than the parent fragment satisfied the type requirements and allowed me to share the `TokenHolderListItemFragment` child fragment for both the Membership Directory and Community queries.

Robinnnnn: sounds good to me! @tbezman is this generally best practice?

PR #748, opened 2022-05-10
[GAL-38] Implement Markdown shortcut buttons

connorrothschild:
## Implement Markdown shortcut buttons

This PR creates buttons for three Markdown formatting options: bolding text, creating list items, and adding links.

The Markdown shortcut buttons are present in 1. the edit user info modal and 2. the collector's note textarea.

https://user-images.githubusercontent.com/13339581/167713041-2c018f02-8722-43e8-b08f-ca7b5ae69295.mov

Behind the scenes, [`<MarkdownShortcuts />`](https://github.com/gallery-so/gallery/compare/connor/markdown?expand=1#diff-f11b2ab46ee976e915f2a98f7be27d477e1e3ad226998aaa4fd9c0fcbd03f6c8) is handling the 
1. **Text selection**, e.g. where your cursor is, and what it is highlighting, and 
2. **Change events**, e.g. it creates a `CustomEvent` that can be passed to `onChange` so that the addition/removal of Markdown is rendered as a change event in the same way that editing text would be.

Then, each of the Markdown options are in their own components: `<Bold />`, `<List />`, and `<Link />`. These are a bit complex because they handle a lot of different cases. [`<List />`](https://github.com/gallery-so/gallery/compare/connor/markdown?expand=1#diff-294b5c2c4791a4d295ad7c7ff5ef2d0dd87dd1f2d579481c10a0c4e953c0f7ac) in particular is sort of a beast. (Yet again illustrating why as much testing as possible would be appreciated 🙏)

Two things that would be helpful before merging: 

### **Try to break things** 

Each of the three Markdown Shortcuts are pretty complicated in nature, so as much testing as you can do to break things will be appreciated. You can play around with the shortcuts in your collectors notes or edit user info modal. 

### **Help me fix my Typescript errors pls** 

There are two in particular: 
1. Getting the addition of Markdown to trigger `onChange` (from the parent) was complicated enough, because it didn't trigger a `ChangeEvent` on the `textarea`. I ended up creating a `CustomEvent` that mimics that can be passed into `onChange` so everything works, but now I'm getting a Typescript error. (This is my first time using `CustomEvent`, could def move away from this solution if anyone has ideas). This leads to a TypeScript error [here](https://github.com/gallery-so/gallery/compare/connor/markdown?expand=1#diff-f11b2ab46ee976e915f2a98f7be27d477e1e3ad226998aaa4fd9c0fcbd03f6c8R29).
2. Passing the `textAreaRef` to `<MarkdownShortcuts />`. There is some issue here with passing a ref from a `forwardRef`. It is pretty confusing. This leads to a TypeScript error [here](https://github.com/gallery-so/gallery/compare/connor/markdown?expand=1#diff-e2b5bbd1c1e6855e61b21e0dcd0343a3d6fe6a03da6b6df34952292da9163667R55).


src/components/core/Markdown/IconContainer.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import styled from 'styled-components';
+
+export default function IconContainer({ icon }: { icon: React.ReactElement }) {
+  return (
+    <StyledIcon
+      onMouseDown={(e) => {
+        // This will prevent the textarea from losing focus when user clicks a markdown icon
```
kaitoo1, line 8: 🧠

src/components/core/Markdown/Bold.tsx
```diff
@@ -0,0 +1,87 @@
...
+      const selectedText = textArea.value.substring(start, end);
+      const selectedTextWithSurrounding = textArea.value.substring(start - 2, end + 2);
+
+      const selectedTextIsBold = selectedText.startsWith('**') && selectedText.endsWith('**');
+      const selectedTextIsSurroundedBold =
+        selectedTextWithSurrounding.startsWith('**') && selectedTextWithSurrounding.endsWith('**');
+
+      // If the selected text includes any bold tags, remove them
+      if (selectedTextIsBold) {
+        const newText = textArea.value.replace(selectedText, selectedText.slice(2, -2));
```
kaitoo1, line 30: it looks like this'll replace the first occurrence of `selectedText`, so if there's more than one instance of the same bold text, it may cause the 1st one to be bolded when you meant to unbold the 2nd

kaitoo1: for example if the text was 

`**this is bold**

**this is bold**`

and you highlighted the second `**this is bold**` and click the bold button

kaitoo1: an alternative could be to use the start/end indices to replace the exact area of the text instead

src/components/core/Markdown/Bold.tsx
```diff
@@ -0,0 +1,87 @@
...
+        textArea.value = newText;
+        setSelectedRange([start, end - 4]); // **Bold** ([0, 8]) -> Bold ([0, 4])
+        return;
+      }
+      if (selectedTextIsSurroundedBold) {
+        const newText = textArea.value.replace(
```
kaitoo1, line 36: same as above

src/components/core/Markdown/Bold.tsx
```diff
@@ -0,0 +1,87 @@
...
+          selectedTextWithSurrounding,
+          selectedTextWithSurrounding.slice(2, -2)
+        );
+        textArea.value = newText;
+        setSelectedRange([start - 2, end - 2]); // **Bold** ([2, 6]) -> Bold ([0, 4])
+        return;
+      }
+      if (selectedText.length > 0) {
+        // If the selected text is not already bold, add it
```
kaitoo1, line 45: thanks for these comments, really helps with following the logic! and great variable names too

src/components/core/Markdown/Bold.tsx
```diff
@@ -0,0 +1,87 @@
...
+        const newText =
+          textArea.value.substring(0, start) +
+          `**${selectedText}**` +
```
kaitoo1, line 48: since we're using template literals for `selectedText`, maybe we could use it for the whole value of `newText`?

or does it not fit in one line and look ugly 👀

src/components/core/Markdown/IconContainer.tsx
```diff
@@ -0,0 +1,33 @@
...
+        e.preventDefault();
+      }}
+    >
+      {icon}
+    </StyledIcon>
+  );
+}
+
+const StyledIcon = styled.div`
+  // margin-right: 4px;
```
kaitoo1, line 18: nit: can remove?

src/components/core/Markdown/Link.tsx
```diff
@@ -0,0 +1,73 @@
...
+        return;
+      }
+      if (selectedText.length > 0) {
+        // If the selected text is not already a link, add it
+        const newText =
+          textArea.value.substring(0, start) +
+          `[${selectedText}](https://)` +
+          textArea.value.substring(end);
+        textArea.value = newText;
+        setSelectedRange([start + 11 + selectedText.length, start + 11 + selectedText.length]); // Link [0, 4] -> [Link](https://) (end)
```
kaitoo1, line 34: love that the cursor ends up right at the beginning of the url, ready for you to type the rest - super useful!

src/components/core/Markdown/List.tsx
```diff
@@ -0,0 +1,123 @@
...
+  selectedRange: number[];
+  textAreaRef: React.RefObject<HTMLTextAreaElement>;
+  setSelectedRange: (value: number[] | ((prevVar: number[]) => number[])) => void;
+  setUserDragged: (value: boolean) => void;
+}) {
+  const handleClick = useCallback(() => {
+    setUserDragged(false);
+
+    const textArea = textAreaRef.current;
+    if (textArea) {
```
kaitoo1, line 19: same comment about returning early

src/components/core/Markdown/List.tsx
```diff
@@ -0,0 +1,123 @@
...
+      const lineStartIndices = allLines.map((line, index) => {
+        // if (index === 0) {
+        //   return 0;
+        // }
+        // Return the sum of each line's text length up until the current line
+        return allLines.slice(0, index).reduce((acc, curr) => acc + curr.length + 1, 0);
+      });
+
+      // Map over allLines and return an object with the existing string and a boolean indicating if it is selected
+      const allLinesWithSelected = allLines.map((line, index) => {
```
kaitoo1, line 35: this is sick 😮

src/components/core/Markdown/List.tsx
```diff
@@ -0,0 +1,123 @@
...
+    const textArea = textAreaRef.current;
+    if (textArea) {
+      const [start, end] = selectedRange;
+      const selectedText = textArea.value.substring(start, end);
+
+      let allLines = textArea.value.split('\n');
+
+      // Map over each line and store in an array the number of characters in the string up until that point
+      const lineStartIndices = allLines.map((line, index) => {
+        // if (index === 0) {
```
kaitoo1, line 27: nit can remove?
then can remove the `return` from the arrow func

src/components/core/Markdown/Link.tsx
```diff
@@ -0,0 +1,73 @@
...
+    if (textArea) {
+      const [start, end] = selectedRange;
+      const selectedText = textArea.value.substring(start, end);
+
+      // If user is inbetween two brackets, they probably just clicked the link button. Do nothing
+      if (textArea.value.substring(start - 1, end + 1) == '[]') {
+        return;
+      }
+      if (selectedText.length > 0) {
+        // If the selected text is not already a link, add it
```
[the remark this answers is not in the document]
connorrothschild: Good catch, this was not the intention, I've updated the comment to reflect

src/components/core/Markdown/List.tsx
```diff
@@ -0,0 +1,123 @@
...
+  textAreaRef,
+  setSelectedRange,
+  setUserDragged,
+}: {
+  selectedRange: number[];
+  textAreaRef: React.RefObject<HTMLTextAreaElement>;
+  setSelectedRange: (value: number[] | ((prevVar: number[]) => number[])) => void;
+  setUserDragged: (value: boolean) => void;
+}) {
+  const handleClick = useCallback(() => {
```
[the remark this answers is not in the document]
connorrothschild: @kaitoo1 I agree this would be a good candidate for testing. Happy to talk through next steps & the best way to do that

connorrothschild:
@kaitoo1 These comments have been addressed and the bugs you mentioned in Threads are also accounted for 👍 

We can do some more stress testing, and then could we look into the TypeScript errors I mentioned? Still perplexed by those

src/components/core/Markdown/Bold.tsx
```diff
@@ -0,0 +1,91 @@
+import IconContainer from './IconContainer';
+import { useCallback } from 'react';
+
+export default function Bold({
```
Robinnnnn, line 4: fucking love how this is set up

src/components/core/Markdown/List.tsx
```diff
@@ -0,0 +1,119 @@
+import IconContainer from './IconContainer';
+import { useCallback } from 'react';
+
+export default function Bold({
```
Robinnnnn, line 4: guessing you meant to call this `List` haha

src/components/core/Markdown/List.tsx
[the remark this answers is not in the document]
Robinnnnn: +1, could look something like this:
1. render a text area that renders this List component (or bold, etc.)
2. simulate interactions with the component
3. test the result

found this resource: https://testing-library.com/docs/ecosystem-user-event/#typeelement-text-options

image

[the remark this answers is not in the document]
Robinnnnn: overall i don't think testing needs to block this PR if we have a good sense of it working for most cases.

@connorrothschild wanna give it a whirl on your own by going through the docs, and afterwards we can pair some time next week?

in our codebase i would:
1. make a test file next to the component file (e.g. `Bold.test.tsx` next to `Bold.tsx`)
2. define a test. you can use the format we use in other tests, like [this one](https://github.com/gallery-so/gallery/blob/main/src/contexts/collectionEditor/deduplicateObjectByOpenseaIdAndPreferEarliest.test.ts)
3. to run just the test you defined in isolation, you would do: `yarn jest <name of test>`. so if your test suite is `describe('Bold', () => ...)` you should be able to run `yarn jest Bold`



yarn jest deduplicateObjectByOpenseaIdAndPreferEarliest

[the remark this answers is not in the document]
Robinnnnn: oh also make sure to name the file with `.test.tsx` instead of `.test.ts`, bc it'll be rendering JSX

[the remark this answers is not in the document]
Robinnnnn: thinking about this further- i actually think you just need to write one overarching test for `MarkdownShortcuts.test.tsx`, since it encompasses all the different kinds of editing (instead of 3 different test files). this is more integration-y and preferred

Robinnnnn:
connor, **incredible work** so far, holy crap 💪 💪 💪 

the functionality is fantastic. i couldn't break the actual logic of the text area!!

gonna leave some purely design feedback:

1) looks like the inputs now have a new off-white background color that we should be using, both for the header as well as the text area. the button also looks bigger in the spec

| this PR | spec |
| ----- | ----- |
| image | image |

2) same comment as above for the collector's note. also note that the save button now says `SAVE NOTE`
image

3) the icons should be vertically and horizontally the same distance away from the edges

| this PR | spec |
| ----- | ----- |
| Screen Shot 2022-05-13 at 7 45 09 PM | image |

other than that, insaneeee work

Robinnnnn:
ah, i think the end-result needs a design pass as well.

putting the following in:
image

results in this, for bio the "bio" field and the "collection description" fields:
image

src/components/core/Markdown/MarkdownShortcuts.tsx
```diff
@@ -0,0 +1,101 @@
...
+        detail: { value: textAreaRef.current.value },
+      });
+      Object.defineProperty(event, 'target', {
+        value: textAreaRef.current,
+      });
+      // textAreaRef.current.dispatchEvent(event);
+      // onChange(event as React.ChangeEvent<HTMLTextAreaElement>);
+      onChange(event);
+    }
+  }, [textAreaRef?.current?.value, onChange, textAreaRef]);
```
Robinnnnn, line 31: so i read your PR description but i'm a bit confused as to why this is necessary 🤔 

if i understand this correctly, it's saying "if the text area's value changes, fire the onChange event".

is there a reason we can't do this from the parent?

e.g.
```js
// TextArea.tsx

    const textAreaRef = ref || _ref;

    const handleChange = useCallback((event) => {
      const event = new CustomEvent(...)
      onChange(event)
    }, [onChange])

    return (
      <>
        <StyledTextArea
          onChange={handleChange}
```

i'm probably missing something!

Robinnnnn: re: typescript, i think you just have to cast it the way you did with `as`

Robinnnnn: ahhh i think i get it. basically the markdown causes changes to the textArea ref... but that isn't necessarily picked up by the parent, so the parent wouldn't necessarily call `onChange`. is that right?

Robinnnnn:
okay LAST comment for the night-

the forwardRef typescript thing should be fixed by this: https://github.com/gallery-so/gallery/pull/752

it was explained from here: https://stackoverflow.com/questions/65876809/property-current-does-not-exist-on-type-instance-htmldivelement-null

i made the PR against this branch, so if it looks good to you, feel free to merge!

src/components/core/Markdown/MarkdownShortcuts.tsx
connorrothschild: Yep, exactly! Here's an example using the code you sent above:

https://user-images.githubusercontent.com/13339581/168489104-afe00170-ef0c-4c55-940b-8e7faf65c03a.mov

So this is the best way to handle it that I can think of

connorrothschild: @Robinnnnn when I pass `event as React.ChangeEvent<HTMLTextAreaElement>` I get the following:

```
Conversion of type 'CustomEvent<{ value: string; }>' to type 'ChangeEvent<HTMLTextAreaElement>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'CustomEvent<{ value: string; }>' is missing the following properties from type 'ChangeEvent<HTMLTextAreaElement>': nativeEvent, isDefaultPrevented, isPropagationStopped, persistts(2352)
```

We can fix this by instead passing `event as unknown as React.ChangeEvent<HTMLTextAreaElement>`... Feeling sort of hacky but I'm stuck here :/

kaitoo1: good news - took a look at this today and think I found a way to get it to work without the useEffect here!

it appears like the underlying issue is using `textArea.value = newText` to manually update the value. This doesn't trigger the event and onChange handler, as you noted.

I got textArea to detect that its value changed by using the native value setter method instead, then dispatching the input event manually, like so:

Screen Shot 2022-05-16 at 20 03 04

[Reference 1](https://stackoverflow.com/a/61110332) [Reference 2](https://stackoverflow.com/a/46012210)

This triggered the TextArea's onChange handler without the useEffect as desired.

I also put this logic in a helper function called `setValueAndTriggerOnChange`. This way we can do:
`setValueAndTriggerOnChange(textArea, newText)
instead of
`textArea.value = newText;`
whenever we manually set the text area's value (in Bold.tsx, List.tsx, etc) and it automatically gets detected by the onChange handler.

lmk what you think! I think this moves the fix closer to the source of the issue, and leaves us with 1 less effect to think about.

Bonus - I noticed that with the original useEffect implementation, it worked correctly for Markdown Shortcut clicks, but the onChange handler was triggered twice for normal keypresses - I believe once by the TextArea's onChange and once by the useEffect, so this eliminates that too 

I also tweaked the stackoverflow example to remove `window` from `window.HTMLTextAreaElement.prototype` since we're trying to avoid adding usages of `window` for SSR purposes - looks like it still works but might be worth testing a bit more.

I opened a PR with these changes against this branch, #753 , feel free to refer to it or just merge it! I only applied it to `Bold.tsx` so you'll have to add /adapt it to List and Links.

Robinnnnn:
@connorrothschild wanna slap on a feature flag so we can merge?!

connorrothschild:
@Robinnnnn yessssssir i totally forgot, on it

Robinnnnn (approved):

kaitoo1 (approved):
Awesome work @connorrothschild! this was a great PR. thanks for addressing all the feedback! works super well for me, looking forward to telling users about it

PR #749, opened 2022-05-12
make collection name in hover text smaller on mobile

kaitoo1:
On the label that appears when you hover on an NFT, the collection name was a bit bigger than the name name on mobile.
This PR fixes that so that the sizing is consistent.

Before:
Screen Shot 2022-05-12 at 17 02 49

After:
Screen Shot 2022-05-12 at 17 02 43


Robinnnnn (approved):
ty @kaitoo1!

LGTM, but i'm wondering if we should make the text even smaller on mobile? since it's easy to accidentally click into the collection

kaitoo1:
yeah we definitely should, or come up with a different way altogether to display this info on mobile - I don't think there's a font size that would be realistically small enough but still readable
will bring up with jess

PR #753, opened 2022-05-16
alternate solution for triggering onChange when manually setting textarea value

kaitoo1:
alternate solution for triggering onChange when manually setting textarea value as discussed in #748

### DIFF
--- a/src/components/core/Markdown/Bold.tsx
+++ b/src/components/core/Markdown/Bold.tsx
@@ -1,5 +1,6 @@
 import IconContainer from './IconContainer';
 import { useCallback } from 'react';
+import { setValueAndTriggerOnChange } from './MarkdownShortcuts';
 
 export default function Bold({
   selectedRange,
@@ -33,7 +34,7 @@ export default function Bold({
         textArea.value.substring(0, start) +
         selectedText.slice(2, -2) +
         textArea.value.substring(end);
-      textArea.value = newText;
+      setValueAndTriggerOnChange(textArea, newText);
       setSelectedRange([start, end - 4]); // **Bold** ([0, 8]) -> Bold ([0, 4])
       return;
     }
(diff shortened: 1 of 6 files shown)


src/components/core/TextArea/TextArea.tsx
```diff
@@ -125,6 +125,7 @@ export function AutoResizingTextAreaWithCharCount({
 
   const handleChange = useCallback(
     (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+      console.log('handling');
```
kaitoo1, line 128: left this in here to make it easier to validate the behavior. can remove before merging

src/components/core/Markdown/Bold.tsx
```diff
@@ -33,7 +34,7 @@ export default function Bold({
         textArea.value.substring(0, start) +
         selectedText.slice(2, -2) +
         textArea.value.substring(end);
-      textArea.value = newText;
+      setValueAndTriggerOnChange(textArea, newText);
```
kaitoo1, line 37: will need to do this for List and Link I imagine

src/components/core/Markdown/MarkdownShortcuts.tsx
```diff
@@ -1,30 +1,34 @@
...
-  onChange, // The onChange from the TextArea component. This ensures that we track the same event handlers (e.g. a collector's note change) via the addition of just markdown
-}: {
+// Use the TextArea element's native setter and dispatch an input event to trigger the onChange callback
+export function setValueAndTriggerOnChange(textArea: HTMLTextAreaElement, newValue: string) {
+  var nativeTextAreaValueSetter = Object.getOwnPropertyDescriptor(
+    HTMLTextAreaElement.prototype,
+    'value'
+  )?.set;
+
+  if (!nativeTextAreaValueSetter) {
```
kaitoo1, line 16: added this condition to satisfy typescript - nativeTextAreaValueSetter could be undefined.
added to sentry because I'm curious if this will happen in the wild

Robinnnnn: what is this sorcery?!?!?

PR #759, opened 2022-05-20
[GAL-50 GAL-51 GAL-52 GAL-53] WIP Add Follow functionality

kaitoo1:
### This PR adds Follow/Unfollow functionality.
Users can now follow + unfollow other users, and also see a list of followers and following.

CleanShot 2022-05-24 at 00 48 30

### Main new components are:
**Follow Button** 
displays an icon depending on following state, and calls the follow/unfollow mutation. has some icon display logic depending on hover, etc.

**Tooltip**
 replaces react-tooltip. Making our own tooltip was easier to match the spec 

**Follower List component**
 displays a user's followers and following.

**Follower Count** 
displays a user's follower count + opens Follower List when clicked.

**A  Follow "Nav Action" component** 
 wraps all the above into an interactive component that lives in the navbar. Other Nav Actions would be additional circular buttons in the navbar.

+ added useFollowUser and useUnfollowUser mutation hooks


src/components/Follow/FollowButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const followUser = useFollowUser();
+  const unfollowUser = useUnfollowUser();
+  const { pushToast } = useToastActions();
+
+  const handleFollowClick = useCallback(() => {
+    const optimisticNewFollowersList = [{ id: loggedInUserId }, ...user.followers];
+    followUser(user.dbid, optimisticNewFollowersList, user.following).then(() =>
+      pushToast({ message: `You have followed ${user.username}.` })
+    );
```
Robinnnnn, line 42: nit: i think we can play around with `optimisticUpdater` so that we don't have to pass in the optimistic new followers list, but rather handle the append directly on the mutation: https://relay.dev/docs/guided-tour/updating-data/graphql-mutations/#optimistic-updater

so we'd use something like the following, and access our `store` to get the current followers, etc. i'll ask terence about what's ideal!

image

src/components/Follow/FollowButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }, [followUser, pushToast, loggedInUserId, user]);
+
+  const handleUnfollowClick = useCallback(() => {
+    const optimisticNewFollowersList = user.followers.filter(
+      (follower: { id: string }) => follower.id !== loggedInUserId
+    );
+    unfollowUser(user.dbid, optimisticNewFollowersList, user.following).then(() =>
+      pushToast({ message: `You have unfollowed ${user.username}.` })
+    );
```
Robinnnnn, line 51: same nit using async/await

src/components/Follow/FollowButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }, [unfollowUser, pushToast, loggedInUserId, user]);
+
+  const handleClick = useCallback(() => {
+    isFollowing ? handleUnfollowClick() : handleFollowClick();
+  }, [handleFollowClick, handleUnfollowClick, isFollowing]);
```
Robinnnnn, line 56: supernit: because the functions are memoized, i don't think we need to memoize here!

```js
const handleClick = isFollowing ? handleUnfollowClick : handleFollowClick
```

src/components/Follow/FollowButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+    () => isAuthenticatedUsersPage || !loggedInUserId,
+    [isAuthenticatedUsersPage, loggedInUserId]
+  );
+
+  const tooltipText = useMemo(() => {
+    if (!loggedInUserId) {
+      return 'Please sign in to follow.';
+    }
+    return isFollowing ? 'Unfollow' : 'Follow';
+  }, [isFollowing, loggedInUserId]);
```
Robinnnnn, line 69: nit: it's generally unnecessary to memo primitives! better reserved for complex calculations or stabilizing objects/arrays. so `isFollowActionDisabled` can drop the memo:

```js
const isFollowActionDisabled = isAuthenticatedUsersPage || !loggedInUserId
```

really good use of memo for `tooltipText`, since that actually has more complex logic!

src/components/Follow/FollowList.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { useIsMobileOrMobileLargeWindowWidth } from 'hooks/useWindowSize';
+import { useState } from 'react';
+import { graphql, useFragment } from 'react-relay';
+import styled from 'styled-components';
+
+type Props = {
+  userRef: any;
+};
+
+const getFirstLine = (text: string) => (text ? text.split('\n')[0] : '');
```
Robinnnnn, line 15: nit: even if `text` is empty, it'll return an empty string

image

```js
const getFirstLine = (text: string) => text.split('\n')[0]
```

but if the text param is _optional_ `(text?: string)` then your solution makes sense!

src/components/Follow/FollowList.tsx
```diff
@@ -0,0 +1,137 @@
...
+        {userList.length === 0 && (
+          <StyledEmptyList>
+            <BaseM>
+              {displayedList === 'followers' ? 'No followers yet.' : 'Not following anyone yet.'}
+            </BaseM>
+            <Spacer height={48} />
+          </StyledEmptyList>
+        )}
+      </StyledList>
+    </StyledFollowList>
```
Robinnnnn, line 79: this component is cleannnn

src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const DisplayedIcon = useMemo(() => {
+    if (isFollowing) {
+      if (isHovering) {
+        return StyledUnfollowIcon;
+      }
+      return FollowingIcon;
+    }
+    return StyledFollowIcon;
+  }, [isFollowing, isHovering]);
+  console.log('disabled', disabled);
```
Robinnnnn, line 34: toss

kaitoo1:
closing in favor of #787

PR #764, opened 2022-05-23
add tests for follow

kaitoo1:


src/components/Follow/NavActionFollow.test.tsx
```diff
@@ -73,13 +74,15 @@ test.only('it works', async () => {
...
-    </RelayProvider>
+    <ToastProvider>
+      <RelayProvider>
+        <ModalProvider>
+          <Suspense fallback={null}>
+            <Fixture />
+          </Suspense>
+        </ModalProvider>
+      </RelayProvider>
+    </ToastProvider>
```
daniloab, line 85: Maybe creating a component to have all providers and then importing it for each test could be a cool idea and help you.

Once having this component you can start to import it in each test and avoid duplicated code.

```tsx
export const withProviders =
  ({
    environment = Environment,
    initialEntries = ['/'],
    path = '/',
    Component,
    initialState,
    routes,
    history,
    locationState,
  }: WithProviders) =>
  (props: any) => {
    const defaultRoutes = [
      {
        name: 'test',
        path,
        component: Component,
        exact: true,
        routes: props.routes || [],
        modalRoutes: props.modalRoutes || [],
      },
    ];

    const testRoutes = routes || defaultRoutes;

    const defaultHistory = createMemoryHistory({
      initialEntries,
    });

    if (locationState) {
      defaultHistory.push(initialEntries[0], locationState);
    }

    const testHistory = history || defaultHistory;

    return (
      <RelayEnvironmentProvider environment={environment}>
        <Providers
          environment={environment}
          store={createStore(reducer, initialState)}
        >
          <SidebarProvider>
            <Router history={testHistory}>
              <QueryParamProvider ReactRouterRoute={Route}>
                {drawRoutes(testRoutes, null, props)}
              </QueryParamProvider>
            </Router>
          </SidebarProvider>
        </Providers>
      </RelayEnvironmentProvider>
    );
  };
```
then, on the test just import it like this below:
```tsx
it('should render relay component', async () => {
  const environment = createMockEnvironment();
  const Root = withProviders({
    Component: MyReactComponent,
    environment,
  });

  render(<Root />);

  const customMockResolvers = {
    Company: () => ({
      id: `companyId`,
      lang: 'en',
      timezone: 'America/Sao_Paulo',
    }),
  };

  // Providers Query
  act(() => {
    environment.mock.resolveMostRecentOperation((operation) =>
      MockPayloadGenerator.generate(operation, customMockResolvers),
    );
  });

  // MyReactComponent
  act(() => {
    environment.mock.resolveMostRecentOperation((operation) =>
      MockPayloadGenerator.generate(operation, customMockResolvers),
    );
  });

  await waitFor(() => {
    expect(screen.queryByLabelText('YourExpectedLabel')).toBeNull();
  });
});
```

kaitoo1: hey @daniloab  thanks for the suggestion! totally agree that wrapping the providers and making them reusable is the way. we did decide against using a mock relay environment so the test will probably look a bit different

daniloab: looks good!

daniloab: could I ask why you decided to go against the mock relay environment?

tbezman: Those `withProviders` type functions really bother me. They're just react components, is there a compelling argument to treat them any different?

And we decided against mock relay environment for a few reasons
1. Type Safety
2. Knowing when to remove fields (sort of related to type safety)
3. Closeness to what actually happens in production
4. Relay test utils aren't where I'd like them to be today

daniloab: when I want to test if a list of components renders normally I don't want to care or need to mock the whole providers that my components use, or even mock the router stuff. Using with providers I can have sure that once having this already configured every test will have the normal behavior.

tbezman: > yep, they just react components to help to avoid duplication of code when do you have many providers.

Agreed, the way yours is written leads me to believe it's not being treated like a component (creating a new `history` every render, camelCased component name)

> Also, the mock environment works fine for UI here in our case.

I'd like to be consistent across tests where possible. I think we already did the work to avoid the relay mock environment

> when I want to test if a list of components renders normally I don't want to care or need to mock the whole providers that my components use, or even mock the router stuff. Using with providers I can have sure that once having this already configured every test will have the normal behavior.

totally agree, I just think that wrapper should behave just like a regular component. 👍

daniloab: > > yep, they just react components to help to avoid duplication of code when do you have many providers.
> 
> Agreed, the way yours is written leads me to believe it's not being treated like a component (creating a new `history` every render, camelCased component name)
> 
> > Also, the mock environment works fine for UI here in our case.
> 
> I'd like to be consistent across tests where possible. I think we already did the work to avoid the relay mock environment
> 
> > when I want to test if a list of components renders normally I don't want to care or need to mock the whole providers that my components use, or even mock the router stuff. Using with providers I can have sure that once having this already configured every test will have the normal behavior.
> 
> totally agree, I just think that wrapper should behave just like a regular component. 👍

Totally agreed with you too. Maybe, I ain't make myself clear here. We use the mock environment but mocking the type necessary for the test.

Example: list of customers. I will mock my `CustomerConnection: () => connectionFromArray([])` to test each case that I want. An empty connection, connection with one item only, etc. I believe that in the final it's closely with actually you are doing.

Anyway, congratz on the project. 👍

PR #775, opened 2022-05-25
Some bugfixes

Robinnnnn:
Comments in files

### DIFF
--- a/src/components/EnsOrAddress.tsx
+++ b/src/components/EnsOrAddress.tsx
@@ -1,5 +1,6 @@
 import { Suspense } from 'react';
 import useSWR from 'swr';
+import Spacer from './core/Spacer/Spacer';
 import { PlainErrorBoundary } from './PlainErrorBoundary';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


src/components/EnsOrAddress.tsx
```diff
@@ -21,7 +22,13 @@ const EnsName = ({ address }: Props) => {
 };
 
 export const EnsOrAddress = ({ address }: Props) => (
-  <Suspense fallback={null}>
+  <Suspense
+    fallback={
+      // TODO: fallback that takes the height of the text element would appear post-fetch.
+      // long-term, we're going to holistically re-think loading states on the app
+      <Spacer height={18} />
```
Robinnnnn, line 29: instead of `null`, have the fallback be a placeholder the height of the text

src/contexts/modal/ModalContext.tsx
```diff
@@ -109,6 +113,15 @@ function ModalProvider({ children }: Props) {
...
   );
 
+  // close modal on route change
+  const route = useStabilizedRouteTransitionKey();
+  useEffect(() => {
+    if (isModalOpenRef.current) {
+      // bypass onClose as to not navigate the user back mid-route change
+      hideModal(true);
+    }
+  }, [route, hideModal]);
```
Robinnnnn, line 123: 1. hide modal when navigating to a new page. should be default behavior, since keeping a modal open while re-routing is weird. this fixes a related error where going to the Community Page from the detail modal crashed, because the modal remained open, and it was looking at URL params to continue to display NFT detail stuff.
2. also added a `bypassOnClose` prop to simply hide the modal without triggering the `onClose`. this was because the NFT Detail Modal was programmed to go "back" when it was closed, which collided with the redirect to the community page

kaitoo1 (approved):

PR #802, opened 2022-06-02
[GAL-201] make spacing between collection title+desc on gallery page 4px

kaitoo1:
make spacing between collection title+desc on gallery page 4px


Robinnnnn (approved):
LGTM- should we send this up?

edit: checked in with jess and she said it looks good! thanks kaito!

PR #807, opened 2022-06-04
[GAL-203] Implement Profile QR Code

connorrothschild:
This PR creates QR code functionality. On mobile screens, a user's profile will show a QR code icon next to the mobile layout toggle button; when clicked, it will open a fullscreen QR code page.


src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+    // Dynamically import qr-code-styling on the client-side, not compatible with regular Next imports
+    // https://github.com/kozakdenys/qr-code-styling/issues/38
+    if (typeof window !== 'undefined') {
+      if (qrCodeExists) return; // Prevent duplication
+      const QRCodeStyling = require('qr-code-styling');
+
+      const qrCode = new QRCodeStyling({
+        width: 636, // 4 * 159px, scaled down using CSS later
+        height: 636, // 4 * 159px, scaled down using CSS later
```
connorrothschild, line 53: I ran into some resolution issues with the initial height and width. Here, we render the canvas element with a width and height of 636px (4x the initial) and then we scale it down to its proper width and height (159px) using CSS. This makes the QR appear high res

src/scenes/UserGalleryPage/UserGalleryLayout.tsx
```diff
@@ -81,6 +81,7 @@ export const UserGalleryLayout = ({ userRef, queryRef }: Props) => {
       <UserGalleryHeader
         userRef={user}
         showMobileLayoutToggle={showMobileLayoutToggle}
+        showQRCode={isMobile}
```
connorrothschild, line 84: I am showing the QR code anytime the window width is mobile or smaller - let me know if we should change this

src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const [qrCodeExists, setQRCodeExists] = useState(false);
+
+  const handleClick = useCallback(() => {
+    setShowQRCode(true);
+
+    // Dynamically import qr-code-styling on the client-side, not compatible with regular Next imports
+    // https://github.com/kozakdenys/qr-code-styling/issues/38
+    if (typeof window !== 'undefined') {
+      if (qrCodeExists) return; // Prevent duplication
+      const QRCodeStyling = require('qr-code-styling');
```
connorrothschild, line 49: Note the unique way we have to import this library given Next constraints, if we need to adjust this let me know https://github.com/kozakdenys/qr-code-styling/issues/38

src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,207 @@
+import colors from 'components/core/colors';
+import styled from 'styled-components';
+import { useCallback, useState, useRef } from 'react';
+import Spacer from 'components/core/Spacer/Spacer';
+// import QRCodeStyling from 'qr-code-styling';
+
+function CloseIcon() {
+  return (
+    <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kaitoo1, line 9: it looks like we have a `CloseIcon` defined in `src/icons`, can we reuse that? 
I've been putting icons in that folder so it's easy to keep track of what icons we have

src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,207 @@
...
+      <path d="M12.6667 3.33334L3.33333 12.6667" stroke="black" stroke-miterlimit="10" />
+      <path d="M3.33333 3.33334L12.6667 12.6667" stroke="black" stroke-miterlimit="10" />
+    </svg>
+  );
+}
+
+function QRIcon() {
```
kaitoo1, line 16: another good candidate to move to `src/icons` - in case we want to reuse somewhere else in the future

src/scenes/UserGalleryPage/QRCode.tsx
kaitoo1: nicee thanks for the comment - I think it adds enough context that it would be useful to add this comment (or a summary) to the file

kaitoo1: can definitely see myself or someone new in the future wondering why its 4x

src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,207 @@
...
+      });
+
+      qrCode.append(ref.current);
+      setQRCodeExists(true);
+    }
+  }, [qrCodeExists, username]);
+
+  return (
+    <>
+      <StyledFullScreenQR showQRCode={showQRCode}>
```
kaitoo1, line 113: is it possible to use the modal component to display the QR code instead of this custom fullscreen component? 
it has some functionality like the close button we might be able to reuse.

also It looks like the full screen QR is always in the DOM even if it's not open, with opacity:0 - ideally we should make it so it's only in the DOM when it's open

src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,207 @@
...
+const StyledUsernameText = styled.p`
+  font-family: GT Alpina;
+  font-size: 24px;
+  font-weight: 300;
+  line-height: 28px;
+  letter-spacing: -0.04em;
+  text-align: left;
+`;
+
+const StyledHelperText = styled.p`
```
kaitoo1, line 187: we can use the BaseM component for this!

If you click into this component in Figma, you'll see it has a Text Style (BaseM):
CleanShot 2022-06-08 at 22 09 40


These are text styles defined in our Design System, and we have corresponding Text components defined in `Text.tsx` that you can use! You'll see a bunch in there like `TitleL`, `TitleM`, `BaseXL` etc. This should make it super easy to add text according to the spec. The goal is to rarely have to think about styling text per component.

Looks like you can use `TitleM` instead of `StyledUsernameText` too. `TitleM` is a bit special, and is italicized by default - if you wrap the text in <strong>s then it'll be non-italic

src/scenes/UserGalleryPage/UserGalleryCollection.tsx
```diff
@@ -129,37 +129,40 @@ function UserGalleryCollection({ queryRef, collectionRef, mobileLayout }: Props)
...
-                <CopyToClipboard textToCopy={collectionUrl}>
-                  <TextButton text="Share" onClick={handleShareClick} />
-                </CopyToClipboard>
-              </Dropdown>
+              <StyledCopyToClipboard textToCopy={collectionUrl}>
+                <StyledShareButton text="Share" onClick={handleShareClick} />
+              </StyledCopyToClipboard>
             )}
-          </StyledSettingsDropdown>
+            <StyledSettingsDropdown>
```
kaitoo1, line 138: nit: the "SHARE" text and '...' dropdown menu button is bit vertically off center
Screen Shot 2022-06-08 at 22 24 23


spec:
Screen Shot 2022-06-08 at 22 24 38

src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -7,11 +7,16 @@ import { pause } from 'utils/time';
...
+  successText?: string;
 };
 
 const SECTION_DURATION_MS = 400;
 
-export default function CopyToClipboard({ textToCopy, children }: Props) {
+export default function CopyToClipboard({
+  textToCopy,
+  children,
+  successText = 'Link Copied',
```
kaitoo1, line 18:
```suggestion
  successText = 'Copied link to clipboard,
```

kaitoo1: looks like the copy was updated a bit
Screen Shot 2022-06-08 at 22 47 20

src/scenes/UserGalleryPage/QRCode.tsx
kaitoo1: I think it's totally fine to lazy load this here the way you did. Lazy loading doesn't seem frowned upon.

next.js's dynamic import that's suggested in the GitHub issue you linked also seems to work (TIL), but I think it's pretty much the same idea of waiting to load the required code client side. The difference is between loading the `qr-code-styling` library clientside vs loading the `QRCode` component client side, and since the problem is specific to the `qr-code-styling` library, makes sense to dynamic load just that bit.


https://nextjs.org/docs/advanced-features/dynamic-import
Screen Shot 2022-06-08 at 23 08 41

src/scenes/UserGalleryPage/QRCode.tsx
connorrothschild: Hmm, the modal component is styled pretty differently from this fullscreen view. It has a border, is centered in the screen, etc.

We could use the same logic to render a new modal (unstyled, or a new component `FullScreenModal`)

connorrothschild: @kaitoo1  Handled in https://github.com/gallery-so/gallery/pull/807/commits/03deacd9e9c146361a77f62a2753540386825562

src/scenes/UserGalleryPage/UserGalleryCollection.tsx
connorrothschild: Fixed using `background-position: center`

connorrothschild:
[This commit](https://github.com/gallery-so/gallery/pull/807/commits/f689a2e132235bc2e467335628f0d75340d70170) simplifies how we present the "Share" button and the "..." dropdown.

Previously we were using a `isHovering`/`setIsHovering` variable to render the icons, but I believe we can handle this with pure CSS? Unless there is a use case I am missing; worth reviewing @kaitoo1 

I believe this gets us close to the finish line of Jess's designs!

image

image

image

src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,149 @@
...
+            color1: '#ffffff',
+            color2: '#ffffff',
+            rotation: '0',
+          },
+        },
+      });
+
+      qrCode.append(ref.current);
+    }
+  }, [ref, username]);
```
Robinnnnn, line 76: don't think we need `ref` in the dependency array here! unless eslint is asking for it?

src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const { showModal } = useModalActions();
+  const handleClick = useCallback(() => {
+    // Need a brief timeout so that the modal renders before rendering QR code. Otherwise the ref will not exist and renderQRCode cannot append the ref
+    setTimeout(() => {
+      renderQRCode();
+    }, 100);
```
Robinnnnn, line 83: so we have a const for the modal transition time that's `ANIMATED_COMPONENT_TRANSITION_MS` which is 300ms. could we use that here? or would it be too slow?

src/scenes/UserGalleryPage/UserGalleryCollection.tsx
```diff
@@ -123,43 +114,45 @@ function UserGalleryCollection({ queryRef, collectionRef, mobileLayout }: Props)
...
+                      <Spacer height={8} />
+                    </>
+                  )}
+                  <TextButton
+                    text="View Collection"
+                    onClick={handleViewCollectionClick}
+                    underlineOnHover
+                  />
+                </Dropdown>
+              </>
```
Robinnnnn, line 153: nit: can get rid of fragment <></>

src/scenes/UserGalleryPage/QRCode.tsx
Robinnnnn: +1 thank you for the comments!

Robinnnnn:
@connorrothschild i pulled this down and tested locally and it's fiiiiire 🔥 🔥 🔥 

like @kaitoo1 mentioned, the modal actually comes with a new prop called `isFullPageOverride` that you can use for full-page modals.

image

**that said,** let's keep your implementation for now, because i'm in the middle of a refactor of how our modals work, specifically on mobile. so even if you were to use the preferred method, it would just cause a merge conflict.

imo let's merge this baby and set the EMPLOYEE_FLAGS to true. lmk if you need help using it, but there are examples when you search for `isFeatureEnabled`. might need some light graphql work which i'm happy to pair with you on (just ping me on discord)

image

image

after that jess can test on a real device in prod!

src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -45,22 +50,33 @@ const Container = styled.span`
...
+  background: white;
+  border: 1px solid ${colors.offBlack};
+
+  border-radius: 0px;
+  padding: 8px 16px 8px 16px;
+
   width: fit-content;
   pointer-events: none;
 
+  & p {
```
kaitoo1, line 76: there's a toast component in Toast.tsx that we might be able to use here!
you can check for usages of `useToastActions` for examples

src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -45,22 +50,33 @@ const Container = styled.span`
...
-    to { opacity: 0; transform: translateY(-14px); };
+    from { opacity: 1; transform: translateX(-50%) translateY(-28px); };
+    to { opacity: 0; transform: translateX(-50%) translateY(-14px); };
 `;
 
 const Toast = styled.div<{ isActive: boolean }>`
-  position: absolute;
-  background: black;
-  padding: 4px 8px;
+  position: fixed;
```
kaitoo1, line 63: there's a toast component in Toast.tsx that we might be able to use here!
you can check for usages of useToastActions for examples

src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -45,22 +50,33 @@ const Container = styled.span`
...
+  border: 1px solid ${colors.offBlack};
+
+  border-radius: 0px;
+  padding: 8px 16px 8px 16px;
+
   width: fit-content;
   pointer-events: none;
 
+  & p {
+    color: ${colors.offBlack};
```
kaitoo1, line 77: is this overriding the `colors.white` set on L40?

src/scenes/UserGalleryPage/QRCode.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+const StyledBaseM = styled(BaseM)`
+  text-align: center;
+  position: absolute;
+  bottom: 32px;
+`;
+
+// In order to render the QR code at a higher quality, we render it at 10x its initial size, and then scale it down with the following CSS
+const StyledQRWrapper = styled.div`
+ height: 159px;
```
kaitoo1, line 142: super-nit: indent

src/components/CopyToClipboard/CopyToClipboard.tsx
connorrothschild: So much cleaner! but getting a weird duplication/jitter error.

https://user-images.githubusercontent.com/13339581/172883151-c77e7aa9-2a08-4be1-ad0c-9975ad9d9936.mov

I verified the [`useCallback` responsible for pushing the toast](https://github.com/gallery-so/gallery/pull/807/commits/010403c55fc9439521b37c5e175d2ec96b67a487#diff-9f844a7dbe8597affea197b397fc099857b0b7f6c41e51922234ff3020e094c5R18-R23) only runs once. So I wonder if this is an error with rerendering the toast based on ID. (Currently the key is just the toast message, which would be the same across collections)

I get this error in console:

> Warning: Encountered two children with the same key, `Copied link to clipboard`. Keys should be unique so that components maintain their identity across updates.

src/scenes/UserGalleryPage/QRCode.tsx
connorrothschild: Technically I think the duration doesn't matter. (It also works with a `setTimeout` of 0) - the important part is that the modal begins to mount (no matter its animation status, it mounts instantly I think. So long as we wait a millisecond so that the modal is rendered before rendering the QR code, we're good

src/components/CopyToClipboard/CopyToClipboard.tsx
connorrothschild: Fixed!

PR #808, opened 2022-06-04
[GAL-208] Tweak wizard footer copy in collection editor

connorrothschild:
1. "Back" now says "Cancel" on the collection editor
2. During onboarding, this still says "Back"
3. Clicking "Cancel" displays the "Are you sure?" modal


src/flows/shared/components/WizardFooter/WizardFooter.tsx
```diff
@@ -21,6 +23,7 @@ function WizardFooter({
   shouldHideFooter,
   shouldHideSecondaryButton,
   footerButtonTextMap,
+  isOnboarding,
```
connorrothschild, line 26: `WizardFooter` now takes a prop called `isOnboarding` which will determine whether to show "Back" or "Cancel" in the back button. I don't think we could access this with the prior configuration (unless I missed something)

src/scenes/Modals/ConfirmLeaveModal.tsx
```diff
@@ -9,8 +9,35 @@ import useBackButton from 'hooks/useBackButton';
...
+  const { hideModal } = useModalActions();
+  const { back } = useRouter();
+
+  const goBack = useCallback(() => {
+    back();
+    hideModal();
+  }, [back, hideModal]);
+
+  return <Modal onClick={goBack} />;
+}
```
connorrothschild, line 37: I've split up `ConfirmLeaveModal.tsx` to include two components:
* One `ConfirmBackModal` which just runs the generic `back()` from `useRouter` when the user confirms (this mimics the behavior of the back button in the collection editor currently)
* `ConfirmLeaveModal` which contains the back functionality used elsewhere.

`ConfirmBackModal` does not require a username. It simply runs `back()`, which is exactly what clicking the back button did in the old configuration ([see here](https://github.com/gallery-so/gallery/pull/808/files#diff-993ecc44e51efddeff84df7b02d4730e1db35f41bf45645479b0273536572d28L70-L72)).

Tl;dr, this just wraps the `back()` functionality in a modal

src/flows/shared/components/WizardFooter/WizardFooter.tsx
Robinnnnn: oh dang i should've mentioned that you can use `useWizardId` to check if the `id === 'onboarding'`! so you won't need to use this prop.

example usage: https://github.com/gallery-so/gallery/blob/main/src/flows/shared/steps/OrganizeGallery/OrganizeGallery.tsx#L69

connorrothschild: Great, thanks!

connorrothschild:
In onboarding:

image

In collection editor:

image


Leaving collection editor via `cancel` shows a modal:

image

Robinnnnn:
thanks @connorrothschild!!

i think we want to display the "are you sure" modal if the user gets into the editor through the "normal" way (via the navbar)

https://user-images.githubusercontent.com/12162433/172255921-15891ef6-df52-4acf-99f8-82681c5a034c.mov

src/flows/shared/components/WizardFooter/WizardFooter.tsx
```diff
@@ -65,18 +70,60 @@ function WizardFooter({
     next();
   }, [back, collectionId, next, onNext]);
 
+  const { showModal } = useModalActions();
+
   const handlePreviousClick = useCallback(() => {
```
connorrothschild, line 75: On click, we show a `GenericActionModal` which runs the specific action and then hides the modal afterwards. These behaviors are in line with what we talked about in Linear

src/scenes/Modals/GenericActionModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+import Spacer from 'components/core/Spacer/Spacer';
+import styled from 'styled-components';
+import { useModalActions } from 'contexts/modal/ModalContext';
+import { useCallback } from 'react';
+
+export default function GenericActionModal({
+  action,
+  bodyText,
+  buttonText,
+}: {
```
connorrothschild, line 13: `GenericActionModal` takes in an action, the body text, and the button text. This makes it reusable across all components.

After executing `action`, it runs `hideModal()`

connorrothschild: This also removes the need for `ConfirmLeaveModal`, which has since been deleted. This modal essentially renavigated to the `/{username}` but, per Linear, we now want to navigate dynamically based on the query parameters.

Now, we handle all exit navigation from the `WizardFooter` and pass in the intended action to `GenericActionModal` (ensuring that the `Back` button and escape key mimic the same behavior.

src/flows/shared/steps/OrganizeCollection/Editor/CollectionEditor.tsx
```diff
@@ -202,13 +198,6 @@ function CollectionEditor({ viewerRef }: Props) {
 
   const shouldDisplayEditor = stagedNfts.length > 0;
 
-  const { showModal } = useModalActions();
-  const user = viewer.user;
-
-  useKeyDown('Escape', () => {
-    showModal({ content: <ConfirmLeaveModal userRef={user} /> });
-  });
```
connorrothschild, line 210: We no longer need to handle any escape keypresses in the `CollectionEditor` component

src/flows/shared/components/WizardFooter/WizardFooter.tsx
```diff
@@ -86,7 +133,9 @@ function WizardFooter({
     <StyledWizardFooter>
       {shouldHideSecondaryButton ? null : (
         <ActionText color={colors.metal} onClick={handlePreviousClick}>
-          {isFirstStep ? 'Cancel' : 'Back'}
+          {isFirstStep || (step.id === 'organizeCollection' && wizardId !== 'onboarding')
+            ? 'Cancel'
+            : 'Back'}
```
connorrothschild, line 138: This handles the dynamic text

connorrothschild:
@Robinnnnn I think this is ready for another review.

This is a bit more involved now because it refactors the way that we did present the `ConfirmLeaveModal` and instead replaces it with `GenericActionModal`. All is working in my testing!

src/flows/shared/components/WizardFooter/WizardFooter.tsx
```diff
@@ -65,18 +70,60 @@ function WizardFooter({
...
+    // If the user is editing from the collection page directly (has a collection ID), use back()
+    if (wizardId !== 'onboarding' && collectionId) {
+      showModal({
+        content: (
+          <GenericActionModal
+            bodyText="Would you like to stop editing?"
+            buttonText="Leave"
+            action={() => {
+              back();
+            }}
```
Robinnnnn, line 105: nit: `action={back}`

Robinnnnn:
@connorrothschild thanks so much for keeping at this! i know it was a lot to ask with a lot of edge cases!

i made a small commit here: https://github.com/gallery-so/gallery/pull/808/commits/25a192b33566cd0d2a57fc6c9e2b0f26d653e2e7

mostly just adding some clarity and re-ordering the statements in `handlePreviousClick`: https://github.com/gallery-so/gallery/blob/25a192b33566cd0d2a57fc6c9e2b0f26d653e2e7/src/flows/shared/components/WizardFooter/WizardFooter.tsx#L75-L121

PR #809, opened 2022-06-06
Added FF to opensea base URL for mint page

jakzaizzat:
Added FF to opensea base url too. 

Currently, we used testnet opensea API to test. Added conditional to replace testnet api to production opensea api

So, during the launch we can just turn on the FF and everything already in place.


kaitoo1 (approved):
nice! good idea, super useful

PR #815, opened 2022-06-07
[GAL-212] Remove link formatting when clicking on empty link

connorrothschild:
If the user clicks the Link button while their cursor is in a newly created link (`[](https://)`) this will remove that link.

This does not handle the case where a user selects text (`Hello`), creates a link (`[Hello](https://)`) and then clicks again. This is a more difficult problem because the middle text (`Hello`) could be any number of characters; it is difficult to handle in our current configuration.

### DIFF
--- a/src/components/core/Markdown/Link.tsx
+++ b/src/components/core/Markdown/Link.tsx
@@ -16,11 +16,17 @@ export default function Bold({
     const [start, end] = selectedRange;
     const selectedText = textArea.value.substring(start, end);
 
-    // If user is inbetween two brackets, or at the end of (https://), they probably just clicked the link button. Do nothing
-    if (
-      textArea.value.substring(start - 1, end + 1) == '[]' ||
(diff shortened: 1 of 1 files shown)


kaitoo1:
Tested locally and looks good! is it possible to add a test case for this in MarkdownShortcuts.test.tsx?

connorrothschild:
@kaitoo1 I tried to add a test but our tests for `MarkdownShortcuts` are limited by the fact that I can't seem to update cursor selection range :(

And this change triggers based on the surrounding characters (so if cursor is at position 1, verifying that position 0 and position 2 are `[` and `]`

So we can't quite test it until we find a fix for the "update `selectionStart`/`End`" problem in our tests :(

[More here](https://github.com/gallery-so/gallery/blob/main/src/components/core/Markdown/MarkdownShortcuts.test.tsx#L43)

Robinnnnn (approved):
let's goooo thank you connor!

PR #819, opened 2022-06-08
dont log exception if because user is not early access

kaitoo1:
### Context
Sentry alerts keep getting fired because an error gets logged when a user tries to sign in and they don't have a membership nft.
This isn't an app error, so we don't need to log it in Sentry. we still track the event in analytics, so we don't lose this information.

### Changes
Added a util function `isNotEarlyAccessError` to check the error type using the error message, and updated the add wallet + sign in flows to not log the exception if this is true.


Robinnnnn (approved):
LGTM

PR #823, opened 2022-06-09
[GAL-188] improve modal height on mobile

kaitoo1:
this PR fixes the modal height extending beyond the visible screen area, causing the close button and other important elements to be moved off screen.

Tested on iOS simulator + brave 

Before
Screen Shot 2022-06-09 at 20 39 30
After 
Screen Shot 2022-06-09 at 20 37 39

Before
Screen Shot 2022-06-09 at 20 39 19
After
Screen Shot 2022-06-09 at 20 37 47

Regression check for normal modals + NFT Detail page on desktop:
Screen Shot 2022-06-09 at 20 46 06
Screen Shot 2022-06-09 at 20 43 12


src/contexts/modal/AnimatedModal.tsx
```diff
@@ -140,7 +143,7 @@ const StyledContent = styled.div<{ isFullPage: boolean; isPaddingDisabled: boole
     isFullPage
       ? `
     width: 100vw;
-    height: 100vh;
+    min-height: -webkit-fill-available;
```
kaitoo1, line 146: would like to test on android / mobile chrome

Robinnnnn (approved):
🔥🔥🔥🔥🔥🔥

holy cow this is so beautifully elegant

PR #824, opened 2022-06-09
Fix queryref bug

Robinnnnn:
Fix full-screen errors for `DecoratedCollectionEditor could not access queryRef`

### DIFF
--- a/src/contexts/wizard/WizardDataProvider.tsx
+++ b/src/contexts/wizard/WizardDataProvider.tsx
@@ -62,8 +62,8 @@ export default memo(function WizardDataProvider({ id, children }: Props) {
     setIsRefreshingNfts(true);
 
     try {
-      await refreshOpenseaNfts();
       loadQuery({}, { fetchPolicy: 'store-and-network' });
+      await refreshOpenseaNfts();
     } catch (error: unknown) {
       captureException(error);
       if (error instanceof Error) {


src/contexts/wizard/WizardDataProvider.tsx
```diff
@@ -62,8 +62,8 @@ export default memo(function WizardDataProvider({ id, children }: Props) {
     setIsRefreshingNfts(true);
 
     try {
-      await refreshOpenseaNfts();
       loadQuery({}, { fetchPolicy: 'store-and-network' });
+      await refreshOpenseaNfts();
```
Robinnnnn, line 66: **this is a bug that only occurred during onboarding**

on occasion, the `refreshOpenseaNfts` query would fail during onboarding – if the user was signing up during a high-congestion period, for example. since the `await` statement is wrapped in try/catch, on failure it would immediately throw to the catch block and skip the `loadQuery` line.

_without having the query loaded, the pre-load attempt for the collection editor would fail._

Robinnnnn: this would also fail if the user had a ton of NFTs, and the `await` statement took forever / did not respond in time for the collection creation step

kaitoo1: 🧠🧠🧠

PR #829, opened 2022-06-11
[GAL-238] Fix collection header & layout

connorrothschild:
This PR makes collection headers and navbar more in line with the Figma designs specified by Jess.

### DIFF
--- a/src/contexts/globalLayout/GlobalNavbar/LoggedInNav.tsx
+++ b/src/contexts/globalLayout/GlobalNavbar/LoggedInNav.tsx
@@ -49,8 +49,6 @@ function LoggedInNav({ queryRef }: Props) {
   );
 
   const { query: routerQuery } = useRouter();
-  const userOwnsCollectionOrGallery = routerQuery?.username === query?.viewer?.user?.username;
-  console.log(userOwnsCollectionOrGallery);
 
   const handleManageWalletsClick = useCallback(() => {
     showModal({ content: <ManageWalletsModal queryRef={query} /> });
@@ -82,6 +80,7 @@ function LoggedInNav({ queryRef }: Props) {
   }
 
   const username = query.viewer.user?.username;
+  const userOwnsCollectionOrGallery = routerQuery?.username === username;
 
   const hasNotifiction = isFeatureEnabled(FeatureFlag.POSTER_MINT, query) && !isMintPosterDismissed;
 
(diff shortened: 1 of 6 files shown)


src/contexts/globalLayout/GlobalNavbar/LoggedInNav.tsx
```diff
@@ -48,6 +48,8 @@ function LoggedInNav({ queryRef }: Props) {
     queryRef
   );
 
+  // FIXME: GraphQL work goes here to get collection ID, if any (so that we can render separate options in navbar)
```
connorrothschild, line 51: Here is where I imagined the query would go (or combined with the above query)

Robinnnnn: should be able to get collection ID like this:

```js
const { query } = useRouter()
const { collectionId } = query

// if collectionId is present, it means that we're on a collection page
```

src/scenes/CollectionGalleryPage/CollectionGalleryHeader.tsx
```diff
@@ -223,12 +237,8 @@ const StyledCollectionName = styled.div`
...
 const StyledSeparator = styled.div`
   margin: 0 10px;
-  display: none;
+  display: block;
   color: ${colors.metal};
-
-  @media only screen and ${breakpoints.mobileLarge} {
-    display: block;
-  }
 `;
```
connorrothschild, line 242: The new Figma designs have a `/` in the header text on all screen sizes (e.g. `connorr / My Collection`), so we remove the dynamic `display: ` here

src/scenes/UserGalleryPage/LinkButton.tsx
```diff
@@ -3,12 +3,9 @@ import styled from 'styled-components';
 
 import CopyToClipboard from 'components/CopyToClipboard/CopyToClipboard';
 
-export default function LinkButton({ username }: { username: string }) {
+export default function LinkButton({ textToCopy }: { textToCopy: string }) {
   return (
-    <CopyToClipboard
-      textToCopy={`https://gallery.so/${username}`}
-      successText={'Copied link to clipboard'}
-    >
```
connorrothschild, line 11: Now, `LinkButton` takes in the `textToCopy` as a string so that it can be used for different URLs (galleries, collections, etc.)

Robinnnnn (approved):
LGTM!!

PR #852, opened 2022-06-21
QR code refactor

kaitoo1:
### Description
This PR refactors QRCode.tsx to separate out the responsibilities of showing the popover (modal), generating the QR code, and appending it to the dom into different components. 

Doing this helped fix the bug where the QR code is initially rendered without the logo image. To roughly summarize, it does this by generating the QR code at the `UserGalleryHeader` level, and making the QR popover (modal) responsible for appending it to the dom. The underlying problem seemed to be that the QR code was being drawn onto the canvas before the image was properly loaded, causing the image to be missing on the first render, but available on subsequent renders. Separating the QR generation and appending into 2 components that are used at different stages of the user flow (gallery view vs modal view) gives enough time for the image to load. I also found some idiosyncrasies with how usage of `QRCodeStyling` affected the canvas behavior that required configuring it in a very specific way, that I've left comments to explain. 

### Changes
- created a `useQrCode` hook and moved the `new QRCodeStyling()` logic there. 
- created a `QRCodeButton` component responsible for displaying a clickable QR icon and opening a modal on click
- created a `QRCodePopover` component that contains the displayed QR code and username inside the popover. It is responsible for appending the QRCodeStyling canvas to the dom.
- updated the `UserGalleryHeader` component to call the new `useQrCode` hook to generate the QR code at the user gallery level without appending it immediately.
- misc clean up

Before

https://user-images.githubusercontent.com/80802871/174833269-a0c4bf4a-f113-4851-8bce-1521748d0905.mp4

After

https://user-images.githubusercontent.com/80802871/174833292-d784bcc2-2237-408e-9a4c-9c1a3e470966.mp4


src/icons/LinkIcon.tsx
```diff
@@ -0,0 +1,15 @@
+export default function LinkIcon() {
```
kaitoo1, line 1: unrelated, just moving Link icon into its own file

Robinnnnn (approved):

PR #858, opened 2022-06-28
[GAL-234] Add switch network method during sign in

jakzaizzat:
If the user is not in the `mainnet` we show `switch network` button 

**Before**
CleanShot 2022-06-28 at 11 11 46@2x

**Demo**

https://user-images.githubusercontent.com/4480258/176084115-2b707f51-645f-47d6-9a74-9a9a738a6b31.mp4

### DIFF
--- a/src/components/WalletSelector/WalletSelector.tsx
+++ b/src/components/WalletSelector/WalletSelector.tsx
@@ -158,6 +158,18 @@ function WalletSelector({ connectionMode = AUTH, queryRef }: Props) {
     deactivate();
   }, [deactivate]);
 
+  const handleSwitchNetwork = async () => {
+    const provider = await injected.getProvider();
+    provider.request({
+      method: 'wallet_switchEthereumChain',
+      params: [
+        {
+          chainId: '0x1', // Mainnet
+        },
+      ],
+    });
+  };
+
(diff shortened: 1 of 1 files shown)


src/components/WalletSelector/WalletSelector.tsx
```diff
@@ -158,6 +158,18 @@ function WalletSelector({ connectionMode = AUTH, queryRef }: Props) {
...
+    const provider = await injected.getProvider();
+    provider.request({
+      method: 'wallet_switchEthereumChain',
+      params: [
+        {
+          chainId: '0x1', // Mainnet
+        },
+      ],
+    });
+  };
```
Robinnnnn, line 171: can we wrap this in `useCallback`

Robinnnnn:
omfg this is so elegant!!! thanks so much for the fix! 😱😱😱

src/components/WalletSelector/WalletSelector.tsx
jakzaizzat: Sure! Just push the changes

kaitoo1:
brilliant! that looks great, so smooth

kaitoo1 (approved):

PR #861, opened 2022-06-29
[GAL-253] FEEEEEEEEED

kaitoo1:
### Overview
This PR adds the Feed.

High level, it adds a new /home page that can render a Global feed or Viewer (following) feed.
The Feed is populated by a list of different Feed Event type components for different actions such as adding pieces to a collection, following someone, adding a collectors note, etc. 

The list is paginated and infinite scrolls, and is virtualized for performance.

### Change details

Feed page
- created `LandingPage` to house the current landing page (was previously Home)
- created Feed.tsx to house the Feed page. It also adds the `Following | Worldwide` toggle (`FeedNavbarControl`) to the navbar.
  - This will render the ViewerFeed or GlobalFeed depending on the toggled state. ViewerFeed is only available for logged in users.

Feed List components
- `FeedList.tsx` contains the juicy logic to build the list of feed event rows, to handle pagination, infinite scroll, and virtualization.
- It uses `FeedEvent.tsx` to actually render each row. This component checks the event type to decide which component to use to render it. 
  - Currently we support the following types
    - `TokensAddedToCollectionFeedEventData`
    - `CollectionCreatedFeedEventData`
    - `CollectorsNoteAddedToTokenFeedEventData`
    - `UserFollowedUsersFeedEventData`
  - there are additional event types that the backend returns, but these are not spec'd to be displayed yet: 
    - `UserCreatedFeedEventData`
    - `CollectorsNoteAddedToCollectionFeedEventData`
  - These row components can be used for both the Global and Viewer feeds.

components inside the above row components
- `FeedEventTokenPreviews` takes a list of tokens associated with an event (such as adding pieces to a collection) and renders them with the correct size and layout depending on the # of tokens
- `UserFollowedUsersFeedEvent` is a bit unique, I left a comment in code but basically we want to display when someone follows you (logged in user) separate from if they follow other users. However all follow actions in a certain window of time are bundled into a single event. We decided it would be best for now for the front end to handle teasing out the "Followed you" events. It also displays differently if the user followed you back, vs a fresh follow.

Refactors
- refactored the NftPreview a bit so that we could reuse it for the feed (currently used on user gallery page)
- refactored the Followers list to also reuse the list rows

todos

- [x]  bug: click into collection -> click username . Should take to gallery not home
- [ ] error boundary for each list item so an error on an event doesn't blow up feed
- [x] properly put things behind feature flag


src/components/Feed/Events/CollectionCreatedFeedEvent.tsx
```diff
@@ -0,0 +1,104 @@
...
+          }
+        }
+      }
+    `,
+    eventRef
+  );
+  const { push } = useRouter();
+
+  const tokensToPreview = useMemo(() => {
+    return event.collection.tokens.filter((n) => !!n).slice(0, 4);
```
Robinnnnn, line 46: nit: since this is a common thing when using graphql, there's an existing helper function called `removeNullValues`. can use `removeNullValues(event.collection.tokens)`

src/components/Feed/Events/FeedEventNftPreviewWrapper.tsx
```diff
@@ -0,0 +1,85 @@
...
+    tokenRef
+  );
+
+  const { showModal } = useModalActions();
+
+  const handleClick = useCallback(() => {
+    showModal({
+      content: (
+        <StyledNftDetailViewPopover>
+          <NftDetailView username={'kaito'} authenticatedUserOwnsAsset={false} queryRef={token} />
```
Robinnnnn, line 43: remove hardcode

src/components/Feed/Events/TokensAddedToCollectionFeedEvent.tsx
```diff
@@ -0,0 +1,103 @@
...
+          ...EventMediaFragment
+        }
+      }
+    `,
+    eventRef
+  );
+  const { push } = useRouter();
+
+  const tokensToPreview = useMemo(() => {
+    return event.newTokens.filter((n) => !!n).slice(0, 4);
```
Robinnnnn, line 47: nit: removeNullValues

src/components/Feed/Feed.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const handleFollowingModeClick = useCallback(() => {
+    setFeedMode(FOLLOWING);
+    setFeedModeCopy(FOLLOWING);
+  }, [setFeedMode]);
+
+  const handleWorldwideModeClick = useCallback(() => {
+    setFeedMode(WORLDWIDE);
+    setFeedModeCopy(WORLDWIDE);
+  }, [setFeedMode]);
```
Robinnnnn, line 37: hmm a bit confused by the state duplication.

could do something like this
```js
function feedNavbarControl({ feedMode, setFeedMode }) { // get state/setter passed in from parent
  ...
  <StyledNavControlText
    onClick={() => setFeedMode(FOLLOWING)}
    focused={feedMode === FOLLOWING}
  />
  ...
}
```

then we wouldn't need local state in this file. also should be fine to not use `useCallback` since the Text element is super basic

src/components/Feed/Feed.tsx
```diff
@@ -0,0 +1,150 @@
...
+export default function Feed() {
+  const query = useLazyLoadQuery<FeedViewerQuery>(
+    graphql`
+      query FeedViewerQuery {
+        viewer {
+          ... on Viewer {
+            user {
+              dbid
+            }
+            # ...FeedNavbarViewerFragment
```
Robinnnnn, line 96: meant to remove?

src/components/Feed/Feed.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const [feedMode, setFeedMode] = useState<FeedMode>(defaultFeedMode);
+  const { setCustomNavCenterContent } = useGlobalLayoutActions();
+
+  // This effect ensures the Feed controls on the navbar are removed when the Feed unmounts, so that it is not visible when navigating to a different page.
+  useEffect(() => {
+    return () => {
+      setTimeout(() => {
+        setCustomNavCenterContent(null);
+      }, 200);
```
Robinnnnn, line 115: we should be able to use `FADE_TRANSITION_TIME_MS` which is 300ms, used by the navbar. unless the interaction is weird

src/components/Feed/FeedEventTokenPreviews.tsx
```diff
@@ -0,0 +1,56 @@
...
+export type TokenToPreview = EventMediaFragment$key & {
+  token: { dbid: string };
+};
+
+type Props = {
+  tokensToPreview: TokenToPreview[];
+};
+
+const DEFAULT_DIMENSIONS_DESKTOP = 259.33;
+const SMALL_DIMENSIONS_DESKTOP = 190.5;
```
Robinnnnn, line 16: decimals 🤯

src/components/Feed/FeedList.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const rowCount = hasNext ? feedData.edges.length + 1 : feedData.edges.length;
+  // Only load 1 page of items at a time.
+  // Pass an empty callback to InfiniteLoader in case it asks us to load more than once.
+  const loadMoreRows = isNextPageLoading ? () => {} : onLoadNext;
+
+  return (
+    <InfiniteLoader
+      isRowLoaded={isRowLoaded}
+      // loadMoreRows type expects a function that returns a promise, but react-virtualized docs suggest passing an empty callback in some scenarios
+      //@ts-ignore
```
Robinnnnn, line 94: nit: new version of ignoring TS errors is `@ts-expect-error: <reason for ignoring>`

`@ts-ignore` was deprecated

src/components/NftPreview/NftPreview.tsx
```diff
@@ -162,11 +115,13 @@ function NftPreview({ galleryNftRef }: Props) {
   );
 }
 
+// const StyledLink = styled(Link)``;
```
Robinnnnn, line 118: toss

src/components/NftPreview/NftPreview.tsx
```diff
@@ -162,11 +115,13 @@ function NftPreview({ galleryNftRef }: Props) {
...
+
 const StyledA = styled.a`
   display: flex;
   justify-content: center;
   align-items: center;
-  width: 100%;
+  width: fit-content;
```
Robinnnnn, line 124: oh man, let's make sure this doesn't break SVGs haha

src/components/NftPreview/NftPreview.tsx
```diff
@@ -200,13 +155,12 @@ const StyledNftPreview = styled.div<{
...
   align-items: center;
   position: relative;
-  height: fit-content;
   overflow: hidden;
 
   ${({ backgroundColorOverride }) =>
     backgroundColorOverride && `background-color: ${backgroundColorOverride}`}};
 
-  max-height: 80vh;
+
```
Robinnnnn, line 163: i think this used to make sure that portrait NFTs look, but can take a look

src/scenes/Auth/Auth.tsx
```diff
@@ -50,7 +50,7 @@ function Auth({ queryRef }: Props) {
   if (viewer?.__typename === 'Viewer') {
     // If user exists in DB, send them to their profile
     if (viewer.user?.username) {
-      return <GalleryRedirect to={`/${viewer.user.username}`} />;
+      return <GalleryRedirect to="/home" />;
```
Robinnnnn, line 53: 😍

src/utils/string.ts
```diff
@@ -0,0 +1,3 @@
+export function pluralize(count: number, singular: string) {
+  return count === 1 ? singular : `${singular}s`;
+}
```
Robinnnnn, line 3: lovely

src/components/NftPreview/NftPreview.tsx
kaitoo1: ah yes good call had on my todo to check and then forgot

src/components/Feed/Events/CollectorsNoteAddedToTokenFeedEvent.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const size = isMobile ? (windowSize.width - 2 * MARGIN - MIDDLE_GAP) / 2 : IMAGE_SPACE_SIZE;
+
+  const handleEventClick = useCallback(
+    (e) => {
+      e.preventDefault();
+      showModal({
+        content: (
+          <StyledNftDetailViewPopover>
+            <NftDetailView
+              username={'kaito'}
```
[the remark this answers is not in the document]
kaitoo1: done, removed

src/components/Feed/Events/FeedEventNftPreviewWrapper.tsx
kaitoo1: done, removed

src/components/Feed/GlobalFeed.tsx
```diff
@@ -0,0 +1,76 @@
...
+      }
+    `,
+    {
+      last: ITEMS_PER_PAGE,
+    }
+  );
+
+  const { data, hasPrevious, loadPrevious, isLoadingPrevious } = usePaginationFragment<
+    GlobalFeedQuery,
+    any
```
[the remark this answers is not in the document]
kaitoo1: oh so the relay docs said this one doesn't have to be specified, so it suggested using `_` so that the type can be inferred, but type check didn't like it so I just threw `any` in here
https://relay.dev/docs/api-reference/use-pagination-fragment/

do you know if we can use `_`?

src/utils/string.ts
kaitoo1: will need to upgrade in the future for weird plurals but works for now lol

src/components/Feed/Feed.tsx
```diff
@@ -0,0 +1,150 @@
...
+            }
+            # ...FeedNavbarViewerFragment
+          }
+        }
+      }
+    `,
+    {}
+  );
+  const { viewer } = query;
+  const viewerUserId = (viewer?.user?.dbid ?? '') as string;
```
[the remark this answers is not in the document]
kaitoo1: removed cast!
vscode was complaining so I added it but it must have just been slow

src/components/NftPreview/NftPreview.tsx
kaitoo1: ah yup definitely. def don't want to remove, good catch
now that I can point to dev or prod, I'll do more testing with different tokens

src/components/Feed/GlobalFeed.tsx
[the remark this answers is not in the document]
Robinnnnn: ah gotcha, any should be chill then!

pages/home.tsx
```diff
@@ -4,3 +4,39 @@ import GalleryRoute from 'scenes/_Router/GalleryRoute';
...
+ *
+ * If a page *does not* export `getServerSideProps`, returning to the page triggers a
+ * NextJS route change BEFORE the browser believes it has transitioned via `window.popState`.
+ *
+ * However, exporting `getServerSideProps` for some reason ensures the browser's `popState`
+ * fires BEFORE the NextJS route change, which is necessary behavior for our FadeTransitioner.
+ * This is because our FadeTransitioner depends on listening to the `popState` signal to
+ * schedule a graceful scroll restoration before the transition begins.
+ *
+ * To summarize:
```
kaitoo1, line 19: amazing documentation

src/components/Feed/Feed.tsx
kaitoo1: unfortunately because this component gets added to the navbar via `setCustomNavCenterContent()` it doesn't seem to get the updated `feedMode` value if it changes

kaitoo1: if you know of a workaround that would be awesome

kaitoo1: might have to just re-add this component to the navbar whenever the user switches the mode..

src/components/Feed/Feed.tsx
kaitoo1: done

src/components/Feed/Events/CollectionCreatedFeedEvent.tsx
kaitoo1: brilliant

kaitoo1: swapped to removeNullValues

kaitoo1:
> noticed CollectorsNote being added / edited wasn't showing up in the feed, so i think we just need to support this new type
> 
> **EDIT: moving this convo to threads**
> 
> image
Screen Shot 2022-06-30 at 21 23 32

yeet good catch, these are the types  that we don't display rn (but the backend returns) - will be ready to when they get spec'd!

PR #864, opened 2022-07-01
Feed Polish

Robinnnnn:


src/components/Feed/Feed.tsx
```diff
@@ -20,41 +20,30 @@ export type FeedMode = typeof FOLLOWING | typeof WORLDWIDE;
 
 type ControlProps = {
   setFeedMode: (mode: FeedMode) => void;
-  initialFeedMode: FeedMode;
+  feedMode: FeedMode;
 };
 
-function FeedNavbarControl({ setFeedMode, initialFeedMode }: ControlProps) {
-  // Internally mirror the feed mode state so that we can avoid adding it to the dep array, and avoid re-adding these controls to the navbar when the user switches from Following to Worldwide feed
-  const [feedModeCopy, setFeedModeCopy] = useState<FeedMode>(initialFeedMode);
```
Robinnnnn, line 28: fyi @kaitoo1 continuing on this convo: https://github.com/gallery-so/gallery/pull/861#discussion_r910446295

decided we don't need to mirror and deal with re-mounting, esp since it allowed the right element to be highlighted when you click the button from the zero-inbox state.

not worried about re-mounting since under the hood the react reconciler is actually pretty sick and you don't see any flash

PR #869, opened 2022-07-02
Feed refactor

tbezman:


schema.graphql
```diff
@@ -339,7 +339,10 @@ type ViewerGallery {
 type Viewer {
     user: GalleryUser @goField(forceResolver: true)
     viewerGalleries: [ViewerGallery] @goField(forceResolver: true)
+
+    feed(before: String, after: String, first: Int, last: Int): FeedConnection
```
jarrel-b, line 343: Not sure if this actually matters for the frontend, but we'll want `@goField(forceResolver: true)` for this as well

tbezman: nice. makes sense. just a mock to test out my changes on front end. feel free to make whatever changes necessary on backend. I told @kaitoo1 already but I want to make sure we don't make any breaking changes so the website doesn't break for browsers loading the "old" client. 

TL;DR Three commits.
1. Make backend change adding new field to the viewer type
2. Make frontend change to use new field
3. Once 2 is deployed, we can remove the old field.

jarrel-b: ah makes sense!

jarrel-b:
Changes on the backend are now on dev to play with

schema.graphql
Robinnnnn: great call @jarrel-b, will apply!

PR #870, opened 2022-07-03
[GAL-275] Polish pass on Collection Editor (desktop, moweb)

connorrothschild:
## Collection editor view

Figma/dev: (Note: we want to replace the collection description with the # of pieces? just confirming)

image

#### Edit dropdown

Figma/dev: 

**Note: the Figma designs specify an approach to the dropdown that would require a sort of refactor across all dropdowns (change dropdown padding from `12px` -> `8px`, and increase the size of dropdown elements within)**. I think this should be scoped to a different PR, maybe? Or if this dropdown is meant to appear differently, we could style it independent of others

image

image

### Edit page 

**Note: I figured we wanted to keep the "select"/"deselect all" button here?**

Figma/dev:

image

image

### Sidebar

Figma/dev:

**Note: Live display button currently does nothing, so it is commented out.** Unless this functionality is live? I didn't think so. 

image

image

### DIFF
--- a/src/components/core/Dropdown/Dropdown.tsx
+++ b/src/components/core/Dropdown/Dropdown.tsx
@@ -8,7 +8,7 @@ import {
   useState,
 } from 'react';
 import styled from 'styled-components';
-import TextButton from '../Button/TextButton';
+import TextButton, { StyledButton, StyledButtonText } from 'components/core/Button/TextButton';
 import colors from '../colors';
 
 type Props = {
@@ -124,12 +124,12 @@ const StyledDropdownBox = styled.div<StyledDropdownProps>`
   width: max-content;
   background-color: ${colors.white};
 
-  button {
+  ${StyledButton} {
     height: 32px;
     padding: 8px;
   }
 
-  button p {
+  ${StyledButtonText} {
     font-weight: 500;
     color: ${colors.shadow};
   }
--- a/src/components/core/Dropdown/Dropdown.tsx
+++ b/src/components/core/Dropdown/Dropdown.tsx
@@ -117,10 +117,22 @@ const StyledDropdownBox = styled.div<StyledDropdownProps>`
 
   z-index: 1;
 
-  padding: 12px;
-  border: 1px solid ${colors.metal};
+  // min-width: 191px;
+
+  padding: 8px;
+  border: 1px solid ${colors.offBlack};
   width: max-content;
   background-color: ${colors.white};
+
+  button {
+    height: 32px;
+    padding: 8px;
+  }
+
+  button p {
+    font-weight: 500;
+    color: ${colors.shadow};
+  }
 `;
 
 export default Dropdown;
--- a/src/flows/shared/steps/OrganizeCollection/Editor/EditorMenu.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Editor/EditorMenu.tsx
@@ -22,11 +22,6 @@ function EditorMenu() {
 export const MENU_WIDTH = 250;
 
 const StyledTitleS = styled(TitleS)`
-  min-height: 52px;
-  width: 249px;
-  left: 0px;
-  top: 0px;
-  border-radius: 0px;
   padding: 16px 0;
 `;
 
--- a/src/flows/shared/steps/OrganizeCollection/Editor/EditorMenu.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Editor/EditorMenu.tsx
@@ -4,20 +4,32 @@ import { TitleS } from 'components/core/Text/Text';
 import React from 'react';
 import styled from 'styled-components';
 import ColumnAdjuster from './ColumnAdjuster';
+// import LiveDisplayAdjuster from './LiveDisplayAdjuster';
 
 function EditorMenu() {
   return (
     <StyledEditorMenu>
-      <Spacer height={16} />
-      <TitleS>Collection settings</TitleS>
-      <Spacer height={28} />
-      <ColumnAdjuster />
-      <Spacer height={24} />
+      <StyledTitleS>Collection settings</StyledTitleS>
+      <StyledSidebarItem>
+        <ColumnAdjuster />
+      </StyledSidebarItem>
+      {/* <StyledSidebarItem>
+        <LiveDisplayAdjuster />
+      </StyledSidebarItem> */}
     </StyledEditorMenu>
   );
 }
 
 export const MENU_WIDTH = 250;
+
+const StyledTitleS = styled(TitleS)`
+  min-height: 52px;
+  width: 249px;
+  left: 0px;
+  top: 0px;
+  border-radius: 0px;
+  padding: 16px 0;
+`;
 
 const StyledEditorMenu = styled.div`
   display: flex;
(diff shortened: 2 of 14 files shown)


Robinnnnn:
amazing work so far, @connorrothschild ! will answer some Qs here:

> we want to replace the collection description with the # of pieces?

good question, will ask this in the figma. do you have access to make comments in there btw?

> the Figma designs specify an approach to the dropdown that would require a sort of refactor across all dropdowns

yes, let's update the dropdown across the whole app! let's not style it independently. can be this PR or separate.

> I figured we wanted to keep the "select"/"deselect all" button here?

let's actually remove it! sometimes users can have a bajillion NFTs in the sidebar

> live display

yeah lets' keep it commented out

thank you ser!

connorrothschild:
## Dropdown refactor ([changes here](https://github.com/gallery-so/gallery/pull/870/commits/31a2879d2f4e5e7aa4d3dcf39eacaa0bcc99858a))

### Reference



### Dev 



image

image

image

image

The only remaining question is whether the dropdowns should have some `min-width` applied. Jess' prototype in the Figma has a width of 191px for the current dropdown, but this width looks out of place on shorter content e.g. 

image

Once she answers that in the Figma the dropdown refactor is good I believe!

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -117,10 +117,22 @@ const StyledDropdownBox = styled.div<StyledDropdownProps>`
...
-  padding: 12px;
-  border: 1px solid ${colors.metal};
+  // min-width: 191px;
+
+  padding: 8px;
+  border: 1px solid ${colors.offBlack};
   width: max-content;
   background-color: ${colors.white};
+
+  button {
```
kaitoo1, line 127: would prefer to specify the style component that we're overriding where possible

for example if this button style override is for a TextButton component, we can do:
${TextButton} {

}

same for the text

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
```diff
@@ -199,11 +152,7 @@ const Header = styled.div`
   display: flex;
   justify-content: space-between;
   align-items: baseline;
-`;
-
-const StyledSelectButtonWrapper = styled.div`
-  display: flex;
-  justify-content: flex-end;
+  min-height: 52px;
```
kaitoo1, line 155: looks like the min-height here might give a bit too much spacing between the header and searcher.

spec:
Screen Shot 2022-07-06 at 13 30 42

dev:
Screen Shot 2022-07-06 at 13 30 32

src/components/core/Dropdown/Dropdown.tsx
connorrothschild: This was not working:

```
${TextButton} {

}
``` 

I think because `TextButton` is a component with props etc... it returned this error

<details>
<summary>error</summary>

```
Argument of type '({ className, text, onClick, disableTextTransform, disabled, dataTestId, }: Props) => Element' is not assignable to parameter of type 'Interpolation<ThemedStyledProps<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<HTMLDivElement>> & { ...; } & StyledDropdownProps, any>>'.
  Type '({ className, text, onClick, disableTextTransform, disabled, dataTestId, }: Props) => Element' is not assignable to type 'InterpolationFunction<ThemedStyledProps<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<HTMLDivElement>> & { ...; } & StyledDropdownProps, any>>'.
    Types of parameters '__0' and 'props' are incompatible.
      Property 'text' is missing in type 'Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "key" | keyof HTMLAttributes<HTMLDivElement>> & { ...; } & StyledDropdownProps & ThemeProps<...>' but required in type 'Props'.ts(2345)
TextButton.tsx(7, 3): 'text' is declared here.
```

</details>

So I exported the styled button component within `TextButton` and styled that instead ([changes](https://github.com/gallery-so/gallery/pull/870/commits/3f6ece0baae86dded84fbd4eff9f2bfb3df28041))

src/flows/shared/steps/OrganizeCollection/Editor/EditorMenu.tsx
```diff
@@ -1,24 +1,35 @@
...
+      {/* <StyledSidebarItem>
+        <LiveDisplayAdjuster />
+      </StyledSidebarItem> */}
     </StyledEditorMenu>
   );
 }
 
 export const MENU_WIDTH = 250;
 
+const StyledTitleS = styled(TitleS)`
```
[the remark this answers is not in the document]
connorrothschild: You're right! I had hastily copied from the Figma inspect and not removed the unneeded styles

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
connorrothschild: Good catch, the sidebar itself has 16px of padding, so I've made the min-height of the text `calc(52px - 16px)` 

### Updated

<div style="display: flex;">
image

image
</div>

connorrothschild:
Thanks @kaitoo1! 

Waiting on two decisions:
* Whether we render collection descriptions in collection editor, or always show `## pieces`
* Whether dropdowns have a `min-width` applied

Asked both these in Figma - once we have those (and after any other feedback), I think this is good to go

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
```diff
@@ -152,7 +152,7 @@ const Header = styled.div`
   display: flex;
   justify-content: space-between;
   align-items: baseline;
-  min-height: 52px;
+  min-height: calc(52px - 16px); /* Matches the padding of StyledSidebar */
```
kaitoo1, line 155: would it be simpler to add padding-bottom: 16px? (or equivalent <Spacer/>)

Screen Shot 2022-07-06 at 16 32 26

src/flows/shared/steps/OrganizeCollection/Editor/EditorMenu.tsx
```diff
@@ -1,24 +1,30 @@
...
-      <Spacer height={28} />
-      <ColumnAdjuster />
-      <Spacer height={24} />
+      <StyledTitleS>Collection settings</StyledTitleS>
+      <StyledSidebarItem>
+        <ColumnAdjuster />
+      </StyledSidebarItem>
+      {/* <StyledSidebarItem>
+        <LiveDisplayAdjuster />
+      </StyledSidebarItem> */}
```
Robinnnnn, line 17: toss?

connorrothschild:
Updates per Jess's comment in the Figma: 

Figma/dev:

image

image

Our bullet is slightly more bold/prominent than Figma. Do we use bullets anywhere else on the site? Afaik we can't style it using `font-weight` or anything so we might be locked into this appearance unless we replace the `&bull;` with an SVG icon

Robinnnnn (approved):

PR #874, opened 2022-07-07
new Button, ButtonLink components and example usages

holic:
This is a precursor/prerequisite to a larger refactor to help with some SEO and generally make the app more accessible.

This PR moves the current `Button` to `DeprecatedButton` and introduces two new components in its place:
- `Button` with simplified API and better markup
  - moves primary/secondary to a `variant` prop to not collide with the native `type` attribute
  - defaults `type="button"` to ensure these don't submit in the context of a `<form>`
  - removes inner `<h3>` as it doesn't make much sense semantically (maybe an artifact of styled component reuse?)
  - leans on `disabled` attribute for styling instead of an extra selector
  - uses conditional styling for variants instead of separate styled component to simplify rendering logic
- `ButtonLink` that mirrors `Button` but uses `<a>` under the hood for proper linking+accessibility
  - leans on Next.js `Link` for client-side navigation
  - requires `href` prop
  - supports `disabled` in an anchor-compatible, accessible way (aria attribute + negative tab index)
 
Both components maintain strong types/hints for their underlying elements (`<button>` and `<a>`).

These components are demonstrated/used within `LandingPage` as an example of what I am planning to do throughout the app.

This PR also introduces a components page so I could compare the new components visually+functionally with the previous button implementation to make sure they were compatible. It'd be great to expand on this over time into more of a "design system" and "style guide".

image


src/components/core/Button/Button.tsx
```diff
@@ -1,107 +1,118 @@
...
-        <TitleXS color={type === 'primary' ? colors.white : colors.shadow}>{text}</TitleXS>
-      )}
-    </ButtonComponent>
-  );
-}
+// TODO:
+// - should a `loading` button be disabled/non-interactive?
+// - should `Loader` be conditionally inverted based on `variant`?
+// - should `Loader` just always be `mini`?
+// - why is opacity on disabled+loading so much stronger than just disabled?
```
holic, line 13: These are questions I had as I refactored the `Button` while trying to maintain backwards compatibility.

src/components/core/NavLink/NavLink.tsx
```diff
@@ -22,11 +22,11 @@ export default function NavLink({ to, href, children, dataTestId, className, onC
...
       to: to || href,
     });
-  }, [href, to, track]);
 
-  if (onClick) {
-    onClick();
-  }
+    if (onClick) {
+      onClick();
+    }
```
holic, line 28: This seemed like a bug, so I fixed it. Didn't have any usages of `<NavLink onClick={...}>` so it never came up.

src/components/core/Button/DeprecatedButton.tsx
```diff
@@ -0,0 +1,107 @@
+import { memo, useMemo } from 'react';
```
holic, line 1: This whole file is actually a renamed from the previous `Button.tsx` but the diff isn't making that easy to see. Maybe best to look at each commit.

src/scenes/LandingPage/LandingPage.tsx
```diff
@@ -22,41 +19,31 @@ function LandingPage() {
...
-              />
-            </SuppressedHrefWrapper>
+            <ButtonLink
+              href="/home"
+              onClick={() => track('Landing page Explore button click')}
+              dataTestId="explore-button"
+              variant="secondary"
+            >
+              Explore
+            </ButtonLink>
```
holic, line 46: Here's an example of how the upcoming refactor will look for true links. Much cleaner IMO!

holic:
An example of the markup before:
image

And after:
image

src/components/core/Button/Button.tsx
```diff
@@ -1,107 +1,122 @@
...
-  border-style: none;
-
-  height: 32px;
-
+  border: 0;
   cursor: pointer;
 
-  text-transform: uppercase;
   padding: 0 24px;
+  height: 32px;
```
holic, line 30: I don't love fixing the height of a button like this. I tried to move this to be top/bottom padding, but the loading indicator throws this off (it's >1em).

Ideally the loading indicator is 1em so that it matches the text size or, if it needs to be larger, it includes negative margins to avoid the button growing/shrinking in height as `loading` is toggled.

pages/_/components.tsx
```diff
@@ -0,0 +1,123 @@
+import { Button, ButtonLink } from 'components/core/Button/Button';
+import DeprecatedButton from 'components/core/Button/DeprecatedButton';
+import styled from 'styled-components';
+
+export default function DesignPage() {
```
Robinnnnn, line 5: this is an incredible idea! does putting things under `/_/` tell nextjs that this won't be available in prod? i guess it's not a huge deal if it's public

src/components/core/NavLink/NavLink.tsx
Robinnnnn: awesome catch

src/scenes/LandingPage/LandingPage.tsx
Robinnnnn: 😍

pages/_/components.tsx
holic: Nah, it's just security by obscurity. I couldn't think of a name so I went with `_` haha. We could easily add rules in the Next config or something to keep this from being visible in prod, but given that this is all open source, I don't think it's a bad thing that this would exist on the web for reals.

src/components/core/Button/DeprecatedButton.tsx
```diff
@@ -0,0 +1,107 @@
...
+  dataTestId,
+}: Props) {
+  const ButtonComponent = useMemo(
+    () => (type === 'primary' ? StyledPrimaryButton : StyledSecondaryButton),
+    [type]
+  );
+  return (
+    <ButtonComponent
+      // Renaming this prop `buttonStyle` since the `type` prop
+      // already exists for styled components
```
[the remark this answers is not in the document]
holic: This is the old button, this comment was there before :) gonna leave it until I can remove this file entirely.

src/components/core/Button/Button.tsx
```diff
@@ -1,107 +1,122 @@
...
+import { BODY_FONT_FAMILY } from '../Text/Text';
+import Link from 'next/link';
+
+// TODO:
+// - should a `loading` button be disabled/non-interactive?
+// - should `Loader` be conditionally inverted based on `variant`?
+// - should `Loader` just always be `mini`?
+// - why is opacity on disabled+loading so much stronger than just disabled?
+// - does ButtonLink need to handle absolute URLs separately? or does Link do that?
+//   - should ButtonLink also add rel="noopener,noreferer" to target="_blank" links?
```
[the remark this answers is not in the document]
holic: > > does ButtonLink need to handle absolute URLs separately? or does Link do that?
> 
> i think buttons should handle both relative and absolute URLs, where absolute URLs always open in a new tab (noopener/noreferrer)

I found in #877 that Next's `Link` handles external links already (I believe it used to only work for local links, or maybe the React Router one didn't handle external ones). I think it makes sense to not have a hard requirement that all external links open in a new window, but I'll make sure the right `rel` is set if so.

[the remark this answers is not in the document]
holic: > > should `Loader` just always be `mini`?
> 
> if we don't have any other usages, then yes

We have both, but they are visually similar enough that IMO we could just use one. The button size is the same in either case, just the loading indicator changes. I would prob opt to have a 1em tall loading indicator and that be the only option (i.e. remove `mini`).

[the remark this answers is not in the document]
Robinnnnn: sick, let's use mini!

pages/_/components.tsx
Robinnnnn: love it! great idea to use `_`

src/components/core/Button/Button.tsx
```diff
@@ -1,107 +1,145 @@
...
-      {loading ? (
-        <Loader inverted size={mini ? 'mini' : 'small'} />
-      ) : (
-        <TitleXS color={type === 'primary' ? colors.white : colors.shadow}>{text}</TitleXS>
-      )}
-    </ButtonComponent>
-  );
-}
+// TODO:
+// - why is opacity on disabled+pending so much stronger than just disabled?
```
holic, line 10: it's really weird but this seems to have something to do with the animation of the spinner inside the button, not sure if this is a browser thing or a styled-components thing

holic: seems like a Chrome bug: https://twitter.com/kingersoll/status/1545309693022593024

holic:
Latest screenshot of components page with opacity fixes:

image

PR #877, opened 2022-07-07
Add/fix anchors throughout, migrate to new Button

holic:
This PR:
- moves `DeprecatedButton` usages to new `Button` from #874 
- adds proper anchor tags with `href`s where applicable
- fixes `NftPreview` anchor (didn't use `passHref` and the href was wrong anyway)

Most of this came from walking around the site, page type by page type, and making sure each element that was interactive (via click) that also triggered a page navigation had a proper anchor tag set up. I also swapped out the old buttons, but wasn't able to test all of these. I'm pretty confident in these changes but I could use some eyes on the pages/components in this PR to make sure things are still working as expected.

Things I found while working through this that we should follow up on:
- remove spacing around buttons in dropdown menus so they're easier to click/harder to miss
- make dropdowns and modals escapeable
- add loading indicator to page navigation? some link clicks are slow (e.g. back to home) and hard to know it's doing anything
  - preload these as a short term quick fix?
- reduce memo/useMemo usages
- move submit buttons into forms
- replace `dataTestId` with DOM attribute passthrough (e.g. `data-testid`)


Robinnnnn:
> add loading indicator to page navigation? some link clicks are slow (e.g. back to home) and hard to know it's doing anything. preload these as a short term quick fix?

let's definitely preload if users hover over links. relay should have a suggested way to do this. @tbezman any thoughts?

src/components/NftPreview/NftPreview.tsx
```diff
@@ -45,32 +45,38 @@ function NftPreview({
...
         collection @required(action: THROW) {
+          # TODO: what's the difference between id and dbid?
           id
           dbid
+          # TODO: can we make these required in the schema?
+          gallery @required(action: THROW) {
+            owner @required(action: THROW) {
+              username @required(action: THROW)
+            }
+          }
```
Robinnnnn, line 56: first off, you can def mark them as required as it's true that collections will always have a gallery / owner / username

as for why the schema isn't labeled `Gallery!`:

image

i believe it's because it's marked as `@goField(forceResolver: true)` which is a way of denoting "we'll only resolve this data if you ask for it" bc it triggers a sub-query on the backend. whereas if it was required then the backend might always try to grab it.

@radazen can you confirm?

Robinnnnn: (in the meantime ezra i shared the link to your academy award winning graphql/api presentation)

holic: I mentioned this to @Robinnnnn already but I think the thing that had me confused is that I am used to viewing the GraphQL schema as more of a guarantee on what data is stored in the backend rather than a guarantee on how the data is retrieved.

For example, it's unclear if a "gallery has an optional owner" means there are gaps data for some reason (technical debt, or a feature I'm unaware of). And inversely, if an optional field is truly meant to be optional, it's probably not a great idea for me to willy-nilly add `@required`. So I'm left wondering: what are our actual data relationships and what guarantees are there in those relationships (i.e. nullable or non-nullable).

holic: Here's an example where I'm at a loss around what to do:

image
image

I didn't modify/affect this code AFAIK, but getting a runtime error due to a query requiring a prop/data that appears to be missing? Makes it hard to know which things are safe to require or not.

Robinnnnn: looping in @tbezman for the first half of the convo :P

re: erroring feed events it's bc the user who created those events were deleted on dev. will address separately

tbezman: @holic 

> For example, it's unclear if a "gallery has an optional owner" means there are gaps data for some reason (technical debt, or a feature I'm unaware of). And inversely, if an optional field is truly meant to be optional, it's probably not a great idea for me to willy-nilly add @required. So I'm left wondering: what are our actual data relationships and what guarantees are there in those relationships (i.e. nullable or non-nullable).

I try to think about "null" as a cop out for "optional". Any place in our schema where something might not be there for a "good reason", we use [union types in GraphQL](https://graphql.org/learn/schema/#union-types) to represent that. [Here's an example where we do that in our schema](https://github.com/gallery-so/gallery/blob/main/schema.graphql#L360).

At Gallery, `null` is used to represent "unhandled error", or an error that the client' can't do anything reasonable with. Something that would make the client throw their hands up and show a "Something went wrong" message.

The Relay team has an excellent write up [here](https://relay.dev/docs/guides/required-directive/#why-not-implement-this-at-the-schemaserver-level).

---

On a separate note, I would encourage you to try not to use `@required` if possible. Try and think "If the server can't get me this field, can I just not show the HTML for that field? Can I show a sane fallback instead?". 

`@required` is the least graceful method of handling a failure.

---

Honestly a lot of this stuff comes from dealing with the alternatives. At this point in time w/ GraphQL & Relay, this is the most sane way to handle this. I would be happy to get on a call with you and explain how we arrived at this conclusion.

tbezman: @Robinnnnn this will most likely continue to be a pain point for engineers working on the codebase. We should write something up explaining the philosophy to avoid these long conversations in the future.

holic: > I would be happy to get on a call with you and explain how we arrived at this conclusion.

Would love this! I've not worked within this paradigm and the context would be super helpful.

src/scenes/CollectionGalleryPage/CollectionGalleryHeader.tsx
```diff
@@ -135,8 +130,14 @@ function CollectionGalleryHeader({
           <StyledBreadcrumbsWrapper>
             <StyledUsernameWrapper>
               <StyledUsernameAndSeparatorWrapper>
-                <StyledUsernameMobile onClick={handleBackClick}>{username}</StyledUsernameMobile>
-                {collection.name && <StyledSeparatorMobile>/</StyledSeparatorMobile>}
+                {username ? (
+                  <UnstyledLink href={`/${username}`}>
```
holic, line 134: Because this is a link to a profile/username, I didn't think it was necessary to pipe this through a `useBackButton` handler, but let me know if I missed something.

src/scenes/UserGalleryPage/UserGalleryCollection.tsx
```diff
@@ -69,34 +69,22 @@ function UserGalleryCollection({ queryRef, collectionRef, mobileLayout }: Props)
...
-  const username = asPath.split('/')[1];
-  const collectionUrl = `${baseUrl}/${username}/${collection.dbid}`;
-
-  const handleViewCollectionClick = useCallback(
-    (event: React.MouseEvent<HTMLElement>) => {
-      navigateToUrl(`/${username}/${collection.dbid}`, event);
-    },
-    [collection.dbid, navigateToUrl, username]
-  );
+  const username = router.query.username as string;
```
holic, line 80: This is a slightly more robust way of getting the username from the URL than before (parsing it from the URL).

However, it's a little scary to take either approach outside the context of a Next.js page component, where we implicitly expect this component to only be used inside the context of a user page. I'd suggest in cases like this, we pass things like the username in as a prop.

src/utils/navigate.ts
```diff
@@ -1,18 +0,0 @@
-import { useCallback, MouseEvent } from 'react';
-import { useRouter } from 'next/router';
-
-// navigates to the given url. Opens in new tab if meta key (command on mac) is pressed
-export function useNavigateToUrl() {
```
holic, line 5: This wasn't used as far as I could tell, and we should really prefer anchor tags rather than this.

holic:
@Robinnnnn I tagged you for review, feel free to tag others you think would be good to review too.

src/components/NftPreview/NftPreview.tsx
holic: @tbezman Just had a thought along this line, and correct me if I am wrong, but because we specify fragments for our components' data needs, we're no longer in control of when/where that query (i.e. a batch of fragments) gets run, therefore if we have `@required(action: THROW)`, we risk a whole ~page worth of data not returning due to one deep (and potentially inconsequential) field not resolving. Is that an accurate picture of why `@required(action: THROW)` is kind of a bad habit here?

tbezman: Yeah that's definitely a good point to bring  up. You may not know where the nearest Error Boundary is (relative to this component).

I think the easiest and most convincing question you can ask yourself is "Is there something better than an error here?"

I'll reach out on Discord to setup a time.

src/contexts/globalLayout/GlobalNavbar/LoggedInNav.tsx
```diff
@@ -95,10 +93,10 @@ function LoggedInNav({ queryRef }: Props) {
...
         <StyledDropdownWrapper hasNotification={false}>
           <Dropdown mainText={username || 'ACCOUNT'} shouldCloseOnMenuItemClick>
-            <TextButton
-              text="My Gallery"
-              onClick={username ? () => push(`/${username}`) : undefined}
-            />
+            <UnstyledLink href={`/${username}`}>
+              <TextButton text="My Gallery" />
+            </UnstyledLink>
+            <Spacer height={12} />
```
Robinnnnn, line 99: not sure where this spacer came from but it adds an unwanted space in the dropdown. i'll toss it!

src/scenes/CollectionGalleryPage/CollectionGalleryHeader.tsx
Robinnnnn: ahh yeah we wanna keep the back functionality mainly bc if you're scrolled down halfway through the page, then click "back" to the username, you want to end up back where you were. in contrast, just directing them to `/username` puts them back at the top of the page

Robinnnnn: i realize this is quite hacky

src/scenes/UserGalleryPage/UserGalleryCollection.tsx
Robinnnnn: great point!

Robinnnnn (approved):

src/scenes/CollectionGalleryPage/CollectionGalleryHeader.tsx
holic: We chatted about this over Discord and decided it's ~fine to not have "back button" behavior here. If we want to reintroduce this sort of behavior (thinking about a Twitter timeline where you might click into a tweet and hitting a back arrow/link ideally would take you to where you left off in your timeline view), we can do it in a more generic way rather than wiring it up to every click handler.

PR #881, opened 2022-07-08
Collection editor design (mobile)

connorrothschild:
Figma/dev:

image

image

Of note: 
* We removed the width of 777px and instead apply `width: 100%` and a `max-width: 777px` instead. There are 16px of horizontal padding on the collection edit.
* On mobile, we remove the `remainingNFTs` from desktop, and only show the trimmed NFTs.

Do we have a design for (or plans to in the future) make the internal edit collection page accessible on mobile as well? right now it looks like this

image


Robinnnnn (approved):
🔥🔥🔥

PR #883, opened 2022-07-10
[GAL-282] Improve onboarding, user creation flow

Robinnnnn:
## Overview

**This PR requires that a username is passed in on user creation, rather than preemptively creating users with empty usernames during initial onboarding.**

The impetus is twofold:
1. Users with empty usernames (`''`) are still technically authenticated users who could end up in other areas of the app. However, with empty usernames, they are unable to access their main profile page, since our URL scheme is `gallery.so/[username]` aka. `gallery.so/''`. Unless they know how to get themselves back on the onboarding flow to set up their usernames, it's difficult for affected users to recover.
2. Since there can theoretically be multiple users in the database with empty usernames (e.g. lots of users signing up simultaneously), we cannot add a uniqueness constraint to the username field. This led to a separate bug, where we saw multiple users with the same username get created in our DB.

The easiest way to envision the change is to realize that `useLoginOrCreateUser` has been renamed to `useLoginOrRedirectToOnboarding`.

image

image

## What changed?

Comments are sprinkled in amongst the files, but to summarize:

- you no longer need to be authenticated to access the `/welcome` onboarding route, since a user will not be created prior to the visit. however, you'll need to have been redirected through the auth flow.
- username validation now occurs as you type, rather than waiting until you submit
  - the user's cryptographically signed message will not be used immediately to create a new user; instead, it will be forwarded to the onboarding flow where it's submitted once the user decides on a username. since the signature is a one-time key, we want to avoid prompting the user to sign a new nonce upon every user creation attempt.
  - **as a result, we've added real-time username availability checking as the user types into the input, so that when they click "save" it's highly likely that the username is available and their signature is only used once.**
  - this happens to result in an improved UX as well.
- ❗ previously, we began fetching the user's NFTs as soon as onboarding began, since we could assume they were logged in (without a username) by that point. however, we now need to wait to begin fetching until they select their username, which is further into the flow. this will result in less time for NFTs to be fetched before they reach the Collection Creation step.

## TODOs / Tests
- [x] Update schema/mutation to send `bio` field
- [x] Ensure `/welcome` is inaccessible by default
- [x] Trigger onboarding flow when logging in through the navbar
- [x] No regressions for user edit flow for existing users
- [x] Ensure new UX around user creation is chill ([loom](https://www.loom.com/share/d3be97b27871441ea9859e67f9603137))

## Fast Follows
- ❗️ Migrate all users with empty usernames to placeholders. This is because users with empty usernames will no longer be able to access the onboarding flow to "edit" their usernames, as the onboarding flow now only supports User Creations

## Slow Follows
- User creation for gnosis safe is not hooked up. Will make a ticket for this
- This is a longstanding issue, but "unexpected" mutation error messages are not bubbling up to be displayed on the client. Will make a ticket for this


pages/welcome.tsx
```diff
@@ -1,24 +1,15 @@
...
-      }
-    `,
-    {}
-  );
+  // The onboarding flow can only be accessed if the user has been redirected from the auth pipeline,
+  // which populates the internal URL queries with necessary data to create a profile
+  const authPayloadQuery = useAuthPayloadQuery();
+  if (!authPayloadQuery) {
+    return <GalleryRedirect to="/" />;
+  }
```
Robinnnnn, line 12: you no longer need to be authenticated to access the `/welcome` onboarding route, since a user will not be created prior to the visit. however, you'll need to have been redirected through the auth flow.

src/components/Profile/useUserInfoForm.ts
```diff
@@ -12,14 +13,45 @@ import {
...
 
 type Props = {
   onSuccess: (username: string) => void;
   existingUsername?: string;
   existingBio?: string;
-  userId: string;
+  userId?: string;
 };
 
+function useIsUsernameAvailableFetcher() {
```
Robinnnnn, line 27: the user's cryptographically signed message will not be used immediately to create a new user; instead, it will be forwarded to the onboarding flow where it's submitted once the user decides on a username. since the signature is a one-time key, we want to avoid prompting the user to sign a new nonce upon every user creation attempt. **as a result, we've added real-time username availability checking as the user types into the input, so that when they click "save" it's highly likely that the username is available and their signature is only used once.** this happens to result in an improved UX as well.

src/components/Profile/useUserInfoForm.ts
```diff
@@ -28,63 +60,95 @@ export default function useUserInfoForm({
...
-  const handleClearUsernameError = useCallback(() => {
-    setUsernameError('');
+  const handleUsernameChange = useCallback((username: string) => {
+    setUsername(username);
+    usernameFieldIsDirty.current = true;
   }, []);
 
+  const isUsernameAvailableFetcher = useIsUsernameAvailableFetcher();
+
+  const debouncedUsername = useDebounce(username, 500);
```
Robinnnnn, line 111: wait a bit before asking the server whether a username is taken (to prevent spamming it)

src/components/Profile/useUserInfoForm.ts
```diff
@@ -28,63 +60,95 @@ export default function useUserInfoForm({
...
+      return { success: false };
     }
-  }, [username, bio, updateUser, userId, onSuccess]);
+  }, [usernameError, bio, userId, authPayloadQuery, onSuccess, username, updateUser, createUser]);
 
-  const handleClearUsernameError = useCallback(() => {
-    setUsernameError('');
+  const handleUsernameChange = useCallback((username: string) => {
+    setUsername(username);
+    usernameFieldIsDirty.current = true;
```
Robinnnnn, line 106: set whether the field is dirty so the user doesn't immediately see errors

src/components/WalletSelector/mutations/useLoginOrCreateUser.ts
```diff
@@ -1,111 +0,0 @@
-import { usePromisifiedMutation } from 'hooks/usePromisifiedMutation';
```
Robinnnnn, line 1: renamed this file to `useLoginOrRedirectToOnboarding`

src/components/WalletSelector/mutations/useLoginOrRedirectToOnboarding.ts
```diff
@@ -0,0 +1,90 @@
...
+              query: {
+                chain: 'Ethereum',
+                address: authMechanism.mechanism.eoa.chainAddress.address,
+                nonce: authMechanism.mechanism.eoa.nonce,
+                signature: authMechanism.mechanism.eoa.signature,
+              },
+            },
+            '/welcome'
+          );
+        }
```
Robinnnnn, line 76: the query params listed here will not be visible in the URL

src/contexts/wizard/WizardDataProvider.tsx
```diff
@@ -74,14 +65,6 @@ export default memo(function WizardDataProvider({ id, children }: Props) {
     setIsRefreshingNfts(false);
   }, [loadQuery, pushToast, syncTokens]);
 
-  useEffect(() => {
-    if (id === 'onboarding') {
-      void handleRefreshNfts();
-    }
-    // just want this running once for the welcome screen
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
```
Robinnnnn, line 83: ❗ previously, we began fetching the user's NFTs as soon as onboarding began, since we could assume they were logged in (without a username) by that point. however, we now need to wait to begin fetching until they select their username, which is further into the flow. this will result in less time for NFTs to be fetched before they reach the Collection Creation step.

the fetching now begins here: https://github.com/gallery-so/gallery/pull/883/files#diff-227e443aa0348b7881e6bd7e3256beb1c80f48335332b6764b002a5fae96326aR55

src/flows/OnboardingFlow/steps/AddUserInfo.tsx
```diff
@@ -25,51 +25,36 @@ function useWizardConfig({ onNext }: ConfigProps) {
...
-    existingUsername: viewer.user.username ?? undefined,
-    existingBio: viewer.user.bio ?? undefined,
-  });
+  useEffect(() => {
+    if (username.length < 2) {
+      setNextEnabled(false);
+      return;
+    }
+    setNextEnabled(!usernameError);
+  }, [setNextEnabled, username, usernameError]);
```
Robinnnnn, line 44: ensure the "next" button is disabled until the username is confirmed to be valid.

we didn't need this before because we actually needed the user to click the "next" button since errors were displayed _on-submit_, rather than on-type.

src/hooks/api/users/useCreateUser.ts
```diff
@@ -0,0 +1,105 @@
+import { usePromisifiedMutation } from 'hooks/usePromisifiedMutation';
+import { useCallback } from 'react';
+import { graphql } from 'relay-runtime';
+import { AuthPayloadVariables } from './useAuthPayloadQuery';
+
+export default function useCreateUser() {
+  //   const [createUser] = usePromisifiedMutation<any>(
```
Robinnnnn, line 7: will un-comment once we can send `bio` as a field

PR #890, opened 2022-07-15
Liverender

Robinnnnn:
**Add support for displaying interactive NFTs live / HQ on Gallery.**

- token settings are converted from an array to an object to keep track of in local state (`CollectionEditorContext`), then converted back to an array when sent to the server
- remove apply-all option from sidebar as it could potentially incentivize thoughtless curation and slow performance

See comments for additional details / context.

**Other small changes**
- fix GLB file types and ensure they're interactive
- fix rendering issues on firefox / safari / chrome
- introduce new `TitleXSBold` text component, currently used for tooltip
- replace "REMOVE" with trash icon


src/components/Feed/Events/FeedEventNftPreviewWrapper.tsx
```diff
@@ -69,16 +69,18 @@ function FeedEventNftPreviewWrapper({ tokenRef, maxWidth, maxHeight }: Props) {
     >
       <NftPreview
         tokenRef={token}
-        nftPreviewWidth={'100%'}
         previewSize={maxWidth}
         onClick={handleClick}
         hideLabelOnMobile={isMobile}
+        disableLiverender
```
Robinnnnn, line 75: disable liverender on main feed to conserve perf

PR #895, opened 2022-07-18
[GAL-269] Fix collector notes overlap on Audio NFT

jakzaizzat:
## Finding

Find out that the bug only happens on audio NFT that have `<audio>` tag and used the`desktop` breakpoints. It didn't happen on tablet and mobile view

## Solution

Add spacing if its desktop view and only applicable to audio NFT components

## Screenshot

**Before**

image

**After**

CleanShot 2022-07-18 at 09 10 56


Robinnnnn (approved):
sickkkk

PR #897, opened 2022-07-18
[GAL-257] Update NFT "Could not load" state

jakzaizzat:
## Changes Implemented

- [x] Sidebar edit collection
- [x] Staged NFT View
- [x] User Gallery

## Updates

- Added `success` flag to `getVideoOrImageUrlForNftPreview` and `graphqlGetResizedNftImageUrlWithFallback(`
- New `<FailedNftPreview/>` component

### Sidebar 

CleanShot 2022-07-18 at 15 58 14

### Staged View

https://user-images.githubusercontent.com/4480258/179468311-ad950c1a-2118-406a-a3f0-86f74b203a07.mp4

### User Gallery

CleanShot 2022-07-18 at 16 03 01


src/components/NftPreview/NftPreview.tsx
```diff
@@ -98,11 +104,16 @@ function NftPreview({
           width={nftPreviewWidth}
           backgroundColorOverride={backgroundColorOverride}
         >
-          <NftPreviewAsset
-            tokenRef={token}
-            // we'll request images at double the size of the element so that it looks sharp on retina
-            size={previewSize * 2}
-          />
+          {token.media?.__typename === 'UnknownMedia' ? (
```
jakzaizzat, line 107: @Robinnnnn Is there any edge cases that I'm missing by checking the media type? 

Lmk if we would like to change on how to validate failed nft. 

Other thing I can think of is by using `graphqlGetResizedNftImageUrlWithFallback` function

Robinnnnn: @jakzaizzat so we basically need to use this: `getVideoOrImageUrlForNftPreview`, which might be the function you're referring to, to see whether a URL is available or not.

`UnknownMedia` does not necessarily mean we don't have a URL... it simply means the backend failed to categorize what kind of asset it was. BUT it could definitely still have a URL that is render-able!

src/components/NftPreview/NftPreview.tsx
```diff
@@ -120,14 +126,19 @@ function NftPreview({
     if (shouldLiverender && token.media?.__typename === 'HtmlMedia') {
       return <NftDetailAnimation mediaRef={token} />;
     }
+
+    if (token.media?.__typename === 'UnknownMedia') {
+      return <FailedNftPreview onLoad={setContentIsLoaded} />;
+    }
```
Robinnnnn, line 132: hmm so `UnknownMedia` does not necessarily mean the asset lacks URLs. the backend can still send URLs even for this type:
image

so i don't think we need to show this 🤔 unless this fixes an issue?

src/flows/shared/steps/OrganizeCollection/Editor/StagedNftImage.tsx
```diff
@@ -21,13 +22,28 @@ function StagedNftImage({ tokenRef, size, hideLabel, setNodeRef, ...props }: Pro
...
+        }
         ...getVideoOrImageUrlForNftPreviewFragment
       }
     `,
     tokenRef
   );
 
   const reportError = useReportError();
+
+  if (token.media?.__typename === 'UnknownMedia') {
```
Robinnnnn, line 36: same comment as above

src/flows/shared/steps/OrganizeCollection/Sidebar/SidebarNftIcon.tsx
```diff
@@ -73,6 +77,15 @@ function SidebarNftIcon({ tokenRef, EditModeToken }: SidebarNftIconProps) {
     [contractAddress]
   );
 
+  if (token.media?.__typename === 'UnknownMedia') {
```
Robinnnnn, line 80: same comment

src/components/NftPreview/NftPreview.tsx
jakzaizzat: Thanks for the info robin! I misunderstood this `UnknownMedia` type

I've updated the `getVideoOrImageUrlForNftPreview` with additional flag

Robinnnnn (approved):
wooooo thank you jakz!!! 👑

PR #909, opened 2022-07-26
[GAL-303] Feed: user preview card on hover

jakzaizzat:
## Changes
- Add `<HoverCardOnUsername/>` component
- Some query fragment passed into sub component for follow button 

## Not logged in user
https://user-images.githubusercontent.com/4480258/180928501-015f0980-ef37-4010-bf91-29590a6bd428.mp4

## Logged in user

https://user-images.githubusercontent.com/4480258/180928629-5a58a049-8d0e-4c69-b8ff-779b41a0a537.mp4

### Notes
In future, we can abstract into  `popover` component when we want to use it on other use-case likes collection hover, etc

### DIFF
--- a/src/components/HoverCard/HoverCardOnUsername.tsx
+++ b/src/components/HoverCard/HoverCardOnUsername.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useCallback, useState } from 'react';
 import colors from 'components/core/colors';
 import InteractiveLink from 'components/core/InteractiveLink/InteractiveLink';
 import { BaseM, TitleM } from 'components/core/Text/Text';
@@ -13,6 +13,7 @@ import transitions, {
 } from 'components/core/transitions';
 import { HoverCardOnUsernameFollowFragment$key } from '__generated__/HoverCardOnUsernameFollowFragment.graphql';
 import { useLoggedInUserId } from 'hooks/useLoggedInUserId';
+import { useRouter } from 'next/router';
 
 type Props = {
   userRef: HoverCardOnUsernameFragment$key;
@@ -21,6 +22,8 @@ type Props = {
 
 export default function HoverCardOnUsername({ userRef, queryRef }: Props) {
   const [isHovering, setIsHovering] = useState(false);
+
+  const router = useRouter();
 
   const user = useFragment(
     graphql`
@@ -59,6 +62,14 @@ export default function HoverCardOnUsername({ userRef, queryRef }: Props) {
     setIsHovering(false);
   };
 
+  const handleClick = useCallback(
+    (e) => {
+      e.preventDefault();
+      router.push(`${user.username}`);
+    },
+    [user]
+  );
+
   const loggedInUserId = useLoggedInUserId(query);
   const isLoggedIn = !!loggedInUserId;
 
@@ -68,7 +79,7 @@ export default function HoverCardOnUsername({ userRef, queryRef }: Props) {
         <InteractiveLink to={`/${user.username}`}>{user.username}</InteractiveLink>
       </StyledLinkContainer>
       {isHovering && (
-        <StyledCardWrapper isHovering={isHovering}>
+        <StyledCardWrapper isHovering={isHovering} onClick={handleClick}>
           <StyledCardContainer>
             <StyledCardHeader>
               <StyledHoverCardTitleContainer>
--- a/src/components/HoverCard/HoverCardOnUsername.tsx
+++ b/src/components/HoverCard/HoverCardOnUsername.tsx
@@ -69,23 +69,29 @@ export default function HoverCardOnUsername({ username, userRef, queryRef }: Pro
         <InteractiveLink to={`/${username}`}>{username}</InteractiveLink>
       </StyledLinkContainer>
 
-      <StyledCardContainer isHovering={isHovering}>
-        <StyledCardHeader>
-          <StyledHoverCardTitleContainer>
-            {isLoggedIn && <FollowButton userRef={user} queryRef={query} />}
-            <StyledCardUsername>{user.username}</StyledCardUsername>
-          </StyledHoverCardTitleContainer>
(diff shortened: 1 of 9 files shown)


src/components/HoverCard/HoverCardOnUsername.tsx
```diff
@@ -0,0 +1,168 @@
...
+import unescape from 'lodash/unescape';
+import FollowButton from 'components/Follow/FollowButton';
+import transitions, {
+  ANIMATED_COMPONENT_TRANSLATION_PIXELS_LARGE,
+} from 'components/core/transitions';
+import { HoverCardOnUsernameFollowFragment$key } from '__generated__/HoverCardOnUsernameFollowFragment.graphql';
+import { useLoggedInUserId } from 'hooks/useLoggedInUserId';
+
+type Props = {
+  username: string;
```
kaitoo1, line 18: could we get the username from the user fragment instead of passing it in as a prop?

jakzaizzat: Definitely! Overlooked on this

src/components/HoverCard/HoverCardOnUsername.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const loggedInUserId = useLoggedInUserId(query);
+  const isLoggedIn = !!loggedInUserId;
+
+  return (
+    <StyledContainer onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseLeave}>
+      <StyledLinkContainer>
+        <InteractiveLink to={`/${username}`}>{username}</InteractiveLink>
+      </StyledLinkContainer>
+
+      <StyledCardWrapper isHovering={isHovering}>
```
kaitoo1, line 71: does the hover behavior work if we only displayed `StyledCardWrapper` if `isHovering` is true?
like this:
{isHovering && <StyledCardWrapper>...</> }
It looks like we render all the card wrappers in the feed and set opacity to 1 if the user hovers on one, but maybe it's possible to make it so that we only render a single card wrapper when the user hovers on a name?

src/components/HoverCard/HoverCardOnUsername.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const loggedInUserId = useLoggedInUserId(query);
+  const isLoggedIn = !!loggedInUserId;
+
+  return (
+    <StyledContainer onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseLeave}>
+      <StyledLinkContainer>
+        <InteractiveLink to={`/${user.username}`}>{user.username}</InteractiveLink>
+      </StyledLinkContainer>
+
+      <StyledCardWrapper isHovering={isHovering}>
```
kaitoo1, line 71: also I think the card should be clickable, and navigate to the user's gallery

Screen Shot 2022-07-26 at 14 56 28

kaitoo1:
THIS IS SO COOL
awesome addition to the feed, nice work @jakzaizzat !
thanks for addressing one comment already, just left two more

src/components/HoverCard/HoverCardOnUsername.tsx
jakzaizzat: Yes it works. We can do that way too. 

Might need to change a little bit on how the animation appeared

jakzaizzat: Updated it 👌

src/components/HoverCard/HoverCardOnUsername.tsx
jakzaizzat: Added clickable link to the card 🙏🏻

jakzaizzat:
> THIS IS SO COOL awesome addition to the feed, nice work @jakzaizzat ! thanks for addressing one comment already, just left two more

Thanks kaito! Fix both of the comments. Lmk if I missed anything

kaitoo1 (approved):
Nice, changes look great, thanks for addressing 🙏

Robinnnnn:
incredible work, @jakzaizzat !!! 🔥🔥🔥 

a couple things on UI:
- i think the top-padding is a bit large compared to the mocks.
- the animation looks a bit strong. wonder if we can reduce it and add opacity transition.

Screen Shot 2022-07-26 at 11 38 49 AM

i'll take a stab at fixing real quick!

Robinnnnn:
here's my commit! https://github.com/gallery-so/gallery/pull/909/commits/c575cad188882cfddd4560fed11363d7845241f7

https://user-images.githubusercontent.com/12162433/181060265-0e443475-ded6-4e0c-a006-f6cf47eccda3.mov

### Updated animation
The translation up/down is a little bit smaller, and opacity transition works. Both translation + opacity changes are now controlled with `transition:`, instead of CSS keyframe animations.

image

### Moved CardWrapper outside of renderer

Having the wrapper outside allows us to apply styles to it before the children get mounted.

```js
// before
{isHovering &&
  <CardWrapper isHovering={isHovering}> <-- problem: isHovering is ALWAYS true for this component
    <contents />
  </CardWrapper>}

// after
<CardWrapper isHovering={isHovering}>
  {isHovering && <contents>}
</CardWrapper>
```


### New state
we have two states: `isHovering` and `isActive`. isHovering determines whether the user is hovering over the component, while isActive determines whether the component should be mounted or not. without this, upon "mouse leave" the component immediately disappears, instead of fading out.
- before: mouse leave => immediately unmount component, no chance to display fade-out
- after: mouse leave => begin fade-out => setTimeout for fadeout duration => unmount component

jakzaizzat:
Sick improvement @Robinnnnn  ! So brilliant on how to tacke the animation state

Thanks for the sick update!

Robinnnnn:
> Sick improvement @Robinnnnn ! So brilliant on how to tacke the animation state
> 
> Thanks for the sick update!

woohoo! will merge + deploy, thank you jakz!!

PR #916, opened 2022-07-29
[GAL-305] Virtualize Sidebar

jakzaizzat:
### Changes 
Apply react virtualize to the sidebar in edit gallery page.

**Notes on how it works**

Utilize the `List]` component but added 3 items in each row.

| blank | 0 | 1 |
|-------|---|---|
| 2     | 3 | 4 |
| 5     | 6 | 7 |

**Demo**

https://user-images.githubusercontent.com/4480258/181875673-be42f6ff-f0ac-408a-b5af-f17aa9e92ba9.mp4

### DIFF
--- a/src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
@@ -204,7 +204,7 @@ const SidebarTokens = ({ nftFragmentsKeyedByID, tokens }: SidebarTokensProps) =>
     // add blank block button at the beginning of list
     if (index === 0) {
       items.push(
-        <StyledAddBlankBlock onClick={handleAddBlankBlockClick} size={SIDEBAR_ICON_DIMENSIONS}>
+        <StyledAddBlankBlock onClick={handleAddBlankBlockClick}>
           <StyledAddBlankBlockText>Add Blank Space</StyledAddBlankBlockText>
         </StyledAddBlankBlock>
       );
--- a/src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
@@ -62,8 +62,6 @@ function Sidebar({ tokensRef, sidebarTokens, viewerRef }: Props) {
 
   const tokens = removeNullValues(allTokens);
 
-  const { stageTokens } = useCollectionEditorActions();
-
   const [debouncedSearchQuery, setDebouncedSearchQuery] = useState('');
   const [searchResults, setSearchResults] = useState<string[]>([]);
 
@@ -92,15 +90,6 @@ function Sidebar({ tokensRef, sidebarTokens, viewerRef }: Props) {
     );
   }, [nftFragmentsKeyedByID, tokensFilteredBySearch]);
 
-  const handleAddBlankBlockClick = useCallback(() => {
-    const id = `blank-${generate12DigitId()}`;
-    stageTokens([{ id, whitespace: 'whitespace' }]);
-    // auto scroll so that the new block is visible. 100ms timeout to account for async nature of staging tokens
-    setTimeout(() => {
-      document.getElementById(id)?.scrollIntoView({ behavior: 'smooth' });
-    }, 100);
-  }, [stageTokens]);
-
   const { isRefreshingNfts, handleRefreshNfts } = useWizardState();
 
   return (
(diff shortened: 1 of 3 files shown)


src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
```diff
@@ -196,17 +192,62 @@ const SidebarTokens = ({ nftFragmentsKeyedByID, tokens }: SidebarTokensProps) =>
...
+    return (
+      <Selection key={key} style={style}>
+        {items}
+      </Selection>
+    );
+  };
+
+  return (
+    <StyledListTokenContainer>
+      <AutoSizer>
```
jakzaizzat, line 239: Didn't implement any `CellMeasurer` since we already have a fixed height and width of the child which is 60 x 60 - `SidebarNftIcon`

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
```diff
@@ -196,17 +192,62 @@ const SidebarTokens = ({ nftFragmentsKeyedByID, tokens }: SidebarTokensProps) =>
...
+    // add blank block button at the beginning of list
+    if (index === 0) {
+      items.push(
+        <StyledAddBlankBlock onClick={handleAddBlankBlockClick}>
+          <StyledAddBlankBlockText>Add Blank Space</StyledAddBlankBlockText>
+        </StyledAddBlankBlock>
+      );
+    }
+
+    const fromIndex = index === 0 ? 0 : index * COLUMN_COUNT - 1;
```
jakzaizzat, line 210: This part is kind of hacky. Would love to know any thought if its could be better

Robinnnnn: no problem :) unfortunately this is the result of our hacky setup with the whitespace adder

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
```diff
@@ -196,17 +192,62 @@ const SidebarTokens = ({ nftFragmentsKeyedByID, tokens }: SidebarTokensProps) =>
...
+  };
+
+  return (
+    <StyledListTokenContainer>
+      <AutoSizer>
+        {({ width, height }) => (
+          <List
+            rowRenderer={rowRenderer}
+            rowCount={Math.ceil(displayedTokens.length / COLUMN_COUNT)}
+            rowHeight={60 + 19} // height of SidebarNftIcon 60 + gap
```
Robinnnnn, line 244: nit: make these constants. `60` and `19` are used multiple times in this file, could probably use

```js
const SIDEBAR_ICON_DIMENSIONS = 60
const SIDEBAR_ICON_GAP = 19
```

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
kaitoo1: hmm maybe we could have a memoized value thats like `const rows = useMemo()`, which would memoize the list of rows, including the placeholder. 
Then rowRenderer doesn't have to compute the rows every time, or think about the whitespace block and handle any row index specially. 

Thinking the memoized value would splice a blank block placeholder to the beginning of `displayedTokens`, so that it looks like this `['placeholder', token1, token2, etc]`
and then it would convert it into an an array of rows: `[[row1 items including placeholder], [row2items]]`

and then rowRenderer could use this list instead of `displayedTokens` ?
just an idea

jakzaizzat: That's interesting idea @kaitoo1 . Since its expensive func, it make sense to memoized it + 2d array. Will try this idea!

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
jakzaizzat: Moved it to constant file 🙌🏻

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
jakzaizzat: Updated it with memoized rows! It doesnt seems so hacky like before 😬

kaitoo1 (approved):
Thanks for addressing the comments @jakzaizzat ! LGTM

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
kaitoo1: nice!! that's great

Robinnnnn: awesome, great idea @kaitoo1!!

Robinnnnn (approved):

PR #919, opened 2022-07-30
Fix collector's note title styling

connorrothschild:
The collector's note title is now aligned with the styling of other "NFT Detail Text" (`Owner`, `Creator`, etc). It applies the styles from `TitleXS`

Before

image

After

image


Robinnnnn (approved):
🙏

PR #920, opened 2022-07-30
Add fallback fonts

connorrothschild:
This explicitly renders `serif` as a fallback font for `GT Alpina`, and `Helvetica, Arial, sans-serif` as a fallback for `ABC Diatype`.

I can't actually replicate Jess' problems with the brief flash of incorrect fonts, but for posterity here is a comparison of each font with its fallback:

### Serif (`GT Alpina`, fallback)

image
image

### Sans (`ABC Diatype`, fallback)

image
image

### DIFF
--- a/src/components/core/Text/Text.tsx
+++ b/src/components/core/Text/Text.tsx
@@ -1,8 +1,8 @@
 import styled from 'styled-components';
 import colors from '../colors';
 
-const TITLE_FONT_FAMILY = 'GT Alpina, serif';
-export const BODY_FONT_FAMILY = 'ABC Diatype, Helvetica, Arial, sans-serif';
+const TITLE_FONT_FAMILY = "'GT Alpina', serif";
+export const BODY_FONT_FAMILY = "'ABC Diatype', Helvetica, Arial, sans-serif";
 
 type TextProps = {
   color?: colors;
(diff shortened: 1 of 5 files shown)


src/components/core/Text/Text.tsx
```diff
@@ -1,8 +1,8 @@
 import styled from 'styled-components';
 import colors from '../colors';
 
-const TITLE_FONT_FAMILY = 'GT Alpina';
-export const BODY_FONT_FAMILY = 'ABC Diatype';
+const TITLE_FONT_FAMILY = "'GT Alpina', serif";
+export const BODY_FONT_FAMILY = "'ABC Diatype', Helvetica, Arial, sans-serif";
```
connorrothschild, line 5: In addition to adding fallbacks, we also wrap the primary font in quotation marks. Although functionally this shouldn't make a difference, the [spec])https://www.w3.org/TR/CSS2/fonts.html#font-family-prop) does  recommend wrapping fonts with spaces in quotation marks:

> To avoid mistakes in escaping, it is recommended to quote font family names that contain white space, digits, or punctuation characters other than hyphens:
`body { font-family: "New Century Schoolbook", serif }`

[more fun reading](https://mathiasbynens.be/notes/unquoted-font-family)

Robinnnnn: sick! great research

Robinnnnn (approved):
fire!!

PR #921, opened 2022-07-30
[GAL-321] Fix follower/following popover on mobile width

connorrothschild:
The following/follower popover had a lower width breakpoint (`useIsMobileWindowWidth`) than `FollowList` (`useIsMobileOrMobileLargeWindowWidth`). Practically, this meant that the `FollowList` applied a `width` and `height` of 100% on any screen between mobile -> tablet width, while `fullscreen` was not yet true for the modal.

This resulted in a modal that looked like this:

image

Fixed:

image


Robinnnnn (approved):
oh awesome catch!

PR #922, opened 2022-07-30
[GAL-322] Open images in new tab from NFT Detail Page

connorrothschild:
This opens images in a new tab when clicked from an NFT detail page. 
* Audio: Does not open in new tab, because I think these are usually two separate components (one for audio, one for video)
* Video: Clicking pauses/plays videos
* **GIFs: Open in new tab**
* **SVGs: Open in new tab**
* **Images: Open in new tab**

### DIFF
--- a/src/scenes/NftDetailPage/NftDetailAsset.tsx
+++ b/src/scenes/NftDetailPage/NftDetailAsset.tsx
@@ -19,14 +19,9 @@ import { StyledImageWithLoading } from 'components/LoadingAsset/ImageWithLoading
 type NftDetailAssetComponentProps = {
   tokenRef: NftDetailAssetComponentFragment$key;
   maxHeight: number;
-  inDetailPage: boolean;
 };
 
-function NftDetailAssetComponent({
-  tokenRef,
-  maxHeight,
-  inDetailPage,
-}: NftDetailAssetComponentProps) {
+function NftDetailAssetComponent({ tokenRef, maxHeight }: NftDetailAssetComponentProps) {
   const token = useFragment(
     graphql`
       fragment NftDetailAssetComponentFragment on CollectionToken {
@@ -39,6 +34,7 @@ function NftDetailAssetComponent({
             }
             ... on ImageMedia {
               __typename
+              contentRenderURL
             }
             ... on HtmlMedia {
               __typename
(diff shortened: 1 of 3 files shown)


connorrothschild:
There may be a simpler way to do this, but right now we pass the `inDetailPage` prop down to the primary image component (`ImageWithLoading). (I can't access the image url src from the `NftDetailAsset` component, so we can't pass it to `window.open()`)

src/scenes/NftDetailPage/NftDetailAsset.tsx
```diff
@@ -34,6 +34,7 @@ function NftDetailAssetComponent({ tokenRef, maxHeight }: NftDetailAssetComponen
             }
             ... on ImageMedia {
               __typename
+              contentRenderURL
```
Robinnnnn, line 37: how to get the image src field from `NftDetailAsset`!

src/scenes/NftDetailPage/NftDetailAsset.tsx
```diff
@@ -66,7 +67,15 @@ function NftDetailAssetComponent({ tokenRef, maxHeight }: NftDetailAssetComponen
...
       return <NftDetailAudio tokenRef={token.token} />;
     case 'ImageMedia':
-      return <NftDetailImage tokenRef={token.token} maxHeight={maxHeight} />;
+      return (
+        <NftDetailImage
+          tokenRef={token.token}
+          maxHeight={maxHeight}
+          // @ts-expect-error: we know contentRenderURL is present within the media field
+          // if token type is `ImageMedia`
+          onClick={() => window.open(token.token.media.contentRenderURL)}
```
Robinnnnn, line 76: this was the result of an issue we face where even after filtering for `ImageMedia` token media types, typescript doesn't recognize that `contentRenderURL` is an available field

Robinnnnn:
> There may be a simpler way to do this, but right now we pass the `inDetailPage` prop down to the primary image component (`ImageWithLoading). (I can't access the image url src from the `NftDetailAsset`component, so we can't pass it to`window.open()`)

thanks @connorrothschild! i just pushed up a commit that switches to using `onClick`, which leverages the image URL available for the nft detail page. easiest way to see the change is just by browsing the diff: https://github.com/gallery-so/gallery/pull/922/files

will go ahead and merge this!

PR #923, opened 2022-08-01
[GAL-323] Popover component when clicking on links from the feed

connorrothschild:
This PR creates two changes to support popover components to verify navigation when a user clicks on an external link

### 1. Markdown is now supported in feed

The content passed into our `StyledNote` is now passed into our `Markdown` component ([see change](
https://github.com/gallery-so/gallery/pull/923/files#diff-cd0356a1bacbd0b4c7256769d68afa4638e61c1836aa9e9d0a30d9ac50b192a8R118-R120)]. Associated with this change is a new prop called `inheritLinkStyling`, which essentially tells the `Markdown` component to skip its usual styling of links, and instead inherit styles of the parent text. In our case, we wanted to do this so that the links rendered in Markdown would inherit the same size, font family, etc. as the note (which was an H3 title).

By passing this prop into the `InteractiveLink` components, we are able to bypass the traditional link styling provided in `InteractiveLink` and maintain important styles like font size and family.

In action, markdown looks like this in feed:

image

Bolded text will unitalicize. This is thanks to [this style](https://github.com/gallery-so/gallery/blob/main/src/components/core/Text/Text.tsx#L45-L58). (We can undo this if you want)

image

We are able to adjust any appearance (e.g. remove the underline) of the rendered links thanks to `inheritLinkStyling`. (We might want to rename that prop if we do make more significant changes)

Here is how all markdown-rendered content looks: 

image

Let me know if we want to change any default appearances. Might loop in Jess for this?

### 2. Clicking on an external link will open a modal to confirm navigation intent

https://user-images.githubusercontent.com/13339581/182513456-6879f653-ec39-4b7b-92ab-3c9c52414592.mov

Behind the scenes, we pass any external link to a new component called [`InteractiveLinkNeedsVerification`](https://github.com/gallery-so/gallery/compare/connor/link-popover?expand=1#diff-77cb0c31415e5e596238569d18218d01333b8193fa2c695dfa2ffcef4016c3f9R98-R132) which essentially just passes shows a new modal—[`NavigateConfirmation`](https://github.com/gallery-so/gallery/pull/923/files#diff-702dc5388a1a3df93cef4bed5c1ec6407ad5f2fef3632dde3630cdce74e9007a)—to show `onClick`.

#### Known bug

Clicking on an external link when looking at an NFT preview from feed will close the feed before opening the modal, I assume because the preview itself is considered a fullscreen modal. [Noted here.](https://github.com/gallery-so/gallery/pull/923/files#diff-77cb0c31415e5e596238569d18218d01333b8193fa2c695dfa2ffcef4016c3f9R112-R121)

https://user-images.githubusercontent.com/13339581/182514030-59807ea4-421b-4a2b-8c56-a14c5676c07f.mov

### DIFF
--- a/src/components/Feed/Events/CollectorsNoteAddedToTokenFeedEvent.tsx
+++ b/src/components/Feed/Events/CollectorsNoteAddedToTokenFeedEvent.tsx
@@ -16,6 +16,7 @@ import unescape from 'utils/unescape';
 import { useTrack } from 'contexts/analytics/AnalyticsContext';
 import HoverCardOnUsername from 'components/HoverCard/HoverCardOnUsername';
 import { CollectorsNoteAddedToTokenFeedEventQueryFragment$key } from '__generated__/CollectorsNoteAddedToTokenFeedEventQueryFragment.graphql';
+import Markdown from 'components/core/Markdown/Markdown';
 
 type Props = {
   eventRef: CollectorsNoteAddedToTokenFeedEventFragment$key;
@@ -114,7 +115,9 @@ export default function CollectorsNoteAddedToTokenFeedEvent({ eventRef, queryRef
           </StyledMediaWrapper>
           <Spacer width={MIDDLE_GAP} />
           <StyledNoteWrapper>
-            <StyledNote>{unescape(event.newCollectorsNote ?? '')}</StyledNote>
+            <StyledNote>
+              <Markdown text={unescape(event.newCollectorsNote ?? '')} inheritLinkStyling={true} />
+            </StyledNote>
           </StyledNoteWrapper>
         </StyledContent>
       </StyledEvent>
--- a/src/components/Feed/Events/CollectorsNoteAddedToTokenFeedEvent.tsx
+++ b/src/components/Feed/Events/CollectorsNoteAddedToTokenFeedEvent.tsx
@@ -116,7 +116,7 @@ export default function CollectorsNoteAddedToTokenFeedEvent({ eventRef, queryRef
           <Spacer width={MIDDLE_GAP} />
           <StyledNoteWrapper>
             <StyledNote>
-              <Markdown text={unescape(event.newCollectorsNote ?? '')} inheritLinkStyling={true} />
+              <Markdown text={unescape(event.newCollectorsNote ?? '')} inheritLinkStyling />
             </StyledNote>
           </StyledNoteWrapper>
         </StyledContent>
--- a/src/components/core/InteractiveLink/InteractiveLink.tsx
+++ b/src/components/core/InteractiveLink/InteractiveLink.tsx
@@ -123,7 +123,7 @@ export function InteractiveLinkNeedsVerification({
       showModal({
         content: <NavigateConfirmation href={href} />,
         isFullPage: false,
-        headerText: 'Are you sure?',
+        headerText: 'Leaving gallery.so?',
       });
     },
     [showModal, track]
--- a/src/components/core/InteractiveLink/InteractiveLink.tsx
+++ b/src/components/core/InteractiveLink/InteractiveLink.tsx
@@ -16,6 +16,7 @@ type InteractiveLinkProps = {
   className?: string;
   disabled?: boolean;
   onClick?: (event?: React.MouseEvent<HTMLElement>) => void;
+  // allows the parent to override default link styles
   inheritLinkStyling?: boolean;
 };
 
@@ -36,6 +37,7 @@ export default function InteractiveLink({
 
       track('Link Click', {
         to: to || href,
+        needsVerification: false,
       });
 
       if (onClick) {
@@ -96,18 +98,23 @@ export default function InteractiveLink({
 }
 
 export function InteractiveLinkNeedsVerification({
-  inheritLinkStyling,
+  inheritLinkStyling = false,
   href,
   children,
 }: {
-  inheritLinkStyling: boolean;
+  inheritLinkStyling?: boolean;
   href: string;
   children: ReactNode;
 }) {
   const { showModal } = useModalActions();
+  const track = useTrack();
+
   const handleClick = useCallback(
     (href) => {
-      // track('What to track here?');
+      track('Link Click', {
+        to: href,
+        needsVerification: true,
+      });
 
       // FIXME: This conflicts with viewing the NFT preview page, because that is a modal.
       // So running `showModal` will close that preview modal, and then open the verify navigation modal.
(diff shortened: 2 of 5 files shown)


src/components/Feed/Events/CollectorsNoteAddedToTokenFeedEvent.tsx
```diff
@@ -114,7 +115,9 @@ export default function CollectorsNoteAddedToTokenFeedEvent({ eventRef, queryRef
           </StyledMediaWrapper>
           <Spacer width={MIDDLE_GAP} />
           <StyledNoteWrapper>
-            <StyledNote>{unescape(event.newCollectorsNote ?? '')}</StyledNote>
+            <StyledNote>
+              <Markdown text={unescape(event.newCollectorsNote ?? '')} inheritLinkStyling={true} />
```
Robinnnnn, line 119: nit: if a prop is set to `true`, you can actually just use the name of the prop and nothing else!
```js
<Markdown
  text={unescape(event.newCollectorsNote ?? '')}
  inheritLinkStyling
/>
```

src/components/core/InteractiveLink/InteractiveLink.tsx
```diff
@@ -1,19 +1,22 @@
...
-type Props = {
+type InteractiveLinkProps = {
   to?: string;
   href?: string;
   children: ReactNode;
   size?: string; // 'M', 'L', 'XL'
   className?: string;
   disabled?: boolean;
   onClick?: (event?: React.MouseEvent<HTMLElement>) => void;
+  inheritLinkStyling?: boolean;
```
Robinnnnn, line 19: let's add a comment above this line:
```js
// use this prop to allow the parent to override default link styles
```

src/components/core/InteractiveLink/InteractiveLink.tsx
```diff
@@ -81,17 +95,59 @@ export default function InteractiveLink({
...
+  children,
+}: {
+  inheritLinkStyling: boolean;
+  href: string;
+  children: ReactNode;
+}) {
+  const { showModal } = useModalActions();
+  const handleClick = useCallback(
+    (href) => {
+      // track('What to track here?');
```
Robinnnnn, line 110: great Q, let's just do this!

```js
      track('Link Click', {
        to: href,
        needsVerification: true
      });
```

and in the above tracking we can use `needsVerification: fakse`

src/components/core/InteractiveLink/InteractiveLink.tsx
```diff
@@ -81,17 +95,59 @@ export default function InteractiveLink({
...
+
+      // FIXME: This conflicts with viewing the NFT preview page, because that is a modal.
+      // So running `showModal` will close that preview modal, and then open the verify navigation modal.
```
Robinnnnn, line 113: great observation, made a ticket: https://linear.app/galleryso/issue/GAL-324/support-multiple-modals

src/components/core/InteractiveLink/InteractiveLink.tsx
```diff
@@ -81,17 +95,59 @@ export default function InteractiveLink({
...
 }
 
-export const StyledAnchor = styled.a<{ disabled?: boolean }>`
-  color: ${({ disabled }) => (disabled ? colors.porcelain : colors.shadow)};
+export function InteractiveLinkNeedsVerification({
+  inheritLinkStyling,
+  href,
+  children,
+}: {
+  inheritLinkStyling: boolean;
```
Robinnnnn, line 103: this prop is listed as required here, but optional above in the regular `InteractiveLink`. let's keep it optional and default to false?

src/components/core/InteractiveLink/NavigateConfirmation.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <ButtonContainer>
+        <StyledCancelButton onClick={hideModal}>Cancel</StyledCancelButton>
+        <StyledButton
+          onClick={() => {
+            window.open(href);
+            hideModal();
+          }}
+        >
+          Navigate
+        </StyledButton>
```
Robinnnnn, line 22: could we add a bit more left/right padding on the button? example of another one below

image

image

src/components/core/Markdown/Markdown.tsx
```diff
@@ -1,10 +1,14 @@
+// import { useCallback } from 'react';
```
Robinnnnn, line 1: toss

src/components/core/InteractiveLink/InteractiveLink.tsx
```diff
@@ -81,17 +95,59 @@ export default function InteractiveLink({
...
   }
+
+  ${({ inheritStyles }) =>
+    inheritStyles
+      ? ''
+      : css`
+          font-family: ${BODY_FONT_FAMILY};
+          font-size: 14px;
+          line-height: 18px;
+        `}
```
Robinnnnn, line 152: interesting, do we need to wrap in `css` here?

we might also be able to do something like:
```js
font-family: ${({ inheritStyles }) => ? 'inherit' : BODY_FONT_FAMILY};
font-size: ${({ inheritStyles }) => ? 'inherit' : '14px'};
line-height: ${({ inheritStyles }) => ? 'inherit' : '18px'};
```

Robinnnnn:
this is really awesome @connorrothschild!!! amazing work!

left some comments, once they're addressed i think this will be great to roll out 👍 

re: designs, great questions, but i think simply having this functionality is already a big improvement from how things look in the feed today. so i say we ship this version, then start a separate thread w/ Jess outlining the cases / screenshots you described

Robinnnnn:
ah, can we make sure to support Collection Descriptions as well? for individual collections! tbh this is more common than NFT-specific notes

image

image

connorrothschild:
@Robinnnnn all of the above are addressed! 

Collection descriptions are now supported. One note is that clicking on links in collection descriptions would open the collection, and so the modal would immediately close. Now, when a user clicks on an `InteractiveLinkNeedsVerification`, we prevent default first: https://github.com/gallery-so/gallery/pull/923/commits/b4a9a26125639daf14bfa31e5be788ca1d539974

Robinnnnn:
thanks @connorrothschild! amazing.

looks like dev is down rn as we're waiting on #927 to merge, so it's a bit hard to test. i'll test later, and once everything is looking good i'll merge this PR!

do you mind opening a new Thread with some screenshots and tagging jess for review? esp around formatting

Robinnnnn:
made a small copy update to the popover!

image

PR #927, opened 2022-08-02
Multi-Section

kaitoo1:
This PR adds multi-section functionality to the front end.
Screen Shot 2022-08-09 at 0 31 11

- [x] Can add new sections
- [x] Can move items between sections
- [x] Can set different column widths per section
- [x] Can move sections
- [ ] Can delete a section (waiting for final design)
- [x] Editor for new and existing collections
- [x] Backwards compatible with collections that were created before this feature (collections that don't have sections)
- [x] Live display setting still works

A majority of the notable changes are in [CollectionEditorContext.tsx](https://github.com/gallery-so/gallery/pull/927#diff-aab4cd1a1861fadb49e1d6efdb5d7fc6fd244dac3a79028d29156cf60b69edb5) and [StagingArea.tsx](https://github.com/gallery-so/gallery/pull/927#diff-1740f91d2fd8435c5cb08836db5f360695da8d1a86f07b70ee4dd71b9214ee92). These include changes to how the staged collection state is managed locally, and how the staging area works with 2 layers of drag and drop (sections + items).
We previously used a simple array of staged items to track the collection state, I changed it to an object that has staged Sections as values. Each Section has a list of items. You can see the type for this object, `StagedCollection`, in `OrganizeCollection/types.ts`

There's a new component, [Section.tsx](https://github.com/gallery-so/gallery/pull/927#diff-99b73d9700cb9ab41ecfb902e319339e85dae1b5d736e10cc99d1118b735923c) that renders a section of items. It is accompanied by [DroppableSection.tsx](https://github.com/gallery-so/gallery/pull/927#diff-0835297075cbc3026005d9469b8dab5ef0408e0384b4e26c7fa8f77e6cb85f42) and [SectionDragging.tsx](https://github.com/gallery-so/gallery/pull/927#diff-e64d112bd26230fc09990d2210a1486daade5b17cee66a348dc648354a226acf) which render the draggle and dragging sections, respectively.


src/components/core/Text/Text.tsx
```diff
@@ -106,3 +106,11 @@ export const TitleDiatypeL = styled(H1)`
   line-height: 24px;
   letter-spacing: -0.01em;
 `;
+
+export const TitleDiatypeM = styled(H2)`
```
kaitoo1, line 110: design included new text style

src/contexts/collectionEditor/CollectionEditorContext.tsx
```diff
@@ -1,32 +1,38 @@
...
-import { DragEndEvent } from '@dnd-kit/core';
-import { EditModeToken, StagingItem } from 'flows/shared/steps/OrganizeCollection/types';
-import { UpdateCollectionTokensInput } from '__generated__/useUpdateCollectionTokensMutation.graphql';
+import { DragEndEvent, UniqueIdentifier } from '@dnd-kit/core';
+import {
+  EditModeToken,
+  StagedCollection,
+  StagingItem,
+} from 'flows/shared/steps/OrganizeCollection/types';
+import { generate12DigitId } from 'utils/collectionLayout';
```
Robinnnnn, line 9: supernit: why not use [uuid](https://www.npmjs.com/package/uuid)?

src/flows/shared/steps/OrganizeCollection/Editor/CollectionEditor.tsx
```diff
@@ -212,10 +211,12 @@ function CollectionEditor({ viewerRef }: Props) {
     setSidebarTokens,
     stageTokens,
     unstageTokens,
-    whitespaceList,
+    collectionBeingEdited,
+    setStagedCollectionState,
   ]);
 
-  const shouldDisplayEditor = stagedNfts.length > 0;
+  useEffect(() => {}, [stagedCollectionState]);
```
Robinnnnn, line 218: ?

src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
```diff
@@ -72,63 +251,145 @@ function StagingArea({ tokensRef, stagedItems }: Props) {
...
+              />
+            ) : (
+              activeItem && (
+                <StagedItemDragging
+                  tokenRef={activeItemRef || null}
+                  isEditModeToken={isEditModeToken(activeItem)}
+                  size={
+                    IMAGE_SIZES[
+                      findSection(activeId)
+                        ? localStagedCollection[findSection(activeId)!].columns
```
Robinnnnn, line 386: nit: if you want to avoid using `findSection(activeId)!`, in the `findSection` func you can throw an error instead of returning `undefined`. that way the type signature will say it always returns a string

src/scenes/UserGalleryPage/UserGalleryCollections.tsx
```diff
@@ -82,7 +84,10 @@ function UserGalleryCollections({ galleryRef, queryRef, mobileLayout }: Props) {
       nonNullCollections.filter((collection) => {
         const isNotHidden = !collection.hidden;
         const hasTokens = collection.tokens?.length;
-        const hasWhitespace = collection.layout?.whitespace?.length;
+        const hasWhitespace =
+          collection.layout?.sectionLayout?.find((layout) => !!layout?.whitespace?.length) !==
+          undefined;
```
Robinnnnn, line 89: nit: shouldn't need to compare to undefined

src/utils/collectionLayout.test.ts
```diff
@@ -42,3 +47,84 @@ describe('insertWhitespaceBlocks', () => {
     expect(whitespacesAndNfts[2].id).toEqual(tokens[0].id);
   });
 });
+
+describe('parseCollectionLayout', () => {
```
Robinnnnn, line 51: niceee!

Robinnnnn (approved):
a tour de force. bravo ser. 👏 👏 👏

Robinnnnn:
some UI notes. none of these are blocking:

1) instead of "NFTs" should we use "pieces"? will be consistent with the copy in the sidebar, which says "Search pieces"
image

2) small nit, but the "+" icon pictured above seems very slightly off-center. i think it could be 1 pixel higher. it's more noticeable on an external monitor

3) right now i'm hovering over the first section, while the second section is Active. could we make the hover color slightly more opaque / less blue to denote the difference?

image

4) it wasn't 100% clear to me that i had to click and hold the "Section" text to drag each section around. i wonder if we could add a drag handle icon? either to the left of the word, or to the right of each section? i know it has design implications. i think we could use something like [this icon](https://cdn.iconscout.com/icon/free/png-256/drag-indicator-1779315-1512591.png), white color, against a blue rectangular background

Screen Shot 2022-08-08 at 10 22 22 PM

src/flows/shared/steps/OrganizeCollection/Editor/CollectionEditor.tsx
kaitoo1: good catch! probably had a console log there. removed

src/scenes/UserGalleryPage/UserGalleryCollections.tsx
kaitoo1: removed

src/contexts/collectionEditor/CollectionEditorContext.tsx
kaitoo1: oh yeah I think we had this convo before 😅 totally down for uuid, I just stuck with what we're using for the whitespace blocks. and since we don't save these ids to the db so I wasn't pressed to change it..  maybe we can swap it all out as a follow up or the next we visit the editor for multi chain!

src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
kaitoo1: done!

kaitoo1:
great call on ui comments!
1. updated copy
Screen Shot 2022-08-09 at 23 04 08
 
2. replaced "+" text with svg icon
Screen Shot 2022-08-09 at 23 04 12
 
3. will get design input for follow up as discussed
4. added a drag handle icon, will confirm with Design if this is ok to ship
Screen Shot 2022-08-09 at 23 06 16

kaitoo1:
improved error handling for the editor too. added a toast if updating a collection fails. (we already showed an error for creating a collection, via the description editor modal). and added explicit logging to sentry

Screen Shot 2022-08-09 at 23 00 09

PR #928, opened 2022-08-02
Ensure final row of NFTs are displayed in sidebar

Robinnnnn:
Only full rows were being pushed into the set of rows to display, and the last row was being ignored if it wasn't full.

**Before** – only 2 zorbs appearing in sidebar
image

**After** – all zorbs listed
image


jakzaizzat:
Thanks for fixing this @Robinnnnn 🙏🏻 Somehow I misslook this stuff

PR #929, opened 2022-08-03
Virtualize main gallery

jakzaizzat:
**Changes**

1. Added react virtualized to the main gallery page
2. Removed the slicing collection array + scroll event
3. Added new hooks to monitor element size - [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver)

**How it works**

We consider one collection as one row in the virtualized list

CleanShot 2022-08-04 at 16 29 32

**Notes on how the height calculation works**

1. Seems like the way react virtualized work is its calculate the sizing before rendering it in the DOM. In this case it only get the height of the `collection name` and `collection description`. 

CleanShot 2022-08-04 at 16 33 39

2. To get the exact sizing of the collection, we may need to wait for the NFT to load (especially for live render). 
3. I've added a callback function in the `UserGalleryCollection` components,  it will trigger whenever the height of the row was changed.
4. If changes, we asked the react virtualize to update the row height in their cache - [Reference](https://github.com/bvaughn/react-virtualized/blob/master/docs/CellMeasurer.md#using-cellmeasurer-with-images)

**Demo**
https://user-images.githubusercontent.com/4480258/182798811-1fa9717f-9ba7-4193-be0b-3a76205df219.mp4

### DIFF
--- a/src/hooks/useResizeObserver.ts
+++ b/src/hooks/useResizeObserver.ts
@@ -1,0 +1,26 @@
+import { useState, useEffect } from 'react';
+
+export default function useResizeObserver(ref: React.RefObject<HTMLElement>) {
+  const [dimensions, setDimensions] = useState({
+    width: 0,
+    height: 0,
+  });
+
+  useEffect(() => {
+    const observeTarget = ref.current;
+
+    if (!observeTarget) {
+      return;
+    }
+
+    const observer = new ResizeObserver((entries) => {
+      const { width, height } = entries[0].contentRect;
+      setDimensions({ width, height });
+    });
+    observer.observe(observeTarget);
+
+    return () => observer.disconnect();
+  }, [ref]);
+
+  return dimensions;
+}
--- a/src/scenes/UserGalleryPage/UserGalleryCollections.tsx
+++ b/src/scenes/UserGalleryPage/UserGalleryCollections.tsx
@@ -129,30 +129,37 @@ function UserGalleryCollections({ galleryRef, queryRef, mobileLayout }: Props) {
   return (
     <StyledUserGalleryCollections>
       <Spacer height={isMobile ? 48 : 80} />
-      <div>
-        <WindowScroller>
-          {({ height, registerChild, scrollTop, onChildScroll }) => (
-            <AutoSizer disableHeight>
-              {({ width }) => (
-                <div ref={registerChild}>
-                  <List
-                    ref={listRef}
-                    autoHeight
-                    width={width}
-                    height={height}
-                    onScroll={onChildScroll}
-                    rowHeight={cache.current.rowHeight}
-                    rowCount={numCollectionsToDisplay}
-                    scrollTop={scrollTop}
-                    deferredMeasurementCache={cache.current}
-                    rowRenderer={rowRenderer}
-                  />
-                </div>
-              )}
-            </AutoSizer>
-          )}
-        </WindowScroller>
-      </div>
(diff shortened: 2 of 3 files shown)


src/scenes/UserGalleryPage/UserGalleryCollections.tsx
```diff
@@ -115,19 +98,68 @@ function UserGalleryCollections({ galleryRef, queryRef, mobileLayout }: Props) {
...
+                  autoHeight
+                  width={width}
+                  height={height}
+                  onScroll={onChildScroll}
+                  rowHeight={cache.current.rowHeight}
+                  rowCount={numCollectionsToDisplay}
+                  scrollTop={scrollTop}
+                  deferredMeasurementCache={cache.current}
+                  rowRenderer={rowRenderer}
+                  overscanIndicesGetter={({
```
jakzaizzat, line 148: Not really sure why my `overscanRowCount` didnt work here. Replace it with snippet to render some buffer to avoid flickering.

[Similar issue](https://github.com/bvaughn/react-virtualized/issues/1582)

kaitoo1:
Good point about having to wait for NFTs to load, and nice solution to measure on load @jakzaizzat !
Looks like it works great for regular scrolling 🚀

I noticed if I use the scroll bar to quickly move up or down the page, the page contents and position shift, like things are being re-measured again. I recorded a video, it's too big to attach here so will share in the front end eng chat.

It looks like if a `UserGalleryCollection` is removed from the DOM and added again, it re-measures itself, causing the list to shift a little once it's done measuring. Since we have the `CellMeasurerCache` cache, I wonder if we can cache these measured heights so that if a specific `UserGalleryCollection` has already been rendered before, we can just use the cached value instead of calling `onLoad` again?

jakzaizzat:
Definitely notice some flickering when scrolling quickily. Thanks for the idea @kaitoo1! The `onLoad` should be only call once

Robinnnnn:
@kaitoo1 that's a genius idea 🧠 

and @jakzaizzat this is AMAZINGGGG 🔥🔥🔥🔥

src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,26 @@
+import { useState, useEffect } from 'react';
+
+export default function useResizeObserver(ref: React.RefObject<HTMLElement>) {
```
Robinnnnn, line 3: 😍

src/scenes/UserGalleryPage/UserGalleryCollections.tsx
```diff
@@ -115,19 +98,69 @@ function UserGalleryCollections({ galleryRef, queryRef, mobileLayout }: Props) {
     return <EmptyGallery message={emptyGalleryMessage} />;
   }
 
+  const rowRenderer = ({ index, key, parent, style }: ListRowProps) => {
```
Robinnnnn, line 101: nit: wrap this in `useCallback` just to be safe. i can make the commit since i have the file open!

jakzaizzat: Thanks @Robinnnnn  for the update 🙏🏻

Robinnnnn (approved):
LFGGGG

kaitoo1 (approved):

kaitoo1:
thanks for implementing the idea!! @jakzaizzat  glad it worked out! 👏

PR #934, opened 2022-08-09
[GAL-285] Fix gnosis safe auth

Robinnnnn:
- [x] Account creation
- [x] Account login

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -1,0 +1,5 @@
+{
+  "name": "Gallery",
+  "description": "Share your collection with the world.",
+  "iconPath": "https://gallery.so/favicon.ico"
+}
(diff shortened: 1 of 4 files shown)


public/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Gallery",
+  "description": "Share your collection with the world.",
+  "iconPath": "https://gallery.so/favicon.ico"
+}
```
Robinnnnn, line 5: added this so that we can be recognized as a "Safe App" by gnosis safe. there's more we can do but this at least gets us over the first step

kaitoo1: nice! great idea

kaitoo1 (approved):

PR #939, opened 2022-08-11
Asset sizing improvements

Robinnnnn:
Cleans up a lot of complex asset handling logic that was leading to regressions.

**This PR may lead to further regressions, but I believe the simplifications and improvements are a step in the right direction.** A perfect solution would require a deeper cleanup.

### Before / After

Fix stretched out images

fix stretched out images

fix stretched out images 2

fix stretched out images 3

^ Tall images look better aligned when window is tall

tall images look better aligned when window is tall

Fix cutoff on tall iframes

nft detail - fix cutoff on iframes

main gallery - fix cutoff on iframes

Fix stretched iframes

fix stretched iframes

Ensure no regression when live-rendering squiggles

ensure no regression when live-rendering squiggles

Ensure no regression when live-rendering videos

ensure no regression when live-rendering videos

Fix opengraph issue

image

### Known issue: tall live display iframe on gallery

This is because it's difficult to determine an iframe's inner contents / dimensions. Without enforcing an aspect ratio, the iframe images would appear very small – only with a height of ~150px. This would also cause a regression in square/wide iframes, so for now we're enforcing an aspect ratio of 1.

known issue - tall live display iframe on gallery

### DIFF
--- a/src/components/NftPreview/GalleryNftPreviewWrapper.tsx
+++ b/src/components/NftPreview/GalleryNftPreviewWrapper.tsx
@@ -1,6 +1,5 @@
-import ShimmerProvider, { useContentState } from 'contexts/shimmer/ShimmerContext';
+import ShimmerProvider from 'contexts/shimmer/ShimmerContext';
 import { useIsMobileWindowWidth } from 'hooks/useWindowSize';
-import { useMemo } from 'react';
 import { graphql, useFragment } from 'react-relay';
 import { GalleryNftPreviewWrapperFragment$key } from '__generated__/GalleryNftPreviewWrapperFragment.graphql';
 import NftPreview from './NftPreview';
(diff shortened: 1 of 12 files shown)


src/components/NftPreview/GalleryNftPreviewWrapper.tsx
```diff
@@ -55,34 +54,7 @@ function GalleryNftPreviewWrapper({ tokenRef, columns }: Props) {
...
-    // this could be a 1-liner but wanted to make it explicit
-    if (columns === 1) {
-      if (isMobile) {
-        return '100%';
-      }
-      if (aspectRatioType === 'wide') {
-        return '100%';
-      }
-      if (aspectRatioType === 'square' || aspectRatioType === 'tall') {
-        return '60%';
```
Robinnnnn, line 72: @jakzaizzat getting rid of the logic in this PR

jakzaizzat: Goit it

PR #940, opened 2022-08-11
Multi-section Delete follow ups 

kaitoo1:
- [x]  Set another section as active after deleting section, so that you can immediately select more items
- [x] Add keyboard shortcut to press Backspace to delete a section

### DIFF
--- a/src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
@@ -26,6 +26,7 @@ import { arrayMove, SortableContext, verticalListSortingStrategy } from '@dnd-ki
 import { FOOTER_HEIGHT } from 'flows/shared/components/WizardFooter/WizardFooter';
 
 import {
+  useActiveSectionIdState,
   useCollectionEditorActions,
   useStagedCollectionState,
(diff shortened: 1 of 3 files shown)


src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
```diff
@@ -331,6 +333,16 @@ function StagingArea({ tokensRef }: Props) {
 
   const sensors = useSensors(useSensor(MouseSensor), useSensor(TouchSensor));
 
+  const activeSectionId = useActiveSectionIdState();
+  const handleBackspacePress = useCallback(() => {
+    if (activeSectionId) {
+      const itemIds = stagedCollectionState[activeSectionId]?.items.map(({ id }) => id);
+      deleteSection(activeSectionId, itemIds);
```
Robinnnnn, line 340: what if `itemIds` is undefined? might be a bit safer to add:

`const itemIds = staged... ?? []`

Robinnnnn (approved):

src/flows/shared/steps/OrganizeCollection/Editor/StagingArea.tsx
kaitoo1: good call, done!

Robinnnnn (approved):

PR #945, opened 2022-08-13
[GAL-324] Enable multi-modal

Robinnnnn:
## Overview

Enable modal stacking by storing an array of modals in state rather than building the context around a singleton. Ended up having the nice effect of cleaning up a bunch of anti-pattern-y stuff where we were setting and un-setting config variables as the modal came and went.

**Tests**
- [x] Ensure existing usages of showModal / hideModal are unaffected
- [x] Ensure `Escape` and other hotkeys still work
- [x] Ensure hitting "Back" keeps modal open
- [x] Ensure `isModalOpenRef` prop is maintained

**Before**

https://user-images.githubusercontent.com/12162433/184471307-47da74fe-1978-4f2c-af58-c212f6145105.mov

**After**

https://user-images.githubusercontent.com/12162433/184471310-0356ebf7-9cca-4702-992e-2470a25e0076.mov

**Before**

https://user-images.githubusercontent.com/12162433/184471313-993e4a61-b21d-4d70-91fa-bb5da85642c2.mov

**After**

https://user-images.githubusercontent.com/12162433/184471356-cac6a8fb-1060-4800-8671-5c0551a5c4a6.mov

### DIFF
--- a/src/contexts/modal/AnimatedModal.tsx
+++ b/src/contexts/modal/AnimatedModal.tsx
@@ -1,21 +1,28 @@
-import { ReactElement, useCallback, useEffect, useMemo } from 'react';
+import { ReactElement, useEffect, useMemo } from 'react';
 import styled, { css, keyframes } from 'styled-components';
 import colors from 'components/core/colors';
 import transitions, {
+  ANIMATED_COMPONENT_TRANSITION_MS,
   ANIMATED_COMPONENT_TRANSLATION_PIXELS_LARGE,
 } from 'components/core/transitions';
 import breakpoints from 'components/core/breakpoints';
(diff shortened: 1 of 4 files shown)


src/contexts/modal/AnimatedModal.tsx
```diff
@@ -108,6 +104,7 @@ const _ToggleFade = styled.div<{ isActive: boolean }>`
     css`
       ${isActive ? fadeIn : fadeOut} ${transitions.cubic}
     `};
+  animation-fill-mode: forwards;
```
Robinnnnn, line 107: keeps opacity at 0 after fadeout (prevents resetting)

PR #947, opened 2022-08-15
Fix NFT spacing on safari

jakzaizzat:
**Finding**
* This bug can be replicable on Safari desktop and iOS devices
* Seems like on safari, the NFT preview use 100% height on the container altought we already inherit the sizing from the asset itself. 
* Only set the `initial` height on Safari browser.

**Before**

CleanShot 2022-08-15 at 10 38 09
CleanShot 2022-08-15 at 10 43 07

**After**

https://user-images.githubusercontent.com/4480258/184568403-4a664bc8-7059-4e10-88d0-6d9d22b8541e.mp4

https://user-images.githubusercontent.com/4480258/184568726-82909674-e6ed-45be-a0e3-71a9b4707623.mp4

### DIFF
--- a/src/components/NftPreview/NftPreview.tsx
+++ b/src/components/NftPreview/NftPreview.tsx
@@ -222,6 +222,13 @@ const StyledNftPreview = styled.div<{
   width: ${({ fullWidth }) => (fullWidth ? '100%' : 'auto')};
   height: inherit;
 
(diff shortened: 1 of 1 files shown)


Robinnnnn (approved):
amazing, thank you jakz!!

PR #950, opened 2022-08-15
Cleanup peer dep warnings

tbezman:
This is all that's left. I'll address these one by one 

- `react-albus` they need to drop a new release to support React 17 / 18
- `@web3-react/walletlink-connector` We aren't using React Native (they should mark it as optional)
- `react-spring` Same thing as above, they support a bunch of different platforms and we only use web so we can just ignore these (they should mark them as optional)
- `react-virtualized` Pretty sure this package isn't being maintained anymore. They landed a PR to support React 17 / 18 but never actually did an NPM release so....
- I'm going to stay away from the synpress stuff until I get a better grip.

image

### Other Random Stuff
- Ignored the jetbrains `.idea` folder
- Fixed our `next.config.js` since we're upgraded.
- **Nothing else had breaking changes so we should be golden.**


Robinnnnn (approved):
🔥🔥🔥🔥

PR #952, opened 2022-08-17
[GAL-205] Broken footer on mobile

jakzaizzat:
1.The broken footer only happens on large mobile devices ( > 420px )
2. Removed the `isFixed` props since we didn't used it

**Demo**

Before
CleanShot 2022-08-17 at 09 24 44

After
CleanShot 2022-08-17 at 09 25 39


src/contexts/globalLayout/GlobalFooter/GlobalFooter.tsx
```diff
@@ -80,26 +77,6 @@ const StyledGlobalFooter = styled.div<StyledFooterProps>`
 
   background-color: ${colors.white};
 
-  // TODO: all of this can go away with new layout + NFT Detail Modal update
```
Robinnnnn, line 83: 😍😍

Robinnnnn (approved):
😍

PR #953, opened 2022-08-17
[GAL-127] Render links in community page description

jakzaizzat:
If there is any url in the community description, it will automatically render it as anchor text

Before

CleanShot 2022-08-17 at 13 23 15

After

https://user-images.githubusercontent.com/4480258/185041081-c03011e3-9eee-405d-a904-05725a0f06af.mp4

CleanShot 2022-08-17 at 13 38 22

### DIFF
--- a/src/scenes/CommunityPage/CommunityPageView.tsx
+++ b/src/scenes/CommunityPage/CommunityPageView.tsx
@@ -12,6 +12,7 @@ import { useCallback, useEffect, useRef, useState } from 'react';
 import TextButton from 'components/core/Button/TextButton';
 import breakpoints from 'components/core/breakpoints';
 import TokenHolderList from 'components/TokenHolderList/TokenHolderList';
+import { VALID_URL } from 'utils/regex';
 
 type Props = {
   communityRef: CommunityPageViewFragment$key;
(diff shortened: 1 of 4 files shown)


src/scenes/CommunityPage/CommunityPageView.tsx
```diff
@@ -57,6 +58,18 @@ export default function CommunityPageView({ communityRef }: Props) {
...
 
+  function formatUrl(str: string) {
+    let matchesUrls = str.match(VALID_URL);
+    let formattedText = str;
+    matchesUrls?.map((url: string) => {
+      formattedText = formattedText.replace(url, `[${url}](${url})`);
+    });
+
+    return formattedText;
+  }
```
Robinnnnn, line 69: nit: functions defined within components should be wrapped in `useCallback`.

that said, because this is more of a utility function, can we just put this in `util/formatUrl.ts`?

src/utils/regex.test.tsx
```diff
@@ -14,4 +14,21 @@ describe('regex', () => {
     testRegex(ALPHANUMERIC_UNDERSCORES, validUsernames, true);
     testRegex(ALPHANUMERIC_UNDERSCORES, invalidUsernames, false);
   });
+
+  test('VALID_URL', () => {
```
Robinnnnn, line 18: 😍

Robinnnnn:
yeeee thank you for this!

kaitoo1 (approved):

src/scenes/CommunityPage/CommunityPageView.tsx
jakzaizzat: Sure! Just moved it to `util` folder

PR #956, opened 2022-08-18
Move connected address to top of screen

kaitoo1:
- [x] away from where it’s currently displayed under mint button
- [x] non-interactive, flat black text. truncated
- [x] confirmed no regression on desktop and mobile, doesn't affect layout or responsiveness
- [x] nothing display if wallet not connected

wallet connected (desktop)
Screen Shot 2022-08-18 at 21 46 20

wallet connected (mobile)
Screen Shot 2022-08-18 at 21 46 28

 no wallet connected
Screen Shot 2022-08-18 at 21 45 01

### DIFF
--- a/src/scenes/MerchStorePage/SingleItemPage.tsx
+++ b/src/scenes/MerchStorePage/SingleItemPage.tsx
@@ -18,6 +18,8 @@ import { useMintMerchContract } from 'hooks/useContract';
 import useMintContractWithQuantity from 'hooks/useMintContractWithQuantity';
 import { useIsMobileOrMobileLargeWindowWidth } from 'hooks/useWindowSize';
 import { ethers } from 'ethers';
+import { useAccount } from 'wagmi';
+import { truncateAddress } from 'utils/wallet';
 
 export default function ItemPage({
   label,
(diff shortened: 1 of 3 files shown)


src/scenes/MerchStorePage/SingleItemPage.tsx
```diff
@@ -43,37 +47,40 @@ export default function ItemPage({
...
-              tokenId={tokenId}
-              // We don't want to disable the button if the user is not logged in. So we don't simply check equality between these two variables, which will be undefined if the user is not logged in.
-              disabled={publicSupply - usedPublicSupply == 0}
-            />
-          </StyledPriceQuantityAndPurchaseContainer>
-        </StyledContent>
-      </StyledWrapper>
-    </StyledPage>
+    <>
+      <StyledConnectedAddress>{address}</StyledConnectedAddress>
```
kaitoo1, line 51: just added this line.
rest of change is just wrapping everything in <></>.

Ideally would have put the connected address within the StyledPage but that would require modifying the layout of StyledPage to be full page width, height etc. Didn't want to cause a conflict or regressions at this stage given the timing and small scope of this change, so I just added the address to be position: absolute and rendered above the existing page content so that it can be positioned at the top corner.

PR #959, opened 2022-08-18
Add info about shipping to single item page and change year to 2022

kaitoo1:
- [x] Add info about shipping to single item page
- [x] change 2023 to 2022

Screen Shot 2022-08-19 at 0 32 15

Screen Shot 2022-08-19 at 0 32 05


connorrothschild:
👍

PR #960, opened 2022-08-18
Merch Drop Countdown

connorrothschild:
image

image

The app constantly checks to see which time we are closest to (e.g. which time have we just passed, and what's next), and dynamically updates the countdown and renders new text accordingly. For example if we had passed allowlist minting time, and gallery members minting was up next, this would appear: 

image

Here's a video of the transition between states in action:

https://user-images.githubusercontent.com/13339581/185452252-c9090b24-c90f-4e5c-935f-9fbb4a7ae570.mov

(It technically switches 1 second late but I hope this doesn't matter lol)

---

Once we reach the public mint time, the countdown will disappear:

https://user-images.githubusercontent.com/13339581/185452544-74b6f482-c1d1-425e-83d8-d5f8285cc569.mov


connorrothschild:
For reference this is the site where I verified times: http://www.timestamp-converter.com/

src/scenes/MerchStorePage/Countdown.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <>
+      {showCounter && (
+        <StyledContainer>
+          <BaseM>{text}</BaseM>
+          <Spacer height={12} />
+          <StyledCountdown>
+            <StyledCountdownText>
+              <StyledNumber>{hours === 'NaN' ? ' ' : hours}</StyledNumber>
```
Robinnnnn, line 49: should be able to use [isNaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN)

Robinnnnn (approved):
masterpiece

PR #975, opened 2022-08-25
Backspace bugfix on editor

Robinnnnn:
**Problem:** Pressing backspace in the searchbar of the editor ended up remaining sections from the main editing area.

**Solution:** Ignore keypress events if they're coming from inputs


tbezman (approved):

kaitoo1:
thanks for fixing!!

PR #979, opened 2022-08-29
Fix modal width

Robinnnnn:
The navigation confirmation modal had a hard-coded width of 400px, making it appear cropped for screen widths <400px.

In a future PR, the width should be dictated when opening the modal, such as:
```js
showModal({
  width: '400px',
  ...
})
```
and the ModalContext should use that to set the width in non-mobile scenarios. This will avoid us having to set custom logic on the inner content.

| Before | After |
| ----- | ----- |
|  image | image |

### DIFF
--- a/src/components/core/InteractiveLink/NavigateConfirmation.tsx
+++ b/src/components/core/InteractiveLink/NavigateConfirmation.tsx
@@ -4,19 +4,16 @@ import { Button } from 'components/core/Button/Button';
 import colors from 'components/core/colors';
 import Spacer from 'components/core/Spacer/Spacer';
 import { BaseM } from '../Text/Text';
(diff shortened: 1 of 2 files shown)


tbezman:
@Robinnnnn can't we do something like `width: min(100vw, 400px)` ? or am i missing something

Robinnnnn:
@tbezman excellent idea! implemented here: [3a7fec6](https://github.com/gallery-so/gallery/pull/979/commits/3a7fec61d0536ea8cff79aa9e1e2d85758e59ac3)

 i couldn't use straight 100vw because we're setting the width for the *inner* content, which would result in this:

image

which is why we add the artificial padding. def some refactoring to do in the future

PR #983, opened 2022-08-30
[GAL-358] Fix filtering bug with NFT Sidebar Search

jakzaizzat:
**Changes**

Used the generated rows instead of calculation by amount of tokens to set virtualize rows.

**Demo**

I have **3 Moonclones NFT** in my account

**Before**
CleanShot 2022-08-30 at 17 05 28

**After**
CleanShot 2022-08-30 at 17 05 40


Robinnnnn (approved):
sweet catch!

PR #984, opened 2022-08-30
[GAL-342] [GAL-343] Add/Remove tezos wallet flow

jakzaizzat:
**Changes**

- Allow user to add tezos wallet into existing account
- Update UI for disconnect account
- Add chain icon in the list of accounts
- Used tooltip to display add/remove account

**Demo**

**Disconnect UI**
CleanShot 2022-09-01 at 15 15 14@2x

**Tooltips**
CleanShot 2022-09-01 at 15 16 12@2x

**Toastr**
CleanShot 2022-09-01 at 15 16 36@2x

**Full flow**

https://user-images.githubusercontent.com/4480258/187855123-c1b8a60b-e820-446f-97a4-6feb2e274fb3.mp4

### DIFF
--- a/src/components/WalletSelector/multichain/tezos/TezosAddWallet.tsx
+++ b/src/components/WalletSelector/multichain/tezos/TezosAddWallet.tsx
@@ -1,0 +1,279 @@
+import { Button } from 'components/core/Button/Button';
+import colors from 'components/core/colors';
+import { BaseM, TitleS } from 'components/core/Text/Text';
+import { useCallback, useEffect, useMemo, useState } from 'react';
+import styled from 'styled-components';
+import { Web3Error } from 'types/Error';
+import Spacer from 'components/core/Spacer/Spacer';
+import {
+  ADDRESS_ALREADY_CONNECTED,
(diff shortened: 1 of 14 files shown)


src/components/WalletSelector/multichain/tezos/TezosAddWallet.tsx
```diff
@@ -0,0 +1,261 @@
...
+type Props = {
+  queryRef: TezosAddWalletFragment$key;
+  reset: () => void;
+};
+
+let beaconClient: DAppClient;
+
+if (typeof window !== 'undefined') {
+  beaconClient = new DAppClient({ name: 'Gallery' });
+}
```
Robinnnnn, line 43: i wonder if it matters if this client is different from the one that's created for the regular login flow?

e.g. user logs in with tezos => first DAppClient created => user adds tezos wallet => 2nd DAppClient created

i don't think this would break anything, but ideally these are shared via React Context

Robinnnnn (approved):
wooooo incredible work! ⚡ ⚡ ⚡

Robinnnnn (approved):

PR #985, opened 2022-08-30
[GAL-363] Update Failed NFTs w/o Refreshing Page

tbezman:
### Reproduction
1. Find a token with `__typename: UnknownMedia`
2. Refresh using the UI
3. UI has updated real data but continues to show a failure

I'm pretty confident there is a bug in Relay where `useFragment`'s `FragmentResource` cache is not getting the updated data due to a changing `__typename`.

### The Gnarly Hack
`useFragment` has an internal LRU cache. This cache holds the denormalized fragment view from the normalized Relay cache. This denormalized view isn't getting updated in the cache. **So we manually blow the whole fucking thing**.
- This does not cause a new network request
- Relay simply has to rebuild the denormalized view of the cache
- This is not going to hurt us

### Other Things
- Update CI to run `graphql-codegen` as a part of unit tests
- Bump a bunch of testing deps
- Get us on the paved road for Next.js tests
  - We don't need babel anymore!!! 🎉
 - Added some re-usable GraphQL mocks for `AppProviders` to make it easy for others to write tests
 - Add a functional test to ensure this never happens again.
   - I only wrote the test for `NftDetailAsset` but this fix should work everywhere.


package.json
```diff
@@ -58,22 +57,21 @@
     "styled-components": "^5.2.3",
     "swr": "^1.1.1",
     "utf-8-validate": "^5.0.7",
-    "uuid": "^8.3.2",
+    "uuid": "9.0.0-beta.0",
```
tbezman, line 60: this was required to work with the next.js swc compiler. esm stuffs

jest.config.js
```diff
@@ -1,35 +1,41 @@
-const { pathsToModuleNameMapper } = require('ts-jest/utils');
+const { pathsToModuleNameMapper } = require('ts-jest');
```
tbezman, line 1: nothing spicy in this file

src/scenes/NftDetailPage/NftDetailAsset.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+      fireEvent.click(await findByTestId('RefreshButton'));
+    });
+
+    // Make sure we see the fallback's loading state
+    expect(await findByText('Loading...')).toBeVisible();
+
+    // See the updated image!
+    expect(await findByAltText('Test Token Name')).toBeVisible();
+  });
+});
```
Robinnnnn, line 134: 💦

package.json
Robinnnnn: lul

Robinnnnn (approved):

PR #993, opened 2022-09-08
Update to Cypress 10

jakzaizzat:
**Changes**
- Seems like synpress needed Cypress 10 to work.
- Migrated to cypress 10 - https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-version-10-0


Robinnnnn (approved):
thanks so much, jakz! 👑

PR #998, opened 2022-09-09
Enable tezos button shortcut

Robinnnnn:
- always enable shortcut
- update shortcut to use Shift + T (instead of Ctrl + T)
- update `useMultiKeyDown` to be case insensitive
- update auth copy from wallets => chains

### DIFF
--- a/src/components/core/Button/Button.tsx
+++ b/src/components/core/Button/Button.tsx
@@ -40,7 +40,7 @@ const StyledButton = styled.button<StyledButtonProps>`
   &:not(:disabled) {
     cursor: pointer;
   }
-  &[aria-disabled="true"] {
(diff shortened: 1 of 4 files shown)


kaitoo1 (approved):

src/components/core/Button/Button.tsx
```diff
@@ -106,7 +106,6 @@ const StyledButton = styled.button<StyledButtonProps>`
         background: transparent;
         color: ${colors.shadow};
         border: 1px solid transparent;
-        transition: all ${transitions.cubic};
```
Robinnnnn, line 109: this was unnecessary

jakzaizzat (approved):

tbezman (approved):

PR #999, opened 2022-09-09
Stack spacing

tbezman:
**The first two commits are me deprecating the old stuff, [this commit](https://github.com/gallery-so/gallery/pull/999/commits/8fc04905a30acaec8cc03ba722ccb40239dc0902) has all the good stuff**.

Read this doc to gain context https://www.notion.so/Why-the-Spacer-component-is-bad-297ed40731d24b80926659b1cb8fc58d


kaitoo1 (approved):
love it!! thanks for pointing out the last commit too that was super helpful

PR #1005, opened 2022-09-13
[GAL-402] Spacer migration

jakzaizzat:
**Changes**
- Migrated all the old `Spacer` component with the new `Stack` - https://github.com/gallery-so/gallery/pull/999 
- Remove `DepreciateSpacer` component
- Added `inline-flex`  & `flex-wrap` support to `Stack` component


src/components/core/Spacer/Stack.tsx
```diff
@@ -21,8 +21,9 @@ export const HStack = styled.div<{
   justify?: CSSProperties['justifyContent'];
   grow?: boolean;
   shrink?: boolean;
+  inline?: boolean;
 }>`
-  display: flex;
+  display: ${({ inline }) => (inline ? 'inline-flex' : 'flex')};
```
jakzaizzat, line 26: Added support for `inline-flex`

src/scenes/CommunityPage/CommunityPageView.tsx
```diff
@@ -62,32 +62,52 @@ export default function CommunityPageView({ communityRef }: Props) {
...
-      </StyledListWrapper>
-      <DeprecatedSpacer height={64} />
+          </StyledHeader>
+          <VStack gap={isMobile ? 32 : 64}>
+            <VStack gap={isMobile ? 32 : 80}>
+              <Spacer />
+              <MemberListFilter />
+            </VStack>
+            <Spacer />
+          </VStack>
```
tbezman, line 99: pretty sure we don't need two vstacks here. I think this is because of the Spacer misuse.

tbezman (commented):
@jakzaizzat you are my hero. thank you so much for doing this!!!!

I think the only thing left we need to crush is the removal of most, if not all of the `Spacer` uses.

If you need spacing in the beginning of an element, then that element should have padding to push its children inwards. 

I know this is really tedious work that requires insane attention to detail and I really appreciate it.

tbezman:
@jakzaizzat I also pushed up a few commits of some bugs  / changes I thought made sense

src/components/ManageWallets/ManageWallets.tsx
```diff
@@ -80,36 +80,36 @@ function ManageWallets({ newAddress, queryRef }: Props) {
...
-          chain={wallet.chainAddress.chain}
-          setErrorMessage={setErrorMessage}
-          userSigninAddress={userSigninAddress}
-          setRemovedAddress={setRemovedAddress}
-          isOnlyWalletConnected={wallets.length === 1}
-        />
-      ))}
-      <DeprecatedSpacer height={16} />
+    <VStack gap={16}>
+      <Spacer />
```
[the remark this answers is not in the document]
jakzaizzat: Thanks @tbezman for the insight! Seems like I overly used the `<Spacer/>` component

src/components/core/Spacer/Stack.tsx
```diff
@@ -21,15 +21,19 @@ export const HStack = styled.div<{
...
   gap: 0 ${({ gap }) => gap ?? 0}px;
 
   align-items: ${({ align }) => align ?? 'unset'};
   justify-content: ${({ justify }) => justify ?? 'unset'};
 
   flex-grow: ${({ grow }) => (grow ? '1' : 'unset')};
 
+  flex-wrap: ${({ wrap }) => wrap ?? 'unset'};
```
jakzaizzat, line 35: Added `flex-wrap` support too

src/scenes/CommunityPage/CommunityPageView.tsx
jakzaizzat: Updated this block!

jakzaizzat:
@tbezman Removed the unnecessary Spacer component! It was definitely misused of the component.

tbezman (approved):
LETS GOOOOOOOOOOOOOOOOOO

PR #1013, opened 2022-09-17
Paste URL onto text to create link

connorrothschild:
This PR creates the magic "paste-to-link" functionality seen on Notion, etc.

https://user-images.githubusercontent.com/13339581/190868703-53a47f5b-1c38-411a-8acc-6a1b6ab44076.mov

https://user-images.githubusercontent.com/13339581/190869107-442536e0-24bf-46e5-b5f7-fe026fda0a8a.mov

It only triggers the "paste-to-link" when the user's clipboard text begins with `http`. Otherwise, the paste behavior will be default (overwriting the selected text)

https://user-images.githubusercontent.com/13339581/190868800-ba8d2513-77c4-4d9f-b127-5e6a0e547e7b.mov

---

We wrap the pasting functionality in a `try/catch` block which means if it errors out for any reason, it will default to normal pasting behavior.

[`navigator.clipboard.readText()` is not supported in Firefox](https://stackoverflow.com/questions/67440036/navigator-clipboard-readtext-is-not-working-in-firefox), and so this functionality will not work there. Because this will fail in the `try` block, Firefox will simply default to normal pasting behavior.

### DIFF
--- a/src/components/core/TextArea/TextArea.tsx
+++ b/src/components/core/TextArea/TextArea.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import noop from 'utils/noop';
 import colors from '../colors';
 import { BaseM } from '../Text/Text';
-import MarkdownShortcuts from '../Markdown/MarkdownShortcuts';
+import MarkdownShortcuts, { setValueAndTriggerOnChange } from '../Markdown/MarkdownShortcuts';
 
 type TextAreaProps = {
   className?: string;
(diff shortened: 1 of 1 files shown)


src/components/core/TextArea/TextArea.tsx
```diff
@@ -30,6 +30,47 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
     ref
   ) => {
+    // If the user is pasting a link, automatically place it in a Markdown link
+    const handleKeyDown = useCallback(
+      async (event) => {
+        if (event.key === 'v' && event.metaKey) {
+          // Wrap in a try/catch because navigator.clipboard.readText() is not supported in Firefox. If it fails, pasting behavior will be default
+          try {
+            // FIXME: Handle Typescript error from forwardRef
+            if (ref.current) {
```
connorrothschild, line 40: Need help on resolving this `forwardRef` typescript error that I have encountered before but still do not understand 😐

src/components/core/TextArea/TextArea.tsx
```diff
@@ -30,6 +30,47 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
+
+                const textBeforeLink = textArea.value.substring(0, selectionStart);
+                const textAfterLink = textArea.value.substring(
+                  selectionStart + pastedText.length,
+                  selectionEnd + pastedText.length
+                );
+                const newValue = `${textBeforeLink}[${selectedText}](${pastedText})${textAfterLink}`;
+
+                const newSelectionStart = selectionStart + 1;
+                const newSelectionEnd = selectionStart + selectedText.length + 1;
```
connorrothschild, line 59: `selectionStart + 1` accounts for the new starting bracket `[`, and `selectionStart + selectedText.length + 1` accounts for the starting bracket `[` and `<highlighted text>`. So when the user pastes, their new "selection" (e.g. the new highlight) is `[<one character past old selection string>, <end of old selection string>]`.

src/components/core/TextArea/TextArea.tsx
Robinnnnn: cc @tbezman the typescript overlord

Robinnnnn (approved):
😍 wow, thanks for the quick turnaround ser!

src/components/core/TextArea/TextArea.tsx
tbezman: woah, this is sick.

let's leave this comment in the code as a comment

src/components/core/TextArea/TextArea.tsx
tbezman: I'll check this out Monday, unless urgent. let's wait to merge

PR #1026, opened 2022-09-27
[GAL-464] Implement Badges UI on Profile username

jakzaizzat:
**Changes**

Show art gobbler badge if the user has badge with `image URL`. 

In this case, we mock with ENS collection. 

**Demo on Profile**

https://user-images.githubusercontent.com/4480258/192439948-df90f0a9-e86c-4807-bb52-be9ffaa840ea.mp4

### DIFF
--- a/src/components/Badge/Badge.tsx
+++ b/src/components/Badge/Badge.tsx
@@ -1,0 +1,41 @@
+import InteractiveLink from 'components/core/InteractiveLink/InteractiveLink';
+import Tooltip from 'components/Tooltip/Tooltip';
+import { useCallback, useState } from 'react';
+import styled from 'styled-components';
+
+type Props = {
+  name: string;
+  imageURL: string;
+};
+
+export default function Badge({ name, imageURL }: Props) {
+  const [showTooltip, setShowTooltip] = useState(false);
+  const handleMouseEnter = useCallback(() => {
+    setShowTooltip(true);
+  }, []);
+
+  const handleMouseExit = useCallback(() => {
+    setShowTooltip(false);
+  }, []);
+
+  // TODO: replace with badge contract address
+  const communityUrl = '/community/0x31982936afe7fc432ec1a2766bdba4d07a779b30';
+
+  return (
+    <InteractiveLink to={communityUrl}>
+      <StyledTooltip text={name} showTooltip={showTooltip} />
+      <StyledBadge src={imageURL} onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseExit} />
+    </InteractiveLink>
+  );
+}
+
+const StyledTooltip = styled(Tooltip)<{ showTooltip: boolean }>`
+  opacity: ${({ showTooltip }) => (showTooltip ? 1 : 0)};
+  transform: translateY(${({ showTooltip }) => (showTooltip ? -28 : 0)}px);
+`;
+
+const StyledBadge = styled.img`
+  height: 16px;
+  width: 16px;
+  cursor: pointer;
+`;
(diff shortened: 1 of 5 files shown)


src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const handleMouseEnter = useCallback(() => {
+    setShowTooltip(true);
+  }, []);
+
+  const handleMouseExit = useCallback(() => {
+    setShowTooltip(false);
+  }, []);
+
+  // TODO: replace with badge contract address
+  const communityUrl = '/community/0x31982936afe7fc432ec1a2766bdba4d07a779b30';
```
jakzaizzat, line 22: @Robinnnnn What's the best way to get contract address art gobblers for community page?

Seems like their collection used opensea storefront - https://opensea.io/collection/art-gobbler

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <InteractiveLink to={communityUrl}>
+      <StyledTooltip text={name} showTooltip={showTooltip} />
+      <StyledBadge src={imageURL} onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseExit} />
+    </InteractiveLink>
+  );
+}
+
+const StyledTooltip = styled(Tooltip)<{ showTooltip: boolean }>`
+  opacity: ${({ showTooltip }) => (showTooltip ? 1 : 0)};
+  transform: translateY(${({ showTooltip }) => (showTooltip ? -28 : 0)}px);
```
Robinnnnn, line 34: could we make the translation a little bit less intense? maybe something like `-28 : -24`

src/components/Badge/Badge.tsx
Robinnnnn: hmm good question, so i think the badge should come with the contract address for us to link to it. this means we'll need to update the backend schema to return the contract:

image

^ this just returns an ID, but we the schema should really be `contract: Contract`. will ask someone to implement this tomorrow.

Robinnnnn: in the meantime, best practice graphql patterns is to pull the badge data as far down the react tree as possible – so `{ name, imageURL }: Props` should really be `{ badgeRef }: Props`

more expanded:
```ts
type Props = {
    // this type will be auto-generated by Relay once you hit "save".
    badgeRef: BadgeFragment$key 
}

export default function Badge({ badgeRef }: Props) {
    const badge = useFragment(
        graphql`
            fragment BadgeFragment on Badge {
                name
                imageURL
                contract {
                    address
                }
            }
        `,
        badgeRef
    )

    const { name, imageURL, contract: { address }} = badge;

    ...
}

// parent component
const user = useFragment(
    graphql`
      fragment UserGalleryHeaderFragment on GalleryUser {
        username @required(action: THROW)
        dbid
        bio
        badges {
          name
          ...BadgeFragment <-- main change
        }
      }
    `,
    userRef
  );

  ...

  {user.badges.map(badge => <Badge key={badge.name} badgeRef={badge} />)}

```

src/components/Badge/Badge.tsx
jakzaizzat: Updated!

src/components/Badge/Badge.tsx
jakzaizzat: TIL. Thanks a lot @Robinnnnn for the detail explanation!

jakzaizzat:
@Robinnnnn I've updated the graphql schema and included in the badge community url

Robinnnnn (approved):
🔥🔥🔥🔥🔥

thank you jakz! so stokeddd!

PR #1032, opened 2022-09-28
add initial percy snapshot tests for user galleries

jackiehluo:


cypress/integration/pages/page.ts
```diff
@@ -6,4 +6,8 @@ export default class Page {
     cy.switchToCypressWindow();
     cy.wait(1000);
   }
+
+  closeGlobalAnnouncement() {
+    cy.get('[data-testid="close-modal"]').click();
+  }
```
Robinnnnn, line 12: this is a great catch!

once we toss the modal, could we make this only trigger if the modal exists? i wonder if something like this is allowed:

```js
cy.get('[data-testid="close-modal"]')?.click();
```

Robinnnnn: could leave a comment like `// close modal if blocking main view`

PR #1034, opened 2022-09-29
[GAL-471] Fix gallery width

jakzaizzat:
## Problem

Seems like this issue happens only on **Single Collection page** and there is only **one NFT inside** the collection. 

## Demo gallery width comparing side by side Main gallery <-> Collection page

**Before**

https://user-images.githubusercontent.com/4480258/192917240-cf66d05d-8d57-47ea-81c0-a65afc04d9e9.mp4

**After**

https://user-images.githubusercontent.com/4480258/192917122-8a4f1506-293a-406a-877e-e8f9c1866c03.mp4

## Mobile view if there is one NFT / none

**Before**

CleanShot 2022-09-29 at 09 25 48

**After**

CleanShot 2022-09-29 at 09 26 15


Robinnnnn (approved):
sickkk thank you!!

PR #1035, opened 2022-09-29
[GAL-463] Default Editor if not connected

jakzaizzat:
**Changes**

Only applicable  to the users that didn't connect any wallet based on selected editor chain

1. Show **Connect {selectedChain} wallet** Button that will trigger the manage accounts modal
2. Sync button will be disabled
3. Can't add blank space

**Demo**

**Scenario without tezos account**

https://user-images.githubusercontent.com/4480258/192956157-9332377d-8973-433a-a603-498e28dfa61c.mp4

**Scenario without ethereum account**

https://user-images.githubusercontent.com/4480258/193236356-1443b606-3d2b-4b7f-aca1-89caa98acf51.mp4

**Scenario without ethereum account for POAP**

https://user-images.githubusercontent.com/4480258/193236860-bfc3bfb5-6726-491c-9980-44c6869de3aa.mp4

**Note**
I'll open a follow up PR for the auto refresh!

### DIFF
--- a/src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
@@ -88,7 +88,7 @@ function Sidebar({ tokensRef, sidebarTokens, queryRef }: Props) {
       (wallet) => wallet?.chainAddress?.chain === 'Tezos'
     );
 
-    return isUserHasTezosWallet;
+    return isUserHasTezosWallet ?? false;
   }, [query, selectedChain]);
 
   const editModeTokensSearchResults = useMemo(() => {
@@ -152,6 +152,7 @@ function Sidebar({ tokensRef, sidebarTokens, queryRef }: Props) {
         {!isSearching && (
           <>
             <SidebarChainSelector
+              isTezosAccountConnected={isTezosAccountConnected}
               queryRef={query}
               selected={selectedChain}
               onChange={setSelectedChain}
--- a/src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
+++ b/src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
@@ -79,8 +79,9 @@ function Sidebar({ tokensRef, sidebarTokens, queryRef }: Props) {
   const sidebarTokensAsArray = useMemo(() => convertObjectToArray(sidebarTokens), [sidebarTokens]);
 
   // Only show blank space + add account button
-  // if the user don't have tezos account
-  // & the selected chain is tezos
+  // 1. if the user don't have tezos account
+  // 2. the selected chain is tezos
+  // In future, we might enabled this on other chain too
   const isTezosAccountConnected = useMemo(() => {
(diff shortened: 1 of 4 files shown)


src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
```diff
@@ -63,6 +78,20 @@ function Sidebar({ tokensRef, sidebarTokens, queryRef }: Props) {
...
+  // In future, we might enabled this on other chain too
+  const isTezosAccountConnected = useMemo(() => {
+    if (selectedChain !== 'Tezos') return true;
+
+    const isUserHasTezosWallet = query.viewer?.user?.wallets?.some(
+      (wallet) => wallet?.chainAddress?.chain === 'Tezos'
+    );
+
+    return isUserHasTezosWallet ?? false;
+  }, [query, selectedChain]);
```
Robinnnnn, line 93: sorry for not clarifying this earlier, i also think we should enable this for ETH, because in theory someone could be connected with tezos wallet but _not_ with ETH. in this scenario we should probably produce the same behavior. 

POAP is an edge case, because it's tied to ETH.

so we could do something like this:
```js
const ownsWalletFromSelectedChain = useMemo(() => {
    let chain = selectedChain;
    if (selectedChain === 'POAP') {
        chain = 'Ethereum'
    }

    const ownsWalletFromChain = query.viewer?.user?.wallets?.some(
        (wallet) => wallet?.chainAddress?.chain === selectedChain
    );

    return ownsWalletFromChain ?? false
}, [query, selectedChain]);
```

Robinnnnn: this has some implications for the rest of the logic, which i've commented below

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
```diff
@@ -125,22 +153,30 @@ function Sidebar({ tokensRef, sidebarTokens, queryRef }: Props) {
...
+
+      {isTezosAccountConnected ? (
+        <SidebarTokens
+          isSearching={isSearching}
+          tokenRefs={nonNullTokens}
+          selectedChain={selectedChain}
+          editModeTokens={tokensToDisplay}
+        />
+      ) : (
+        <AddWalletSidebar queryRef={query} />
```
Robinnnnn, line 178: given the change above, we probably want to make this component to be more flexible to say "you do not own any {chain} NFTs".

we'll also want the CTA to say "connect {chain} wallet", but if chain is POAP it should say ethereum (or "ETH" if the button is too long)

src/flows/shared/steps/OrganizeCollection/Sidebar/Sidebar.tsx
jakzaizzat: Don't feel sorry @Robinnnnn! It make totally sense to apply on all chains. 

Thanks for the guide! I've updated it

Robinnnnn (approved):
amazingggg thank you jakz!!!

PR #1038, opened 2022-09-30
Fix Create first collection UI

jakzaizzat:
Seems like a UI regression on Spacer

**Before**

CleanShot 2022-09-30 at 16 48 28@2x

**After**

CleanShot 2022-09-30 at 17 01 31@2x


Robinnnnn (approved):
incredible catch, jakz!!

PR #1044, opened 2022-10-05
[GAL-466] Block "Next" step until username is valid

tbezman:
I cleaned up some of the logic in `useUserInfoForm` while I was at it.

The new flow is we track this new state `isUsernameValid` which starts as `false`. 
- Every time the username is changed, we reset it to `false`.
- Only when the validation passes do we finally set it to `true`.

All of this logic feels very imperative. Bringing the bottom bar (the thing with the "next" button) into the step component itself should help make this logic simpler. In the future, the "Next" button itself can `useLazyLoadQuery` with the validation state. The fallback state for that component would be a disabled version of the button. 

For now, this is fine.

### DIFF
--- a/src/components/Profile/useUserInfoForm.ts
+++ b/src/components/Profile/useUserInfoForm.ts
@@ -66,6 +66,9 @@ export default function useUserInfoForm({
 
   const usernameFieldIsDirty = useRef(false);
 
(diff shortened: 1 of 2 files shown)


Robinnnnn (approved):
🔥🔥🔥

PR #1051, opened 2022-10-12
[GAL-470] Update empty state

jakzaizzat:
**Changes**

1. Added `<EmptyState>` component
2. Used the new component on authentication wallet flow, editor, and sidebar editor

**Connect wallet**

CleanShot 2022-10-12 at 12 54 16
CleanShot 2022-10-12 at 12 10 54

**Error handler**

CleanShot 2022-10-12 at 12 54 22

CleanShot 2022-10-12 at 14 30 48


Robinnnnn (commented):
@jakzaizzat thank you for this!!

there are a few more empty states we want to handle: 1) in the editor, 2) in the editor sidebar with no wallet is connected

image

for the sidebar, you implemented a new variant where the user gets to add a tezos wallet if they don't already have one connected. maybe the Button can be an optional component within the EmptyState component that appears when you pass in certain props

Robinnnnn:
also, the button in the bottom screenshot is much longer than the top one 😅 can we reduce the width?

image

jakzaizzat:
Thanks @Robinnnnn! Reduced the width of the gnosis button. 

Correct me if Im wrong, I think for the editor and the sidebar editor empy state already been implemented before. 

**Editor**

CleanShot 2022-10-12 at 14 03 25

**Sidebar Editor**

CleanShot 2022-10-12 at 14 03 11

**Sidebar Editor if the wallet not connected**

CleanShot 2022-10-12 at 14 04 01

jakzaizzat:
@Robinnnnn Updated the PR description + add new `EmptyState` component

src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,33 @@
...
+export function EmptyState({ title, description, children }: Props) {
+  return (
+    <VStack align="center">
+      <StyledTitle>{title}</StyledTitle>
+      {description && (
+        <StyledBody>
+          <Markdown text={description} />
+        </StyledBody>
+      )}
+      {children}
```
Robinnnnn, line 21: Q: if the EmptyState simply renders the children that it's passed in, i wonder if it's even worth having it be passed in?

it seems like these would do almost the exact same thing:
```js
// current
<EmptyState>
  <A />
</EmptyState>

vs.

// without children as prop
<>
  <EmptyState />
  <A />
</>
```

so i think we can just remove `children` as a prop, since it's not doing much!

Robinnnnn (commented):
you're a legend

src/components/EmptyState/EmptyState.tsx
jakzaizzat:
```js
<VStack align="center"> <-- align center the children
  <StyledTitle>{title}</StyledTitle>
  {description && (
    <StyledBody>
      <Markdown text={description} />
    </StyledBody>
  )}
  {children}
</VStack>
```

There is a little bit different. The rendered children will be center aligned. If we didn't passed it, we might need to center manually likes


```jsx
<>
  <EmptyState />
  <VStack align='center'>
   <A />
  </Vstack>
</>
```

Robinnnnn (approved):

PR #1053, opened 2022-10-14
Interactions fixes

tbezman:
Fixing an edge case with admires / comments. See this comment
```
   * There's a bit of extra logic you can see where we're checking the `PageInfo.total`
   * as well as checking the actual length of the admires / comments array.
   *
   * This is because there is a possibility that the user will either delete a comment
   * or delete an admire. Under this case, the total count of comments / admires might
   * be some large number, but we only actually have 0, 1, or 2 in the list
   *
   * Imagine this:
   * 1. We load the page, and we hypothetically fetch 5 comments posted by the logged in user
   * 2. There's actually a total of 10 comments on the server, but we only fetched 5
   * 3. The user then goes and deletes all 5 of those comments
   *
   * Now we're in a state where `totalComments` is 5, but we don't actually have any comments
   * to show, so we'll have to fallback to some other UI here. In this case, we'll just show
   * the total number of comments / admires that links to the NotesModal
 ```

Also cleaning up some of the logic around updates with admires

Also handling removing admires

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1027,6 +1027,7 @@ type AdmireFeedEventPayload {
 
 type RemoveAdmirePayload {
     viewer: Viewer
+    admireID: DBID
     feedEvent: FeedEvent @goField(forceResolver: true)
 }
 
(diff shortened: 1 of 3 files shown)


Robinnnnn (approved):
🔥🔥🔥

PR #1055, opened 2022-10-15
[GAL-534] Fix footer overflow on small devices

jakzaizzat:
This overflow bug on happens on small devices (width < 294px).

**Before**

CleanShot 2022-10-15 at 13 37 00

**After**

CleanShot 2022-10-15 at 13 38 03


Robinnnnn (approved):
🔥🔥🔥

PR #1059, opened 2022-10-17
[GAL-519] Improve Admire Comment Jitter

tbezman:
### Changes
- Height recomputations are now done in a `useLayoutEffect` so the user sees no jitter when the height of an event changes
- Fixed pasting of stylized text into the comment box (we were seeing some weird behavior before)

The layout shift you're seeing at the end of the video is from the server stripping a trailing newline (I'm 99% sure)

https://user-images.githubusercontent.com/6754223/196294988-505f15d2-3d4a-4cfb-841f-66b8300907c7.mov

### THE REMAINING ISSUE
There is one more incredibly shitty behavior I'm seeing when scrolling through the list (independent of admire / comment). There seems to be this issue where when you scroll fast enough you can get into this fucked state where scrolling up one pixel shifts you way far, and then scrolling down one pixel scrolls you way far. I'm not sure what the root cause is but it seems deeply rooted in react-virtualized and I'm hoping this is just fixed when we migrate to `react-window`

### DIFF
--- a/src/components/Feed/FeedList.tsx
+++ b/src/components/Feed/FeedList.tsx
@@ -117,8 +117,10 @@ export default function FeedList({
                 // Whenever the height changes, we need to ask react-virtualized
                 // to re-evaluate the height of the item to keep the virtualization good.
                 onPotentialLayoutShift={() => {
-                  measurerCache.clear(index, 0);
-                  virtualizedListRef.current?.recomputeRowHeights();
+                  setTimeout(() => {
+                    measurerCache.clear(index, 0);
+                    virtualizedListRef.current?.recomputeRowHeights(index);
+                  }, 0);
                 }}
                 eventRef={content}
                 key={content.dbid}
--- a/src/components/Feed/FeedList.tsx
+++ b/src/components/Feed/FeedList.tsx
@@ -61,6 +61,7 @@ export default function FeedList({
 
   const measurerCache = useMemo(() => {
     return new CellMeasurerCache({
+      defaultHeight: 400,
       fixedWidth: true,
       minHeight: 0,
     });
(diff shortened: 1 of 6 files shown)


Robinnnnn (approved):

src/components/Feed/FeedList.tsx
```diff
@@ -61,6 +61,7 @@ export default function FeedList({
 
   const measurerCache = useMemo(() => {
     return new CellMeasurerCache({
+      defaultHeight: 400,
       fixedWidth: true,
       minHeight: 0,
     });
```
Robinnnnn, line 67: any reason not to just define this outside the component?

src/components/Feed/Socialize/CommentBox.tsx
```diff
@@ -254,14 +249,33 @@ export function CommentBox({ active, onClose, eventRef, queryRef, onPotentialLay
     }
   }, []);
 
+  // This prevents someone from pasting a styled element into the textbox
+  const handlePaste = useCallback<ClipboardEventHandler<HTMLParagraphElement>>(
+    (event) => {
+      event.preventDefault();
+      const text = event.clipboardData.getData('text/plain');
+
+      // This is deprecated but will work forever because browsers
```
Robinnnnn, line 258: 😂

src/components/Feed/Socialize/Interactions.tsx
```diff
@@ -96,6 +97,17 @@ export function Interactions({ eventRef, queryRef }: Props) {
...
+  const isFirstMount = useRef(true);
+  useLayoutEffect(() => {
+    if (!isFirstMount.current) {
+      onPotentialLayoutShift();
+    }
+
+    isFirstMount.current = false;
+
+    // These are all the things that might cause the layout to shift
+  }, [onPotentialLayoutShift, nonNullAdmires, nonNullComments, totalComments, totalAdmires]);
```
Robinnnnn, line 109: 🔥

PR #1069, opened 2022-10-20
[GAL-549] Fix profile jank on iOS

jakzaizzat:
**FInding**
Seems like the issue happens because the virtualize recompute the height when the window size changed (in this case the height changed)

Solution: We only monitor the `width` of the window, if its changed we will recomputed the height. To cover if the user rotate their phone or resize browser. 

**Demo**

**Before**

https://user-images.githubusercontent.com/4480258/196842375-7f2eb78f-fdf7-4ee8-9f9a-4763a9fc302f.mp4

**After**

https://user-images.githubusercontent.com/4480258/196842430-b479193e-9508-45ad-bdf6-81fa323dd760.mp4


Robinnnnn (approved):
👏

PR #1071, opened 2022-10-21
refactor: use routes constant for all internal paths

evgenyboxer:
Hopefully I've covered all instances of the internal paths! I've replaced them all with the routes constant.


src/scenes/LandingPage/LandingPage.tsx
```diff
@@ -36,11 +39,11 @@ export default function LandingPage() {
       </VStack>
       <StyledBottomContainer gap={12}>
         <HStack gap={8}>
-          <NavLink to="/members">Members</NavLink>
+          <NavLink to={ROUTES.MEMBERS}>Members</NavLink>
           <BaseS>·</BaseS>
-          <NavLink to="/qaulv">Gallery of the Week</NavLink>
+          <NavLink to={ROUTES.USER.ROOT(GALLERY_OF_THE_WEEK_USER)}>Gallery of the Week</NavLink>
```
evgenyboxer, line 44: I assumed that this is a constant we change every week right?

pages/[username]/[collectionId]/[tokenId]/index.tsx
```diff
@@ -22,7 +23,7 @@ export default function NftDetailPage({ username, collectionId, tokenId }: NftDe
 
   // the default "back" behavior from the NFT Detail Page
   // is a redirect to the Collection Page
-  const collectionRoute = `/${username}/${collectionId}`;
+  const collectionRoute = ROUTES.USER.COLLECTION(username, collectionId);
```
Comment (Robinnnnn) [inline pages/[username]/[collectionId]/[tokenId]/index.tsx:26]: curious where this pattern is from? while i can see how this offers safer guarantees, i think the original format makes the route super clear! as opposed to having a level of indirection

src/constants/routes.ts
```diff
@@ -0,0 +1,15 @@
...
+    ROOT: (username: string) => `/${username}`,
+    COLLECTION: (username: string, collectionId: string) => `/${username}/${collectionId}`,
+    COLLECTION_TOKEN: (username: string, collectionId: string, tokenId: string) =>
+      `/${username}/${collectionId}/${tokenId}`,
+  },
+  MEMBERS: '/members',
+  SHOP: '/shop',
+  PRIVACY: '/privacy',
+  TERMS: '/terms',
+} as const;
```
Robinnnnn, line 15: this is so slick 😍

src/scenes/LandingPage/LandingPage.tsx
```diff
@@ -8,6 +8,9 @@ import LogoBracketLeft from 'icons/LogoBracketLeft';
 import LogoBracketRight from 'icons/LogoBracketRight';
 import colors from 'components/core/colors';
 import { HStack, VStack } from 'components/core/Spacer/Stack';
+import { ROUTES } from 'constants/routes';
+
+const GALLERY_OF_THE_WEEK_USER = 'qaulv';
```
Robinnnnn, line 13: this should be set on the server, will make a ticket

pages/[username]/[collectionId]/[tokenId]/index.tsx
Comment (evgenyboxer) [reply pages/[username]/[collectionId]/[tokenId]/index.tsx:26]: Good feedback! For me personally this actually reads fine and provides a somewhat scalable way to make the routes. But, yeah this could get pretty long and perhaps not as clear.

Another option is to use a function that would sit in the routes constant file:
```
export function userCollectionRoute(username: string, collectionId: string) {
  return `${ROUTES.USER.ROOT(username)}/${collectionId}`;
}
```

It's shorter, and still has the same guarantees.

Robinnnnn (approved):
based

tbezman:
@evgenyboxer Have you looked into this? https://github.com/ckastbjerg/next-type-safe-routes

Seems like it could get us most of the benefits without any of the maintenance. 

Also would get us type safety on params and stuff.

Definitely curious to hear your thoughts. Let me know!

evgenyboxer:
Hey @tbezman 

That's absolutely the way to go, had no idea this existed. My only concern with this specific library is that it doesn't look widely used / maintained.

I had a spin on it, and found a couple of issues - 

1. It seems that they require an `api` folder to be present inside the `pages` (took me a little while to debug to reach this conclusion) and this is reproduced in their example repo as well. Workaround is to create an empty `api` folder and then it would generate the types. However, it generates them with `TypeSafeApiRoute` which is essentially assigned to `any` and that kinda breaks the type safety (because the folder is empty). I find it odd that they require an `api` folder... but anyways, if they would at least not generate `TypeSafeApiRoute` it would work well in this case.

2. When using with Next.js there are some warnings coming out of the next.js config, but still it generates the types fine.

3. You can't change the location of where the types are generated, seems there is a PR opened from March but wasn't merged - https://github.com/ckastbjerg/next-type-safe-routes/pull/35

Happy for you to play with this as well and let me know if you find success in using it.

I really like the direction of this library, it is definitely an improvement over my solution.

next.config.js
```diff
@@ -1,7 +1,9 @@
 const relayConfig = require('./relay.config');
 const { withSentryConfig } = require('@sentry/nextjs');
+const withRoutes = require('nextjs-routes/config')();
 
-const moduleExports = {
+/** @type {import('next').NextConfig} */
```
evgenyboxer, line 5: This gives a nice autocomplete for next config options

next.config.js
```diff
@@ -61,4 +63,6 @@ const sentryWebpackPluginOptions = {
   // https://github.com/getsentry/sentry-webpack-plugin#options.
 };
 
-module.exports = withSentryConfig(moduleExports, sentryWebpackPluginOptions);
+const plugins = [withRoutes, (config) => withSentryConfig(config, sentryWebpackPluginOptions)];
+
+module.exports = () => plugins.reduce((config, plugin) => plugin(config), nextConfig);
```
evgenyboxer, line 68: Followed the thread here - https://github.com/cyrilwanner/next-compose-plugins/issues/59

pages/_/components.tsx
```diff
@@ -55,28 +55,28 @@ export default function DesignPage() {
       <Section>
         <TitleM>ButtonLink</TitleM>
         <Examples>
-          <ButtonLink href="#">primary</ButtonLink>
-          <ButtonLink href="#" pending>
+          <ButtonLink href={{ pathname: '/' }}>primary</ButtonLink>
+          <ButtonLink href={{ pathname: '/' }} pending>
```
evgenyboxer, line 59: @tbezman Would that work? or do we need a workaround here?

src/types/Elements.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type AnchorElementProps = AnchorHTMLAttributes<HTMLAnchorElement>;
+
+export type ExternalAnchorElementProps = AnchorElementProps & {
+  href: string;
+};
+
+export type InternalAnchorElementProps = Omit<AnchorElementProps, 'href'> & {
+  href: Route;
+};
```
evgenyboxer, line 12: @tbezman I wasn't sure of a good name for this, so went with something generic (`Elements.ts`) that could house similar types in the future.

next.config.js
evgenyboxer, line 5: This gives us a new autocomplete for the config

src/components/Feed/Socialize/AdmireLine.tsx
```diff
@@ -56,10 +57,14 @@ export function AdmireLine({ admireRef, queryRef }: CommentLineProps) {
     }
   }, [admire.admirer?.dbid, admire.admirer?.username, query.viewer?.user?.dbid]);
 
-  const admirerLink = admire.admirer?.username ? `/${admire.admirer.username}` : '';
+  const admirerLinkRoute: Route = admire.admirer?.username
+    ? { pathname: '/[username]', query: { username: admire.admirer.username } }
+    : { pathname: '/' };
+
+  const admirerLink = route(admirerLinkRoute);
```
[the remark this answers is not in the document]
tbezman: Makes sense to me. Next 13 removes this double tag nonsense so we'll be able to just use the route!

next.config.js
tbezman: YESSSSSS

pages/_/components.tsx
tbezman: No problem here

PR #1079, opened 2022-10-25
Vercel analytics

Robinnnnn:
[awaiting upgrade to react 18]

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "@taquito/utils": "^13.0.1",
     "@tezos-domains/taquito-client": "^1.19.0",
     "@types/relay-test-utils": "^6.0.5",
+    "@vercel/analytics": "^0.1.1",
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "preserve",
+    "jsx": "react-jsx",
```
evgenyboxer, line 17: Nice, was just reading on this. So this works because we are on React 17, and this gives a bit of perf boost right?

Robinnnnn: so i did this to try and address a build process bug that is associated with installing `@vercel/analytics`

image

unfortunately, vercel decided to revert this setting on my behalf! 😠 

image

supposedly an easy fix is to bump to React 18: https://github.com/hashicorp/next-mdx-remote/pull/250#issuecomment-1082395589

PR #1080, opened 2022-10-26
Upgrade to React 18

tbezman:
Actually turned out not being so crazy.

- Looks like we had a bunch of fucked types in `useCallback` that are now fixed.
- Had to force a resolution on react types since third party libs were bringing in react 17 types.


package.json
```diff
@@ -13,6 +13,9 @@
     "cypress:gh": "synpress run --record --key 9d633f22-7630-4a11-9cf4-ff9809332564 --configFile cypress.config.ts --group Desktop -b chromium",
     "find-unused-modules": "ts-prune -s .graphql.ts -i pages/"
   },
+  "resolutions": {
+    "@types/react": "^18.0.0"
+  },
```
tbezman, line 18: Need to make sure any third party libs aren't bringing us down to the React 17 types.

pages/_app.tsx
```diff
@@ -25,9 +25,15 @@ export type MetaTagProps = {
...
-);
+function SafeHydrate({ children }: PropsWithChildren) {
+  const [render, setRender] = useState(false);
+
+  useEffect(() => {
+    setRender(true);
+  }, []);
+
+  return <div suppressHydrationWarning>{render ? children : null}</div>;
+}
```
tbezman, line 36: Achieves the same thing but this is the React approved way of doing it

package.json
```diff
@@ -49,8 +51,8 @@
     "next": "^12.3.0",
     "nextjs-routes": "^0.1.5",
     "qr-code-styling": "^1.6.0-rc.1",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "18",
+    "react-dom": "18",
```
Robinnnnn, line 55: should we lock down a more specific version?

pages/_app.tsx
Robinnnnn: can we add a comment explaining this

src/contexts/boundary/ErrorBoundary.tsx
```diff
@@ -1,12 +1,12 @@
-import { Component } from 'react';
+import { Component, PropsWithChildren } from 'react';
```
Robinnnnn, line 1: `PropsWithChildren` is clutchhh

Robinnnnn (approved):
let's gooooo!

package.json
Robinnnnn: fantastic catch

jakzaizzat: 🧠

jakzaizzat (approved):
LFG!

kaitoo1 (approved):

PR #1088, opened 2022-10-31
[GAL-576] HTML Encoding Chars on Token Name

tbezman:
image

We don't want that ^^^


src/components/NftPreview/NftPreviewLabel.tsx
```diff
@@ -32,6 +34,14 @@ function NftPreviewLabel({ className, tokenRef, interactive = true }: Props) {
 
   const showCollectionName = Boolean(token.name);
 
+  const decodedTokenName = useMemo(() => {
+    if (token.name) {
+      return htmlDecode(token.name);
```
Robinnnnn, line 39: we have something under `utils/unescape` – how does this compare?

Robinnnnn: https://github.com/gallery-so/gallery/blob/main/src/utils/unescape.ts

Robinnnnn (approved):

PR #1094, opened 2022-11-02
[GAL-596] Fix newlines markdown

jakzaizzat:
This bug happens on two different places

1. Bio
1. Collection description

**Before**
CleanShot 2022-11-02 at 12 12 16

**After**
CleanShot 2022-11-02 at 12 12 01


Robinnnnn (approved):
thank you!

PR #1096, opened 2022-11-03
Remove padding in between paragraph blocks for collection description

Robinnnnn:
**Reported Bug**
image

**Before**
- no padding in between paragraphs
- weird expanding behavior on-click

https://user-images.githubusercontent.com/12162433/199828898-02f0a0ba-a8de-4326-b38c-46c2d2d9d4d5.mov

**After**
- ensure padding
- no padding change on click
image

### DIFF
--- a/src/scenes/UserGalleryPage/UserGalleryCollection.tsx
+++ b/src/scenes/UserGalleryPage/UserGalleryCollection.tsx
@@ -256,15 +256,19 @@ const StyledCollectorsNote = styled(BaseM)<{ showMore: boolean }>`
           -webkit-line-clamp: unset;
         `
       : css`
-          // We only care about line clamping on mobile
-          @media only screen and ${breakpoints.tablet} {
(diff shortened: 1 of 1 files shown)


src/scenes/UserGalleryPage/UserGalleryCollection.tsx
```diff
@@ -262,10 +262,6 @@ const StyledCollectorsNote = styled(BaseM)<{ showMore: boolean }>`
           }
 
           -webkit-line-clamp: 2;
-
```
tbezman, line 265: @Robinnnnn we only want to remove this if we're tablet size and greater. logic is a bit hard to follow but basically the idea is if we're doing line clamping, we don't want the padding, and on > tablet, we don't do the line clamp

Robinnnnn: thanks @tbezman, fixed! tested on both desktop + mobile

PR #1102, opened 2022-11-07
GAL-516 add "add email" component and step

kaitoo1:
This PR adds the "Add Email" component, and adds it as a step to the onboarding flow. This step replaces the "Congratulations" last step of the existing onboarding flow. The White Rhino feature flag controls which step is shown.

Screen Shot 2022-11-09 at 23 56 10

The default state before an email address has been added is to show an input.
Screen Shot 2022-11-09 at 23 22 03

Once an email is entered, it will be in an unverified state, required the user to verify via an email they receive.
Screen Shot 2022-11-10 at 0 05 16

The EmailVerificationStatus component reflects the various verification states:
verified
Screen Shot 2022-11-10 at 0 05 20

error
Screen Shot 2022-11-10 at 0 05 26

- [x] Add "Add email" input
- [x] Add simple email address format validation to input
- [x] Add verification state for unverified, verified, error states
- [x] Enable editing a previously added email address
- [x] Add these components to a new step in onboarding
- [x] Put step behind feature flag

### DIFF
--- a/pages/onboarding/add-email.tsx
+++ b/pages/onboarding/add-email.tsx
@@ -4,13 +4,12 @@ import { useLazyLoadQuery } from 'react-relay';
 import { graphql } from 'relay-runtime';
 import styled from 'styled-components';
 
-import { ButtonLink } from '~/components/core/Button/Button';
 import { VStack } from '~/components/core/Spacer/Stack';
-import { BaseM, BaseXL, TitleDiatypeL, TitleL } from '~/components/core/Text/Text';
-import EmailForm from '~/components/Email/EmailForm';
+import { BaseM, TitleDiatypeL, TitleL } from '~/components/core/Text/Text';
 import EmailManager from '~/components/Email/EmailManager';
 import FullPageCenteredStep from '~/components/Onboarding/FullPageCenteredStep';
 import { OnboardingFooter } from '~/components/Onboarding/OnboardingFooter';
+import { addEmailQuery } from '~/generated/addEmailQuery.graphql';
 
 export default function AddEmail() {
   const query = useLazyLoadQuery<addEmailQuery>(
@@ -22,6 +21,9 @@ export default function AddEmail() {
 
             user {
               username
+            }
+            email {
+              email
             }
           }
         }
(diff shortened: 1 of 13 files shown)


pages/onboarding/add-email.tsx
```diff
@@ -0,0 +1,82 @@
...
+    );
+  }
+
+  const { push, back } = useRouter();
+
+  const username = query?.viewer?.user?.username;
+  const savedEmail = query?.viewer?.email?.email;
+
+  const handleNext = useCallback(() => {
+    // track('');
```
kaitoo1, line 48: want to track an event here but didn't think of what yet

src/components/Email/EmailVerificationStatus.tsx
```diff
@@ -0,0 +1,149 @@
...
+          </>
+        );
+      case 'Verified':
+        return (
+          <>
+            <CircleCheckIcon />
+            <BaseS>Verified</BaseS>
+          </>
+        );
+      case 'Failed':
```
tbezman, line 115: What does it mean to be in a `Failed` state? Like how do I fail an email verification? 

Also, what does the `Admin` state mean?

src/components/Email/EmailForm.tsx
```diff
@@ -0,0 +1,172 @@
...
+            }
+          }
+        }
+      }
+    `,
+    queryRef
+  );
+
+  const [updateEmail] = usePromisifiedMutation<EmailFormMutation>(graphql`
+    mutation EmailFormMutation($input: UpdateEmailInput!) @raw_response_type {
```
tbezman, line 41: nit: don't need the `@raw_response_type` since you're not doing an optimistic update

tbezman:
@kaitoo1 good shit

1. should we poll the verification status so the user sees it update when they take action on the email? Does the link in the email take them to a settings page or something where they can see that they're verified? How should I be thinking about this flow?
2. Are we adding a form in a follow up PR so existing users can add their email?

src/components/Email/EmailVerificationStatus.tsx
kaitoo1: `Failed` captures edge cases like if the email couldn't be delivered to the address. Something went wrong with the verification flow. 
We can improve how we handle it in the UI  by adding more nuanced remedy steps for users who end up in this state, but to start this at leasts let us flag users who encountered an issue somewhere in the flow.

this is what Benny said for what`Admin` does:  "Basically this would be a version of Verified that is used in place of Verified on dev and local so we can shoot emails to our admin emails."

src/components/Email/EmailForm.tsx
```diff
@@ -0,0 +1,172 @@
...
+    },
+    [handleSaveClick]
+  );
+
+  return (
+    <form onSubmit={handleFormSubmit}>
+      <VStack gap={8}>
+        <StyledInput
+          onChange={handleEmailChange}
+          placeholder="Enter your email address..."
```
Robinnnnn, line 140: can we remove ellipses

Robinnnnn: alternative: Email address

Robinnnnn (approved):

PR #1106, opened 2022-11-08
Preloading and cors

tbezman:
## Removing CORS preflights

It seems like we make one preflight request per page visit since they are cached. This can actually take ~300ms 
image

Putting a reverse proxy in front of the API on the `gallery.so` domain puts us in a position where we don't have to make CORS requests.

### ⚠️ Slow Reverse Proxy ⚠️
I'm seeing that requests through the reverse proxy are taking quite a bit longer to fulfill. Both environments pointing to prod

**This could be due to Vercel making preview environments not as juicy as production environments. We can test this and revert in production if it seems to be hurting more than it's helping.**

### Direct API hit
image

### Reverse Proxy hit
image

### Old
image

### New
image

The idea is that requests will finish as fast as possible without having to wait on dependent network requests.

### Before this PR
image
1. AuthContext blocks GlobalLayoutContext
2. GlobalLayoutContext blocks the page load

### After this PR
image
1. Previously blocking requests are still blocking to render, but the requests start way earlier
2. Time to render `page` used to be `(authcontext + globallayoutcontext + page)`, now it's `max(authcontext, globallayoutcontext, page)`


next.config.js
```diff
@@ -34,6 +34,15 @@ const nextConfig = {
...
   },
 
+  async rewrites() {
+    return [
+      {
+        source: '/glry/:path*',
+        destination: `${process.env.NEXT_PUBLIC_API_BASE_URL}/glry/:path*`,
+      },
+    ];
+  },
```
tbezman, line 44: This basically says if we hit `gallery.so/glry/graphql/query`, Vercel will pipe the request to `api.gallery.so/glry/graphql/query`

pages/_app.tsx
```diff
@@ -85,7 +102,7 @@ const App: FC<{
         )}
       </Head>
       <SafeHydrate>
-        <AppProvider relayCache={relayCache}>
+        <AppProvider relayEnvironment={relayEnvironment}>
```
tbezman, line 105: we have to pass in a relay environment since we make the requests earlier than the `AppProvider` is created.

src/components/Feed/Feed.tsx
```diff
@@ -28,10 +28,15 @@ export default function Feed({ queryRef, feedMode, setFeedMode }: Props) {
...
+  const firstMountRef = useRef(true);
   // refetch when changing modes. this not only displays up-to-date data, but fixes a bug
   // for users logging in from the feed view who were seeing empty ViewerFeeds
   useEffect(() => {
-    refetch({}, { fetchPolicy: 'store-and-network' });
+    if (!firstMountRef.current) {
+      refetch({}, { fetchPolicy: 'store-and-network' });
+    }
+
+    firstMountRef.current = false;
```
tbezman, line 39: we don't need to be refreshing as soon as the component mounts.

src/components/NotificationsModal/notifications/SomeoneViewedYourGallery.test.tsx
```diff
@@ -88,8 +89,9 @@ async function assertSituation(args: MockResponseArgs, expectedText: string) {
 
   mockGraphqlQuery('SomeoneViewedYourGalleryTestQuery', response);
 
+  const relayEnvironment = createEmptyRelayEnvironment();
```
tbezman, line 92: nice little util for tests

src/contexts/relay/relayFetchFunction.ts
```diff
@@ -1,11 +1,12 @@
...
-import { FetchFunction, GraphQLResponse, RequestParameters } from 'relay-runtime';
+import { FetchFunction, GraphQLResponse, PayloadError, RequestParameters } from 'relay-runtime';
 
-import { _fetch, baseurl } from '~/contexts/swr/fetch';
+import { _fetch } from '~/contexts/swr/fetch';
+import { baseUrl } from '~/utils/baseUrl';
 
 export function getGraphqlHost() {
-  return baseurl;
+  return baseUrl;
```
tbezman, line 9: subtle change, but this one is the frontend url, so like `gallery.so`

src/contexts/relay/relayFetchFunction.ts
```diff
@@ -26,22 +27,31 @@ export const relayFetchFunction: FetchFunction = async (request, variables) => {
...
+        ? { message: error.message, severity: 'CRITICAL' }
+        : { message: 'An unexpected error occurred in relayFetchFunction', severity: 'CRITICAL' };
 
-  return response;
+    return { errors: [payloadError] };
+  } finally {
+    // ---------- begin post-request hooks
+    teardownSentryTracing(transaction);
+    // ---------- end post-request hooks
+  }
```
tbezman, line 54: all of this code wasn't handling errors very well so i made it a bit more resilient

src/types/PageComponentPreloadQuery.ts
```diff
@@ -0,0 +1,9 @@
+import { ParsedUrlQuery } from 'querystring';
+import { Environment } from 'react-relay';
+
+export type PreloadQueryArgs = {
+  query: ParsedUrlQuery;
+  relayEnvironment: Environment;
+};
+
+export type PreloadQueryFn = (args: PreloadQueryArgs) => void;
```
tbezman, line 9: types for stuff related to preloading components

src/types/memo-preload.d.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Override the types on `ExoticComponent` so the following works
+ *
+ * const SomeComponent = memo(()=> {});
+ *
+ * SomeComponent.preloadQuery = () => {};
+ */
```
tbezman, line 7: this

src/utils/baseUrl.ts
```diff
@@ -10,13 +10,7 @@ const getBaseUrl = () => {
...
 
   // Otherwise, we're probably running locally?
-  if (process.env.NODE_ENV === 'production') {
-    return 'https://gallery.so';
-  }
-
-  // TODO: figure out if/what hostname needed for tests
-
-  return `http://localhost:3000`;
+  return 'http://localhost:3000';
```
tbezman, line 13: This assumes that if there's no vercel env var, then we're most likely on localhost, instead of assuming we're on the production url

Robinnnnn (approved):

PR #1108, opened 2022-11-09
[GAL-590] Email popover

jakzaizzat:
**Changes**

- Add new variant to our toastr component - `error`
- new Toggle component
- Add `headerAction` props to our modal component
- Check JWT token on email verification (logged in and non-logged in user)
- Email banner on notification list

**Demo**

**Email Banner + Cog icon**

CleanShot 2022-11-11 at 20 44 58@2x

**New setting modal**

CleanShot 2022-11-11 at 20 45 15@2x

**Verify Email**

https://user-images.githubusercontent.com/4480258/201608155-e5fc73fd-ce0f-49b2-8ecc-5e8dfe3d5df1.mp4


src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+    query?.viewer?.email?.emailNotificationSettings?.unsubscribedFromNotifications || false;
+
+  const isEmailUnsubscribedFromAll =
+    !!query?.viewer?.email?.emailNotificationSettings?.unsubscribedFromAll ?? false;
+
+  // Set the initial state of the email notification toggle
+  // so we can instantly toggle it on/off without waiting for the mutation to complete
+  useEffect(() => {
+    setIsEmailNotificationChecked(isEmailNotificationSubscribed);
+  }, [isEmailNotificationSubscribed]);
```
tbezman, line 70: `setState` in a `useEffect` is generally not my favorite thing, since it can accidentally cause infinite loops.

I think an easier pattern here would be to just initialize the `useState` with the `isEmailNotificationSubscribed` value.

Something like
```ts
  const [isEmailNotificationChecked, setIsEmailNotificationChecked] = useState(isEmailNotificationSubscribed);
```

 If we do that, we can just remove this useEffect

src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  // If the user already have email attached, toggle the email manager ui
+  const userEmail = query?.viewer?.email?.email;
+  useEffect(() => {
+    setIsShowAddEmail(Boolean(userEmail));
+  }, [userEmail]);
```
tbezman, line 76: same thing here

tbezman: I want to call out that the tradeoff here is that if the Relay store updates, our local state won't update anymore. But I feel strongly that you wouldn't want some change in the Relay store to cause "random" changes to happen to a form you're submitting.

tbezman: if that is what we end up wanting, that's fine, but if we initialize `useState` appropriately, at least we won't have a double render.

src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+      setIsPending(true);
+      try {
+        const response = await updateEmailNotificationSettings(
+          unsubscribedFromNotifications,
+          isEmailUnsubscribedFromAll
+        );
+
+        // If its failed, revert the toggle state
+        if (!response?.updateEmailNotificationSettings) {
+          // setIsEmailNotificationChecked(!checked);
```
tbezman, line 93: left over comment?

src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+    setIsShowAddEmail(Boolean(userEmail));
+  }, [userEmail]);
+
+  const [isPending, setIsPending] = useState(false);
+
+  const handleEmailNotificationChange = useCallback(
+    async (checked: boolean) => {
+      const unsubscribedFromNotifications = !checked;
+      setIsEmailNotificationChecked(checked);
+      setIsPending(true);
```
tbezman, line 84: `usePromisifiedMutation` returns `[submit, boolean]`. can we thread that through here so we don't have to manage that state ourselves?

src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+            message:
+              'Settings successfully updated. You will no longer receive notification emails',
+          });
+          return;
+        }
+        pushToast({
+          message: 'Settings successfully updated. You will now receive notification emails',
+        });
+        return;
+      } catch (error) {
```
tbezman, line 104: can we tell the user something went wrong?

src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+        setIsPending(false);
+      }
+    },
+    [isEmailUnsubscribedFromAll, pushToast, reportError, updateEmailNotificationSettings]
+  );
+
+  const toggleEmailNotification = useCallback(() => {
+    // Invert the value from subscribe to unsubscribe
+    handleEmailNotificationChange(!isEmailNotificationChecked);
+  }, [handleEmailNotificationChange, isEmailNotificationChecked]);
```
tbezman, line 120: You're passing this callback to `<Toggle onChange>`. That callback seems to pass the new value as a parameter, if we use that parameter instead of relying on `isEmailNotificationChecked`, then `toggleEmailNotification` will be stable and won't cause re-renders.

src/components/Email/useVerifyEmailOnPage.ts
```diff
@@ -0,0 +1,97 @@
...
+    store,
+    response
+  ) => {
+    if (response.verifyEmail?.__typename === 'VerifyEmailPayload' && response.verifyEmail?.email) {
+      store
+        .get('client:root:viewer')
+        ?.getLinkedRecord('email')
+        ?.setValue('Verified', 'verificationStatus');
+    }
+  };
```
tbezman, line 60: I don't think we need a custom updater here? We should be able to use `optimisticResponse` instead since we're not doing anything too crazy.

`optimisticResponse` should be strongly typed, you can use it like this
```ts
someMutation({
  variables: {},
  optimisticResponse: {
     verifyEmail: {
       // your data here
     }
  }
})
```

As long as you put `@raw_response_type` at the end of your mutation (which you did), the types for the `optimisticResponse` field will be generated for you.

src/contexts/globalLayout/GlobalNavbar/ProfileDropdown/ProfileDropdownContent.tsx
```diff
@@ -56,32 +58,61 @@ export function ProfileDropdownContent({ showDropdown, onClose, queryRef }: Prop
...
+    );
+  }, [hideModal, showModal, query]);
 
   const handleNotificationsClick = useCallback(() => {
     showModal({
       content: <NotificationsModal fullscreen={isMobile} />,
       isFullPage: isMobile,
       isPaddingDisabled: true,
       headerVariant: 'standard',
+      headerActions: isEmailFeatureEnabled ? notificationModalActions : false,
```
tbezman, line 98: instead of making `showModal` take a one off prop, can we modify the one off `NotificationsModal` to show the right stuff? this way, we don't keep bloating the `showModal` function with more options.

src/components/Email/EmailForm.tsx
```diff
@@ -17,9 +17,10 @@ import { HStack, VStack } from '../core/Spacer/Stack';
 type Props = {
   setIsEditMode: (editMode: boolean) => void;
   queryRef: EmailFormFragment$key;
+  onClosed?: () => void;
```
Robinnnnn, line 20: nit: rename => onClose

src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+import useUpdateEmailNotificationSettings from './useUpdateEmailNotificationSettings';
+
+type Props = {
+  queryRef: ManageWalletModalWithEmailFragment$key;
+  newAddress?: string;
+  onEthAddWalletSuccess?: () => void;
+  onTezosAddWalletSuccess?: () => void;
+};
+
+function ManageWalletsModalWithEmail({
```
Robinnnnn, line 25: nit: should we just call this `SettingsModal`?

src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const [isShowAddEmail, setIsShowAddEmail] = useState(false);
+  const { pushToast } = useToastActions();
+  const reportError = useReportError();
+
+  // Invert the value since in the db stored as unsubscribe while in the UI it's subscribe
+  const isEmailNotificationSubscribed =
+    query?.viewer?.email?.emailNotificationSettings?.unsubscribedFromNotifications || false;
+
+  const isEmailUnsubscribedFromAll =
+    !!query?.viewer?.email?.emailNotificationSettings?.unsubscribedFromAll ?? false;
```
Robinnnnn, line 64: hmm i think this can just be the following, without the need for fallbacks, because the `!` will coerce them into booleans:

```js
const isEmailNotificationSubscribed = !query?.viewer?.email?.emailNotificationSettings?.unsubscribedFromNotifications

const isEmailUnsubscribedFromAll = !!query?.viewer?.email?.emailNotificationSettings?.unsubscribedFromAll
```

src/components/Email/ManageWalletModalWithEmail.tsx
```diff
@@ -0,0 +1,193 @@
...
+            </BaseM>
+            <Toggle
+              checked={isEmailNotificationChecked}
+              isPending={isPending}
+              onChange={toggleEmailNotification}
+            />
+          </HStack>
+        </VStack>
+        <StyledButtonContaienr>
+          {isShowAddEmail ? (
```
Robinnnnn, line 148: can we rename this `shouldDisplayAddEmailInput`

src/components/NotificationsModal/NotificationEmailAlert.tsx
```diff
@@ -0,0 +1,81 @@
...
+    hideModal();
+
+    showModal({
+      content: <ManageWalletsModal queryRef={query} />,
+      headerText: 'Manage accounts',
+    });
+  };
+  const handleDismiss = () => {
+    onDismiss();
+  };
```
Robinnnnn, line 43: nit: could we wrap these in `useCallback`?

src/components/core/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,70 @@
...
+type Props = {
+  checked: boolean;
+  isPending?: boolean;
+  onChange: (checked: boolean) => void;
+} & React.InputHTMLAttributes<HTMLInputElement>;
+
+export default function Toggle({ checked, isPending, onChange, ...inputProps }: Props) {
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    onChange(e.target.checked);
+  };
```
Robinnnnn, line 14: useCallback

src/contexts/toast/ToastContext.tsx
```diff
@@ -34,19 +35,23 @@ type ToastType = {
   message: string;
   onDismiss: DismissToastHandler;
   autoClose: boolean;
+  variant: 'success' | 'error';
```
Robinnnnn, line 38: nit: looks like this is repeated in Line 11

could do `type ToastVariant = 'success' | 'error'` and use that

Robinnnnn (commented):
amazing work jakz!!

1. in the video, it looks like the modal "stutters" a little bit – does this happen regularly?

2. the change from "Pending" to "Verified" looks like it causes some layout shift, and the height of the section decreases a bit – can we make sure the height remains consistent during verification?

image

image

src/components/Email/ManageWalletModalWithEmail.tsx
jakzaizzat: Thats a good idea @tbezman! I've updated with this pattern. Definitely much easier to maintain!

src/components/Email/ManageWalletModalWithEmail.tsx
jakzaizzat: TIL: Didn't know that the relay store affect the local state too. 

I've removed this `useEffect` and set the response as a initial local state.

src/components/Email/ManageWalletModalWithEmail.tsx
jakzaizzat: Updated it!

src/components/Email/ManageWalletModalWithEmail.tsx
jakzaizzat: Good idea! Updated to use native checked value instead of `isEmailNotificationChecked`

src/contexts/globalLayout/GlobalNavbar/ProfileDropdown/ProfileDropdownContent.tsx
jakzaizzat: Make sense! Move it to its own hooks/component.

src/components/Email/ManageWalletModalWithEmail.tsx
jakzaizzat: Removed the fallback!

jakzaizzat:
Thanks @Robinnnnn! 

1. Regarding the slutters modal, yes its happens 🤦 . I've updated to make sure its open the modal first and then load the user gallery. 

https://user-images.githubusercontent.com/4480258/201608433-aba37767-19c5-4724-8320-9cd11391f0f8.mp4

2. Layout shift: Updated it!

src/components/Email/useVerifyEmailOnPage.ts
jakzaizzat: @tbezman  Make sense! Since its just a simple updated. 

In this case, we want to update the `verificationStatus` but it doesn't exist inside the `VerifyEmailPayload`

```
type VerifyEmailPayload {
    email: String
}
```

What's your recommend on this? 

1. Should we ask backend to return the verification status, so we can used the `optimisticResponse`
2. Is there any way to update the value of `verificationStatus` from `pending` to `verified` without using `updater`?

The goal is to updated the verification status in this modal (video below):

https://user-images.githubusercontent.com/4480258/201610416-fa50c190-1aa5-4ce4-a109-3e988ad05858.mp4

PR #1109, opened 2022-11-10
Add collection updated component

jarrel-b:
Adds the component to handle the `CollectionUpdatedFeedEvent`. This event always guarantees that there will be a new collectors note and new tokens, otherwise it'll be categorized as their respective feed events by the backend, so no need to handle the cases where either is missing.

Please tear it apart and let me know if I did anything wrong 😄 

The event in all its glory:

Screen Shot 2022-11-09 at 9 28 19 PM


kaitoo1:
Screen Shot 2022-11-10 at 14 43 44

wowowowowow indeed


this is so sick!!! our team is OP

jakzaizzat (approved):
LFG Jarrel!

kaitoo1 (approved):

PR #1111, opened 2022-11-11
Role Feature Flags

tbezman:
Pretty simple, made a new map of `Role => Flag => Boolean`. 

I also updated the `isFeatureEnabled` function to return `true` if any of the checks are `true`. This fixes the annoying situation where I update `PROD_FLAGS` but don't update `EMPLOYEE_FLAGS` and then none of the employees at Gallery can access it

Re: the schema changes
I added a script to fetch the latest schema from `main`. this should stop people's editors from formatting the file. 

In this PR, I just copied the file directly and I didn't see any formatting happening. So I'm not sure who the culprit is 🕵️

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "lint": "eslint . --ext .js,.ts,.tsx,.cjs",
     "typecheck": "nextjs-routes; graphql-codegen; tsc --noEmit",
     "relay": "mkdir __generated__; relay-compiler",
+    "fetch-schema": "curl https://raw.githubusercontent.com/gallery-so/go-gallery/main/graphql/schema/schema.graphql > schema.graphql",
     "relay:watch": "mkdir __generated__; relay-compiler --watch",
     "synpress:open": "synpress open --configFile cypress.config.ts",
     "synpress:run": "synpress run  --configFile cypress.config.ts",
--- a/src/utils/graphql/isFeatureEnabled.tsx
+++ b/src/utils/graphql/isFeatureEnabled.tsx
@@ -18,11 +18,6 @@ const DEV_FLAGS: Record<FeatureFlag, boolean> = {
   EMAIL: true,
 };
 
-const EMPLOYEE_FLAGS: Record<FeatureFlag, boolean> = {
-  WHITE_RHINO: true,
-  EMAIL: true,
-};
-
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -7,6 +7,7 @@
     "lint": "eslint . --ext .js,.ts,.tsx,.cjs",
     "typecheck": "nextjs-routes; graphql-codegen; tsc --noEmit",
     "relay": "mkdir __generated__; relay-compiler",
+    "fetch-schema": "curl https://raw.githubusercontent.com/gallery-so/go-gallery/main/graphql/schema/schema.graphql > schema.graphql",
```
Robinnnnn, line 10: 😍

src/utils/graphql/isFeatureEnabled.tsx
```diff
@@ -1,36 +1,45 @@
...
+  ADMIN: {
+    WHITE_RHINO: true,
+    EMAIL: true,
+  },
+  BETA_TESTER: {
+    WHITE_RHINO: true,
+    EMAIL: true,
+  },
+  // In case a new role gets added, how do we want to handle that new role
+  // if the client hasn't been updated to handle the new role.
```
Robinnnnn, line 36: lazy solution: deploy backend first then update frontend

longer term: have backend return specific feature flags

src/utils/graphql/isFeatureEnabled.tsx
```diff
@@ -1,36 +1,45 @@
...
 };
 
+const ROLE_FLAGS: Record<Role, Record<FeatureFlag, boolean>> = {
+  ADMIN: {
+    WHITE_RHINO: true,
+    EMAIL: true,
+  },
+  BETA_TESTER: {
+    WHITE_RHINO: true,
+    EMAIL: true,
```
Robinnnnn, line 33: let's keep `EMAIL: false` for beta testers

Robinnnnn (approved):

src/utils/graphql/isFeatureEnabled.tsx
tbezman: @Robinnnnn This comment is trying to explain that "the below object is what will be used in the case that the backend deployed something that the frontend doesn't know about"

I'm not literally asking a question, maybe I should change my language here?

PR #1112, opened 2022-11-11
White rhino UI fixes

tbezman:
### Increase font size of "2 admires" to fix jumpy feeling text
https://user-images.githubusercontent.com/6754223/201377982-968c8644-a94b-4e60-923e-2042213da4e8.mov

### Fix alignment of header text on `NotificationsModal`
image

### DIFF
--- a/src/contexts/modal/ModalContext.tsx
+++ b/src/contexts/modal/ModalContext.tsx
@@ -175,7 +175,7 @@ function ModalProvider({ children }: Props) {
     if (isModalOpenRef.current) {
       clearAllModals();
     }
-  }, [route, hideModal]);
+  }, [route, clearAllModals]);
 
   /**
    * EFFECT: Prevent main body from being scrollable while any modals are open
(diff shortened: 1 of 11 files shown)


Robinnnnn:
@tbezman for the admires, can we make the behavior:
- "kaito admired this"
- add a like
- "kaito and 1 other admired this"
- someone else likes it
- "kaito and 2 others admired this"

instead of (2 admired this, 3 admired this, etc.)

if this is a bigger lift we can add this to the white rhino linear board

src/contexts/modal/ModalContext.tsx
```diff
@@ -173,10 +173,9 @@ function ModalProvider({ children }: Props) {
   const route = useStabilizedRouteTransitionKey();
   useEffect(() => {
     if (isModalOpenRef.current) {
-      // bypass onClose as to not navigate the user back mid-route change
-      hideModal({ bypassOnClose: true });
+      clearAllModals();
     }
-  }, [route, hideModal]);
+  }, [route, clearAllModals]);
```
tbezman, line 178: @Robinnnnn any reason why this would be spicy?

PR #1113, opened 2022-11-11
enable marking as spam

jackiehluo:
adds the ability to show/hide collections in the collection editor!

collected view:
Screen Shot 2022-11-11 at 2 02 58 PM

hidden view:
Screen Shot 2022-11-11 at 2 23 52 PM

selector open state:
Screen Shot 2022-11-11 at 2 03 15 PM


src/components/ManageGallery/OrganizeCollection/Sidebar/SidebarTokens.tsx
```diff
@@ -45,6 +50,26 @@ export const SidebarTokens = ({
     tokenRefs
   );
 
+  const setSpamPreference = useSetSpamPreference();
+  const setSpamPreferenceForCollection = useCallback(
+    (address: string, isSpam: boolean) => {
+      const groups = groupCollectionsByAddress({ tokens, editModeTokens });
+      const matchingCollection = groups.find((group) => group.address === address);
```
tbezman, line 57: correct me if i'm wrong but I think there's some indirection here. 

we have an address, and a list of tokens. we then group the tokens by address into a list. then we search for that group to get the tokens.

I think we can just filter the list of tokens by their contract address.

src/components/ManageGallery/OrganizeCollection/Sidebar/SidebarViewSelector.tsx
```diff
@@ -0,0 +1,55 @@
...
+  setSelectedView: (selectedView: SidebarView) => void;
+};
+
+export function SidebarViewSelector({ selectedView, setSelectedView }: SidebarViewSelectorProps) {
+  const [isDropdownOpen, setIsDropdownOpen] = useState(false);
+
+  const onSelectView = (selectedView: SidebarView) => {
+    setSelectedView(selectedView);
+    setIsDropdownOpen(false);
+  };
```
tbezman, line 23: nit: usecallback

src/components/core/Dropdown/Dropdown.tsx
```diff
@@ -10,9 +10,10 @@ type Props = {
   onClose: () => void;
   children?: ReactNode;
   position: 'right' | 'left';
+  style?: React.CSSProperties;
```
tbezman, line 13: maybe instead of passing in a custom `style`, we update `position` to have another `full-width` option to compete with `left` and `right`.

I imagine this won't be the last time we want a full width version of the dropdown

src/components/ManageGallery/OrganizeCollection/Sidebar/SidebarList.tsx
```diff
@@ -203,3 +282,17 @@ const Selection = styled.div`
...
+  margin-left: auto;
+`;
+
+const StyledTooltip = styled(Tooltip)<{ visible: boolean }>`
+  opacity: ${({ visible }) => (visible ? 1 : 0)};
+  width: 50%;
+  right: 0;
+  top: 30px;
+  z-index: 1;
+`;
```
Robinnnnn, line 298: could we make the tooltip a bit wider? it's 130px in figma

image

wondering if we could position it this way as well:

image

src/hooks/api/tokens/useSetSpamPreference.ts
```diff
@@ -0,0 +1,30 @@
...
+      }
+    `
+  );
+
+  return useCallback(
+    async (input: SetSpamPreferenceInput) => {
+      await setSpamPreference({ variables: { input } });
+
+      // TODO: handle error cases
+    },
```
Robinnnnn, line 27: we'll want to make sure this is updated optimistically so that the collection is immediately moved to the hidden folder. @tbezman can you point to any good precedents in the codebase?

Robinnnnn (commented):
for the empty view on the hidden tab, could we just toss the copy? 

Screen Shot 2022-11-11 at 5 06 20 PM

src/components/ManageGallery/OrganizeCollection/Sidebar/SidebarTokens.tsx
jackiehluo: oh nope haha you are 100% right! thanks for catching that, will update

src/components/ManageGallery/OrganizeCollection/Sidebar/SidebarViewSelector.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <Container>
+      <Selector onClick={() => setIsDropdownOpen(true)}>
+        <BaseM>{selectedView}</BaseM>
+        <Arrows style={{ marginRight: 4 }} />
```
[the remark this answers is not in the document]
jackiehluo: oooh okay yes missed this whole subdirectory! okay sweet

src/components/ManageGallery/OrganizeCollection/Sidebar/SidebarList.tsx
jackiehluo: ooh yes!

and re: positioning, i think `<Tooltip />` would need to be reworked to use a react portal so that it can render outside of the sidebar! i was worried it might affect other places it's used in unexpected ways, but i can def add that—thoughts? i think it does make sense to do that in general, probably also for `<Dropdown />`

an easier/more reliable way to handle it versus rolling it from scratch would be using a library (i.e., https://floating-ui.com/docs/floatingportal) that handles positioning, focus, etc., out of the box

src/hooks/api/tokens/useSetSpamPreference.ts
jackiehluo: haha oh yeah i had a few ideas of how to do it but all of them felt janky—but just now reading the relay docs more and seeing `optimisticResponse`

jackiehluo:
okay i addressed most of the comments! only parts unresolved are tooltip positioning and optimistic updates (i added an `optimisticResponse`, but as far as i can tell, it's silently failing? @tbezman is there's anything obvious i missed?)

Screen Shot 2022-11-12 at 12 17 57 PM

jackiehluo:
fixed the optimistic updating with @tbezman's help! thinking more about the tooltip, i think that's probably best suited for its own project and out of scope for this task?

src/components/ManageGallery/OrganizeCollection/Sidebar/SidebarList.tsx
Robinnnnn: ahh wow forgot about the sidebar, and great point about using a portal - that sounds great!

i don't think this is blocking, so i'll create a separate ticket with the floating portal suggestion as a fast-follow :) no need to handle in this PR

Robinnnnn (approved):
absolutely crushed it 💪 💪 💪 💪

PR #1120, opened 2022-11-16
[GAL-670] Email Follow Up

jakzaizzat:
**If no email attached, show the toggle but disabled it**

CleanShot 2022-11-16 at 12 21 08

**Disabled toggle & turn off if the email not verified**

https://user-images.githubusercontent.com/4480258/202082185-09e52bd8-9226-49de-97c9-0c536ac8c08c.mp4

**Instantly updated the verification status**

https://user-images.githubusercontent.com/4480258/202081964-6ef7da88-ed05-4bc9-9c01-b81ae2f00f47.mp4

**Disabled toggle color**

CleanShot 2022-11-16 at 13 01 55

### DIFF
--- a/src/components/Email/EmailForm.tsx
+++ b/src/components/Email/EmailForm.tsx
@@ -40,13 +40,14 @@ function EmailForm({ setIsEditMode, queryRef, onClose }: Props) {
   );
 
   const [updateEmail] = usePromisifiedMutation<EmailFormMutation>(graphql`
-    mutation EmailFormMutation($input: UpdateEmailInput!) {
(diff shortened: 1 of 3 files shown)


kaitoo1:
nice!!!
small comment: I think the disabled toggle should have a lighter color
it looks like right now it's achieved via opacity, can we change it to use the `porcelain` color instead? it looks like that's the spec color

Robinnnnn (approved):

jakzaizzat:
Nice catch @kaitoo1! Updated the disabled toggle to use `porcelain` color and removed `opacity`

PR #1124, opened 2022-11-16
fix width of notifications popover

kaitoo1:
This PR fixes the width of the notifications popover instead of using min-width, so that content does not stretch the popover out.
Before:
Screen Shot 2022-11-16 at 14 50 33

After:
Screen Shot 2022-11-16 at 14 49 37


Robinnnnn:
image

kaitoo1:
good catch fixed
Screen Shot 2022-11-16 at 15 00 29

Robinnnnn (approved):

PR #1132, opened 2022-11-17
[GAL-679] Unsubscribe email

jakzaizzat:
## Unsubscribe token from email

URL: [http://localhost:3000/jakz?settings=true&unsubscribeToken=1234](http://localhost:3000/jakz?settings=true&unsubscribeToken=1234)

**Logged in user**

https://user-images.githubusercontent.com/4480258/202331083-6cec6351-fec4-4c10-9824-013a9d14788b.mp4

**If the user not signed in**

https://user-images.githubusercontent.com/4480258/202331272-c44fe328-7a58-46ea-be84-1c07b4edd337.mp4

### DIFF
--- a/pages/[username]/index.tsx
+++ b/pages/[username]/index.tsx
@@ -3,6 +3,7 @@ import { route } from 'nextjs-routes';
 import { useLazyLoadQuery } from 'react-relay';
 import { graphql } from 'relay-runtime';
 
+import useVerifyEmailOnPage from '~/components/Email/useVerifyEmailOnPage';
 import GalleryViewEmitter from '~/components/internal/GalleryViewEmitter';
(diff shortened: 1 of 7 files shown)


Robinnnnn (commented):
nit:

1. can we actually _**optimistically update**_ the unsubscribed status, so the settings modal _immediately_ shows that the user has unsubscribed? then, if the request fails, the relay cache will update and revert the user to the subscribed state (and we can display an error message). i think we could use `optimisticUpdater`
2. can we have the "you have unsubscribed" toast not auto-dismiss?

jakzaizzat:
Thanks @Robinnnnn!

1. Updated to use `optimisticUpdater`. It will revert back if the request failed
2. Sure! Turn auto dismiss to the toast

Robinnnnn (approved):

PR #1151, opened 2022-11-21
[GAL-718] Fix save button for toggle live render

jakzaizzat:
**Changes**
- Add `collectionMetadataState` too to keep track of token metadata change
- Move the check unsaved part into own hook

**Demo**

https://user-images.githubusercontent.com/4480258/202956927-1a4884d1-1d87-4852-80ad-e70bf4a8debc.mp4


Robinnnnn (approved):
🔥🔥🔥🔥 great catch!

PR #1159, opened 2022-11-23
[GAL-702] Split user follow user event

jakzaizzat:
**Before**

CleanShot 2022-11-23 at 09 08 01

**After**

CleanShot 2022-11-23 at 09 03 16


Robinnnnn (approved):
yeeeee thank you ser!!

kaitoo1 (approved):

PR #1164, opened 2022-11-23
[GAL-709] fix bio appears weird in the follower list modal

jakzaizzat:
**Changes**
Remove all the breakline for bio **only** in follower list modal

**Before**
CleanShot 2022-11-23 at 10 06 39

**After**
CleanShot 2022-11-23 at 10 04 19

### DIFF
--- a/src/components/Follow/FollowListUsers.tsx
+++ b/src/components/Follow/FollowListUsers.tsx
@@ -35,12 +35,16 @@ export default function FollowListUsers({
     track('Follower List Username Click');
   }, [track]);
 
+  const trimBreaklines = useCallback((text: string) => {
+    return text.replace(/(\r\n|\n|\r|\\\n)/gm, '');
+  }, []);
+
   return (
     <StyledList>
       {users.map((user) => (
         <StyledListItem key={user.dbid} href={`/${user.username}`} onClick={handleClick}>
           <TitleS>{user.username}</TitleS>
-          <StyledBaseM>{user.bio && <Markdown text={user.bio} />}</StyledBaseM>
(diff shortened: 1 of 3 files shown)


src/components/Follow/FollowListUsers.tsx
```diff
@@ -35,12 +36,16 @@ export default function FollowListUsers({
     track('Follower List Username Click');
   }, [track]);
 
+  const trimBreaklines = useCallback((text: string) => {
+    return text.replace(BREAK_LINES, '');
+  }, []);
```
Robinnnnn, line 41: we should just be able to useMemo here!

```js
const bio = useMemo(() => user.bio.replace(BREAK_LINES, ''), [user.bio])
```

src/utils/regex.ts
```diff
@@ -10,3 +10,6 @@ export const VALID_URL =
 
 // standard email format x@x.xx
 export const EMAIL_FORMAT = /^[\w-\.]+@([\w-]+\.)+[\w-]{2,4}$/;
+
+// break lines
+export const BREAK_LINES = /(\r\n|\n|\r|\\\n)/gm;
```
Robinnnnn, line 15: nit: would love to see some unit tests with this

src/components/Follow/FollowListUsers.tsx
jakzaizzat: fml. why did I used `useCallback` here 🤦🏻‍♂️

jakzaizzat: Updated!

src/utils/regex.ts
jakzaizzat: Added unit test for this regex!

Robinnnnn (approved):
LFG

PR #1168, opened 2022-11-24
[GAL-724] Only track "Page View" events if user is signed in

jakzaizzat:
**Changes**

- Added `checkAuth` options to `tracking`
- All the `Page View` will be recorded only its authenticated

### DIFF
--- a/src/contexts/analytics/AnalyticsContext.tsx
+++ b/src/contexts/analytics/AnalyticsContext.tsx
@@ -7,7 +7,7 @@ import noop from '~/utils/noop';
 
 type EventProps = Record<string, unknown>;
 
-type TrackFn = (eventName: string, eventProps?: EventProps) => void;
+type TrackFn = (eventName: string, eventProps?: EventProps, checkAuth?: boolean) => void;
 
 const AnalyticsContext = createContext<TrackFn | undefined>(undefined);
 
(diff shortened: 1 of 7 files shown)


src/contexts/analytics/AnalyticsContext.tsx
```diff
@@ -7,7 +7,7 @@ import noop from '~/utils/noop';
 
 type EventProps = Record<string, unknown>;
 
-type TrackFn = (eventName: string, eventProps?: EventProps) => void;
+type TrackFn = (eventName: string, eventProps?: EventProps, checkAuth?: boolean) => void;
```
jakzaizzat, line 10: Added new props here. 

Lmk if we wanna convert this props to `Object` instead of multiple parameters. If yes, need to refactor all the `useTrack` usage

src/contexts/analytics/AnalyticsContext.tsx
```diff
@@ -70,7 +70,12 @@ const AnalyticsProvider = memo(({ children }: Props) => {
   const userIdRef = useRef(userId);
   userIdRef.current = userId;
 
-  const handleTrack: TrackFn = useCallback((eventName, eventProps = {}) => {
+  const handleTrack: TrackFn = useCallback((eventName, eventProps = {}, checkAuth = false) => {
+    // if there is auth, and we don't have a user id, don't track
```
kaitoo1, line 74: 🧠

src/contexts/analytics/AnalyticsContext.tsx
kaitoo1: im ok with this for now!

kaitoo1 (approved):

PR #1169, opened 2022-11-24
[GAL-726] Wrap the comment

jakzaizzat:
**Before**

CleanShot 2022-11-24 at 15 49 00@2x

**After**

CleanShot 2022-11-24 at 15 47 29@2x


Robinnnnn (approved):
🙌

PR #1173, opened 2022-11-29
[GAL-727] Fix feed image event overflow

jakzaizzat:
**Findings**

This bug only happens on desktop with 3 images. 

**Before 1**

CleanShot 2022-11-29 at 11 08 57

**After 1**

CleanShot 2022-11-29 at 11 08 41

**Before 2**

CleanShot 2022-11-29 at 11 10 20

**After 2**

CleanShot 2022-11-29 at 11 10 35


Robinnnnn (approved):
gorgeous

PR #1174, opened 2022-11-29
Fix feed overflow width

jakzaizzat:
There is feed width overflow on desktop

**Before**

CleanShot 2022-11-30 at 07 41 00

**After**

CleanShot 2022-11-30 at 07 41 44


Robinnnnn (approved):
LETS GO

PR #1179, opened 2022-12-05
[GAL-703] feat: add deep linking in home collections

evgenyboxer:
The collections in the `/home` route now will replace the URL with the collection route. When refreshed the close button will correctly send the user back to the `/home` route (done via a new "returnTo" param). The refresh causes the lost of the scroll position - I think its possible to add scroll restoration later on.

### DIFF
--- a/pages/[username]/[collectionId]/[tokenId]/index.tsx
+++ b/pages/[username]/[collectionId]/[tokenId]/index.tsx
@@ -1,6 +1,7 @@
 import { GetServerSideProps } from 'next';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
+import { Route } from 'nextjs-routes';
 import { useCallback, useMemo } from 'react';
 import styled from 'styled-components';
 
@@ -19,23 +20,27 @@ type NftDetailPageProps = MetaTagProps & {
 };
 
 export default function NftDetailPage({ username, collectionId, tokenId }: NftDetailPageProps) {
-  const { push } = useRouter();
+  const { push, query } = useRouter();
 
   // the default "back" behavior from the NFT Detail Page
-  // is a redirect to the Collection Page
-  const collectionRoute = useMemo(
-    () => ({
+  // is a redirect to the Collection Page or a valid returnTo route
+  const returnToRoute = useMemo<Route>(() => {
+    if (query?.returnTo === 'home') {
+      return {
+        pathname: '/home',
+      };
+    }
+    return {
       pathname: '/[username]/[collectionId]',
       query: { username, collectionId },
-    }),
-    [username, collectionId]
-  );
+    };
+  }, [username, collectionId, query?.returnTo]);
 
-  const handleReturnToCollectionPage = useCallback(() => {
-    push(collectionRoute);
-  }, [collectionRoute, push]);
+  const handleReturnToRoute = useCallback(() => {
+    push(returnToRoute);
+  }, [returnToRoute, push]);
 
-  useKeyDown('Escape', handleReturnToCollectionPage);
+  useKeyDown('Escape', handleReturnToRoute);
 
   if (!tokenId) {
     // Something went horribly wrong
@@ -44,7 +49,7 @@ export default function NftDetailPage({ username, collectionId, tokenId }: NftDe
 
   return (
     <>
-      <Link href={collectionRoute}>
+      <Link href={returnToRoute}>
         <StyledDecoratedCloseIcon />
       </Link>
       <GalleryRoute
--- a/src/components/Feed/Events/FeedEventNftPreviewWrapper.tsx
+++ b/src/components/Feed/Events/FeedEventNftPreviewWrapper.tsx
@@ -45,16 +45,20 @@ function FeedEventNftPreviewWrapper({ tokenRef, maxWidth, maxHeight }: Props) {
 
   const { showModal } = useModalActions();
 
-  const handleClick = useCallback(() => {
-    showModal({
-      content: (
-        <StyledNftDetailViewPopover>
-          <NftDetailView authenticatedUserOwnsAsset={false} queryRef={token} />
-        </StyledNftDetailViewPopover>
-      ),
-      isFullPage: true,
-    });
-  }, [showModal, token]);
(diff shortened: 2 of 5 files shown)


src/components/Feed/Events/FeedEventNftPreviewWrapper.tsx
```diff
@@ -45,16 +45,20 @@ function FeedEventNftPreviewWrapper({ tokenRef, maxWidth, maxHeight }: Props) {
...
+  const handleClick = useCallback(
+    ({ onClose }: { onClose?: () => void }) => {
+      showModal({
+        content: (
+          <StyledNftDetailViewPopover>
+            <NftDetailView authenticatedUserOwnsAsset={false} queryRef={token} />
+          </StyledNftDetailViewPopover>
+        ),
+        isFullPage: true,
+        onClose,
```
evgenyboxer, line 57: I needed to add an `onClose` callback to the modal so I can control the URL location after the user closes the modal.

src/utils/browserHistory.ts
```diff
@@ -0,0 +1,11 @@
+export function historyPushState(url: string | URL) {
+  if (typeof window !== 'undefined') {
+    window.history.pushState({}, '', url);
+  }
+}
+
+export function historyReplaceState(url: string | URL) {
+  if (typeof window !== 'undefined') {
```
evgenyboxer, line 8: To make Next.js happy.

pages/[username]/[collectionId]/[tokenId]/index.tsx
```diff
@@ -19,23 +20,27 @@ type NftDetailPageProps = MetaTagProps & {
...
+      };
+    }
+    return {
       pathname: '/[username]/[collectionId]',
       query: { username, collectionId },
-    }),
-    [username, collectionId]
-  );
+    };
+  }, [username, collectionId, query?.returnTo]);
```
Comment (Robinnnnn) [inline pages/[username]/[collectionId]/[tokenId]/index.tsx:37]: Q: should the previous route just specify the entire `returnTo`?

e.g. set `returnTo` literally as `/home` or `/evgency/12345-6789`?

typescript may not be happy but could help with co-location. then we don't even need this `returnToRoute` logic

src/components/NftPreview/NftPreview.tsx
```diff
@@ -127,10 +129,31 @@ function NftPreview({
...
+          route({
+            pathname: '/[username]/[collectionId]/[tokenId]',
+            query: {
+              username: username as string,
+              collectionId,
+              tokenId: token.dbid,
+              returnTo: 'home',
+            },
+          })
+        );
```
Robinnnnn, line 144: any reason we're using replace? i think it's reasonable to assume that clicking "back" would just return the user to /home.

this is the behavior of a collection page as well:
1. visit collection: https://gallery.so/bastiboii/2GytPwuD9sxE2n3cd1yJdh2L2PC
2. click on NFT
3. click back
4. end up back on collection

pages/[username]/[collectionId]/[tokenId]/index.tsx
Comment (evgenyboxer) [reply pages/[username]/[collectionId]/[tokenId]/index.tsx:37]: Yeah I kinda did it to make sure you only pass valid paths. But yeah, I could make sure the previous page passes a query, but also - we prob need a fallback? If so, it prob won't change the logic that much.

Essentially we will have something like

```
  const returnToRoute = useMemo<Route>(() => {
    if (query?.returnTo != null) {
      return {
        pathname: query.returnTo,
      };
    }
    return {
      pathname: '/[username]/[collectionId]',
      query: { username, collectionId },
    };
  }, [username, collectionId, query?.returnTo]);
  ```

src/components/NftPreview/NftPreview.tsx
evgenyboxer: Oh right I forgot to mention why I used replace!

So, I'm not entirely sure why but when I use `pushState` the scroll position is lost as soon as you try to move it. So like everything works (the back button etc), but when you try to scroll away it just scrolls to the top.

I thought to have this as a somewhat temp solution until we fix this issue.

Robinnnnn: ahhh i remember now. it's because on the main gallery page we use `LinkToNftDetailView` instead of imperatively pushing. 
- [component](https://github.com/gallery-so/gallery/blob/main/src/scenes/NftDetailPage/LinkToNftDetailView.tsx)
- [usage](https://github.com/gallery-so/gallery/blob/3dc745d80b407579db61e8f80ea6f68ed351c8ca/src/components/NftPreview/NftPreview.tsx#L204)
- make sure it's paired with [this hook](https://github.com/gallery-so/gallery/blob/3dc745d80b407579db61e8f80ea6f68ed351c8ca/src/scenes/NftDetailPage/useDisplayFullPageNftDetailModal.tsx#L12)
- [usage of the hook](https://github.com/gallery-so/gallery/blob/3dc745d80b407579db61e8f80ea6f68ed351c8ca/src/scenes/UserGalleryPage/UserGallery.tsx#L70)

does this make sense? i know it's a bit confusing. we should probably pair / co-locate the component + hook together

pages/[username]/[collectionId]/[tokenId]/index.tsx
Comment (Robinnnnn) [reply pages/[username]/[collectionId]/[tokenId]/index.tsx:37]: ahh i gotchu, makes sense. okay i think this all relates to [this comment](https://github.com/gallery-so/gallery/pull/1179/files#r1041157039) further below (read it first!)

we have a [hook](https://github.com/gallery-so/gallery/blob/3dc745d80b407579db61e8f80ea6f68ed351c8ca/src/scenes/NftDetailPage/useDisplayFullPageNftDetailModal.tsx#L12) that we need to use to make the modal work, and it has a similar `returnTo` logic. i think we could refactor all that and have the user always pass in a `returnTo` prop into the hook!

in fact maybe we could co-locate / encapsulate everything within the `LinkToNftDetailView` component (described in the comment). so basically the component itself could take a `returnTo` prop, use the hook, etc.

tbezman:
@evgenyboxer Closing this for now. If you feel like picking up on it lemme know and we can re-open it.

PR #1189, opened 2022-12-06
chore: sentry improvements

Robinnnnn:
- [x] disable sentry reporting by default on localhost
  - the biggest offender to our quota was `/__nextjs_original-stack-frame`, which was only being reported from localhost
- [x] reduce sample rate to 5%
- [ ] upgrade sentry to v7


tbezman (approved):

Robinnnnn:
closing in lieu of #1190 + #1191

PR #1207, opened 2022-12-13
GAL-730 remove side padding from feed events on mobile

kaitoo1:
This PR removes the side padding from feed events on mobile, as requested by Design.
This also fixes an issue where images are clipped on mobile.

Before
CleanShot 2022-12-13 at 21 04 40@2x

After
CleanShot 2022-12-13 at 21 05 48@2x


Robinnnnn (approved):
Thank you so much

PR #1218, opened 2022-12-16
[GAL-623] Check valid email - Sendgrid

jakzaizzat:
https://user-images.githubusercontent.com/4480258/208125479-949cd942-b11a-46c9-b2b6-ace59c2bdda6.mp4

### DIFF
--- a/src/components/Email/EmailForm.tsx
+++ b/src/components/Email/EmailForm.tsx
@@ -1,5 +1,5 @@
 import { useRouter } from 'next/router';
-import { useCallback, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { graphql, useFragment } from 'react-relay';
 import { SelectorStoreUpdater } from 'relay-runtime';
 import styled from 'styled-components';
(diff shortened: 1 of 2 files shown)


src/components/Email/EmailForm.tsx
```diff
@@ -164,19 +190,24 @@ function EmailForm({ setIsEditMode, queryRef, onClose }: Props) {
...
+              <Button variant="secondary" disabled={savePending} onClick={handleCancelClick}>
+                Cancel
+              </Button>
+            )}
+            <Button
+              variant="primary"
+              disabled={!isValidEmail || savePending || savedEmail === email || isChecking}
+              onClick={handleSaveClick}
+            >
+              {isChecking ? <Spinner /> : 'Save'}
```
Robinnnnn, line 208: could we make the button width constant when the spinner appears? just to prevent layout shift!

in fact, i think we can use the built-in `pending` prop for buttons!

Screen Shot 2022-12-23 at 3 01 27 PM

Robinnnnn (commented):
i noticed in the video that you're clicking outside of the input field to wait for the email to be verified – what if someone immediately clicks "save" before the server returns validation?

**proposal: keep the button disabled by default, until all verification checks have passed**. that way someone can't quickly click Save after typing but before the debounce window.

src/components/Email/EmailForm.tsx
jakzaizzat: Thanks @Robinnnnn! Somehow I missed this props 🤦

jakzaizzat:
Updated @Robinnnnn!  I made the button used `disabled` state as default and set debounce value to `500ms`

PR #1228, opened 2022-12-20
Increase line clamp for community pages

Robinnnnn:
Also ensure line breaks are respected 

**Before**
image

**After**
image

### DIFF
--- a/src/scenes/CommunityPage/CommunityPageView.tsx
+++ b/src/scenes/CommunityPage/CommunityPageView.tsx
@@ -147,19 +147,12 @@ const StyledCommunityPageContainer = styled.div`
 `;
 
 const StyledBaseM = styled(BaseM)<{ showExpandedDescription: boolean }>`
-  -webkit-line-clamp: ${({ showExpandedDescription }) => (showExpandedDescription ? 'initial' : 4)};
+  -webkit-line-clamp: ${({ showExpandedDescription }) => (showExpandedDescription ? 'initial' : 5)};
   display: -webkit-box;
   -webkit-box-orient: vertical;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


src/scenes/CommunityPage/CommunityPageView.tsx
```diff
@@ -147,19 +147,12 @@ const StyledCommunityPageContainer = styled.div`
...
-  line-clamp: 4;
+  line-clamp: 5;
   display: -webkit-box;
-
-  // allows descriptions with multiple paragraphs to be line clamped properly
-  p,
-  ol,
-  li {
-    display: contents;
-  }
```
Robinnnnn, line 162: @kaitoo1 noticed you added this in the past, but this would prevent line breaks from being respected. i removed it and ensured everything looks normal in existing community pages that have descriptions (art gobblers, crypto coven, gallery awards)

src/scenes/CommunityPage/CommunityPageView.tsx
```diff
@@ -147,19 +147,12 @@ const StyledCommunityPageContainer = styled.div`
 `;
 
 const StyledBaseM = styled(BaseM)<{ showExpandedDescription: boolean }>`
-  -webkit-line-clamp: ${({ showExpandedDescription }) => (showExpandedDescription ? 'initial' : 4)};
+  -webkit-line-clamp: ${({ showExpandedDescription }) => (showExpandedDescription ? 'initial' : 5)};
   display: -webkit-box;
   -webkit-box-orient: vertical;
   overflow: hidden;
-  line-clamp: 4;
+  line-clamp: 5;
```
Robinnnnn, line 154: would look like this otherwise

image

PR #1229, opened 2022-12-21
Fix shop owned quantity

jakzaizzat:
**Notes**
This scenario only happens if the users transferred out his merch nft after they minted. 

**Changes**
Instead of getting the quantity from the **contract**,  we used the existing `getMerch` backend to map all the value

Example: My wallet only have 1 Merch NFT (TShirt)

**Before**

CleanShot 2022-12-21 at 13 57 17

**After**

CleanShot 2022-12-21 at 13 56 36


Robinnnnn (approved):
amazing find- thank you jakz!

PR #1243, opened 2023-01-06
Quiet Sentry Errors GAL-1647 GAL-1083 GAL-1055 GAL-758

tbezman:
- No longer reporting errors related to feed event failures
- Distinguish validation errors so we don't report those to Sentry
- Globally ignore the web3 rejection errors
- Did some refactoring on the NftDetailPage to support the not found stuff.


Robinnnnn (approved):
jesus dude, i knew you'd come up with an insanely beautiful and elegant solution

PR #1251, opened 2023-01-11
Turn on un-authed views

jarrel-b:
This PR re-enables tracking of un-authed views so that we can use it to track trending galleries.


jakzaizzat (approved):
LFG 🚀

Robinnnnn (approved):

PR #1262, opened 2023-01-18
Fix failing cypress tests

Robinnnnn:
- fix homepage test; now we're redirecting to a feed
- disable screenshots; they're failing for an unusual reason and we weren't getting short-term use out of it

### DIFF
--- a/.github/workflows/screenshots.yml
+++ b/.github/workflows/screenshots.yml
@@ -1,6 +1,11 @@
 name: screenshots
 on: [deployment_status]
 jobs:
+  # start disable ---
+  # temporarily disable workflow as the failure mode is unclear
+  # last time this job failed: https://github.com/gallery-so/gallery/actions/runs/3951382775/jobs/6765164811
+  if: ${{ false }}
+  # --- end disable
   run:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/screenshots.yml
```diff
@@ -1,6 +1,11 @@
 name: screenshots
 on: [deployment_status]
 jobs:
+  # start disable ---
+  # temporarily disable workflow as the failure mode is unclear
+  # last time this job failed: https://github.com/gallery-so/gallery/actions/runs/3951382775/jobs/6765164811
+  if: ${{ false }}
+  # --- end disable
```
Robinnnnn, line 8: disabling screenshots

cypress/integration/pages/home/home-page.ts
```diff
@@ -9,8 +9,8 @@ export default class HomePage extends Page {
     return cy.get('[data-testid="sign-in-button"]');
   }
 
-  getExploreButton() {
-    return cy.get('[data-testid="explore-button"]');
```
Robinnnnn, line 13: we deprecated our OG home page, so the explore button is no longer there

tbezman (approved):

PR #1283, opened 2023-01-24
Fix edit gallery links

tbezman:
- Had to thread through a `galleryRef` to the `GalleryNavbar` so the edit button is aware of which gallery we're talking about
- Updated the edit collection button to link to the new editor if you're in the FF
- Updated the new gallery editor to allow for an initial collection id so you get right to the content you want to edit

### DIFF
--- a/pages/[username]/activity.tsx
+++ b/pages/[username]/activity.tsx
@@ -47,7 +47,7 @@ export default function UserFeed({ username, eventId }: UserActivityProps) {
 
   return (
     <GalleryRoute
-      navbar={<GalleryNavbar username={username} queryRef={query} />}
+      navbar={<GalleryNavbar galleryRef={null} username={username} queryRef={query} />}
       element={
         <>
           <GalleryViewEmitter queryRef={query} />
--- a/pages/[username]/galleries/[galleryId].tsx
+++ b/pages/[username]/galleries/[galleryId].tsx
@@ -15,12 +15,13 @@ export default function FocusedGallery({ galleryId, username }: Props) {
   const query = useLazyLoadQuery<GalleryIdFocusedGalleryQuery>(
     graphql`
       query GalleryIdFocusedGalleryQuery($galleryId: DBID!, $username: String!) {
-        galleryById(id: $galleryId) {
+        galleryById(id: $galleryId) @required(action: THROW) {
           ... on Gallery {
-            name
             owner {
               username
             }
+
+            ...GalleryNavbarGalleryFragment
           }
         }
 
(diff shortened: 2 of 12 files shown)


pages/[username]/activity.tsx
```diff
@@ -47,7 +47,7 @@ export default function UserFeed({ username, eventId }: UserActivityProps) {
 
   return (
     <GalleryRoute
-      navbar={<GalleryNavbar username={username} queryRef={query} />}
+      navbar={<GalleryNavbar galleryRef={null} username={username} queryRef={query} />}
```
Comment (tbezman) [inline pages/[username]/activity.tsx:50]: anywhere we're passing `galleryRef={null}` it means there is no gallery in context here

pages/[username]/galleries/[galleryId].tsx
```diff
@@ -35,7 +36,8 @@ export default function FocusedGallery({ galleryId, username }: Props) {
     <GalleryRoute
       navbar={
         <GalleryNavbar
-          galleryName={query.galleryById?.name || undefined}
+          showGalleryNameBreadcrumb
+          galleryRef={query.galleryById}
```
Comment (tbezman) [inline pages/[username]/galleries/[galleryId].tsx:40]: instead of just passing the gallery name if we want to show it, we now have `showGalleryNameBreadcrumb` as a flag for the component.

We're always going to pass a gallery ref if we have one, so ^^ is an alternate way to flip the name on / off for `/username` vs `/username/galleries/galleryId`

pages/gallery/[galleryId]/edit.tsx
```diff
@@ -99,9 +99,10 @@ function NewEditGalleryPageInner({ queryRef }: NewEditGalleryPageInnerProps) {
 
 type Props = {
   galleryId: string;
+  initialCollectionId: string | null;
 };
 
-function NewEditGalleryPage({ galleryId }: Props) {
+function NewEditGalleryPage({ galleryId, initialCollectionId }: Props) {
```
Comment (tbezman) [inline pages/gallery/[galleryId]/edit.tsx:105]: threading through the `initialCollectionId` prop

src/scenes/UserGalleryPage/UserGalleryCollection.tsx
```diff
@@ -148,10 +152,17 @@ function UserGalleryCollection({
...
+                      href={
+                        isMultiGalleryEnabled
+                          ? {
+                              pathname: '/gallery/[galleryId]/edit',
+                              query: { galleryId, collectionId },
+                            }
+                          : {
+                              pathname: '/gallery/[galleryId]/collection/[collectionId]/edit',
+                              query: { galleryId, collectionId },
+                            }
```
tbezman, line 164: pure

src/scenes/UserGalleryPage/UserGalleryCollections.tsx
```diff
@@ -157,7 +157,10 @@ function UserGalleryCollections({ galleryRef, queryRef, mobileLayout }: Props) {
                   rowRenderer={rowRenderer}
                   style={{
                     outline: 'none',
+                    overflowX: 'visible',
+                    overflowY: 'visible',
                   }}
+                  containerStyle={{ overflow: 'visible' }}
```
tbezman, line 163: this is to make sure the icon doesn't get cut off anymore
image

PR #1289, opened 2023-01-25
Moreeeee multigallery fixes

tbezman:
- Fix unsaved changes font weight 
  - image
- Fixed bug when you save a collection layout with no tokens (would fail before)
- Factor out `useSaveHotkey` since it was duped (we cannot use the `useMultiHotkey` hook since it doesn't preventDefault)
  - Still plan on deprecating this long term
- Flash "Saved" text if you hit "Cmd+S" while you have no unsaved changes

https://user-images.githubusercontent.com/6754223/214675648-3f0da313-5056-4310-b6c1-98884489a214.mov

- Create `useGuardEditorUnsavedChanges` hook which just doesn't trigger your callback until you confirm that you have unsaved changes
- Fix 404s on gallery page
- Disable refresh of chains you don't have a wallet for


pages/gallery/[galleryId]/edit.tsx
```diff
@@ -67,6 +68,8 @@ function NewEditGalleryPageInner({ queryRef }: NewEditGalleryPageInnerProps) {
     }
   }, [back, canGoBack, query.viewer, replace]);
 
+  const handleBack = useGuardEditorUnsavedChanges(goBack, hasUnsavedChanges);
```
Comment (Robinnnnn) [inline pages/gallery/[galleryId]/edit.tsx:71]: lord have mercy this is clever

Robinnnnn (approved):
🔥🔥🔥

PR #1291, opened 2023-01-26
Add Primary Wallet selector

kaitoo1:
## Description
This PR updates the Settings popover to add a way for users to set a primary wallet.

**details around behavior:**
- A primary wallet is required. A backend migration has set a default primary wallet for every user.
- A user cannot disconnect their primary wallet directly. They must first unset it as the primary, and then disconnect. A tooltip informs the user about this limitation.

**notable changes:**
- Added a new row to display the primary wallet, and updated the existing rows to only display non-primary wallets.
- Removed the logic around not letting users disconnect the wallet they signed in with. The only thing that matters is that they cannot disconnect their only wallet, which will also be their primary wallet.
- Updated some styling around the error/warning state of the button.

## Screenshots
Desktop
Screen Shot 2023-01-26 at 17 45 01

Mobile (pending design approval)
Screen Shot 2023-01-26 at 17 43 29

tooltip
Screen Shot 2023-01-26 at 17 44 20

### DIFF
--- a/src/components/ManageWallets/ManageWallets.tsx
+++ b/src/components/ManageWallets/ManageWallets.tsx
@@ -12,9 +12,10 @@ import useAddWalletModal from '~/hooks/useAddWalletModal';
 import usePersistedState from '~/hooks/usePersistedState';
 import SettingsRowDescription from '~/scenes/Modals/SettingsModal/SettingsRowDescription';
 import { removeNullValues } from '~/utils/removeNullValues';
-import { truncateAddress } from '~/utils/wallet';
+import { graphqlTruncateAddress, truncateAddress } from '~/utils/wallet';
 
 import ManageWalletsRow from './ManageWalletsRow';
+import PrimaryWalletRow from './PrimaryWalletRow';
 
 type Props = {
   newAddress?: string;
@@ -44,6 +45,13 @@ function ManageWallets({
                   chain @required(action: THROW)
                 }
               }
+              primaryWallet @required(action: THROW) {
+                dbid @required(action: THROW)
+                chainAddress @required(action: THROW) {
+                  chain @required(action: THROW)
+                  ...walletTruncateAddressFragment
+                }
+              }
             }
           }
         }
@@ -52,7 +60,14 @@ function ManageWallets({
     queryRef
   );
 
-  const wallets = useMemo(() => removeNullValues(viewer?.user?.wallets), [viewer?.user?.wallets]);
+  const primaryWallet = viewer?.user?.primaryWallet;
+  const nonPrimaryWallets = useMemo(
+    () =>
+      removeNullValues(viewer?.user?.wallets).filter(
+        (wallet) => wallet.dbid !== primaryWallet?.dbid
+      ),
+    [primaryWallet?.dbid, viewer?.user?.wallets]
+  );
 
   const { pushToast } = useToastActions();
 
@@ -65,8 +80,15 @@ function ManageWallets({
     showAddWalletModal({ onEthAddWalletSuccess, onTezosAddWalletSuccess });
   }, [onEthAddWalletSuccess, onTezosAddWalletSuccess, showAddWalletModal]);
 
-  const addWalletDisabled = useMemo(() => wallets.length >= MAX_ALLOWED_ADDRESSES, [wallets]);
+  const addWalletDisabled = useMemo(
+    () => nonPrimaryWallets.length >= MAX_ALLOWED_ADDRESSES,
+    [nonPrimaryWallets]
+  );
   const [userSigninAddress] = usePersistedState(USER_SIGNIN_ADDRESS_LOCAL_STORAGE_KEY, '');
+
+  // const primaryWallet = useMemo(() => {
+  //   return wallets.find((wallet) => wallet.dbid === viewer?.user.primaryWallet.dbid);
+  // }, [viewer?.user.primaryWallet, wallets]);
 
   useEffect(() => {
     if (removedAddress) {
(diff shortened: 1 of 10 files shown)


src/components/ManageWallets/PrimaryWalletRow.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <PrimaryWallet gap={12}>
+      <TitleDiatypeM>Primary</TitleDiatypeM>
+      <HStack align="center" justify="space-between">
+        <HStack gap={12} align="center">
+          <Icon src={iconUrl} />
+          <StyledWalletAddress>{address}</StyledWalletAddress>
+        </HStack>
+        <div {...getReferenceProps()} ref={reference}>
```
kaitoo1, line 30: I added the wrapper div around Button because the tooltip doesn't appear if the ref is on a disabled Button

src/components/core/Button/Button.tsx
```diff
@@ -18,7 +18,7 @@ const alphaHex = (percentage: number) => {
 };
 
 type StyledButtonProps = {
-  variant?: 'primary' | 'secondary' | 'error';
+  variant?: 'primary' | 'secondary' | 'warning';
```
kaitoo1, line 21: renamed the `error` variant to `warning` to reflect the usage more accurately. this variant was only being used in the wallet settings section.

src/scenes/Modals/SettingsModal/SettingsModal.tsx
```diff
@@ -240,8 +240,6 @@ function SettingsModal({
 }
 
 const StyledManageWalletsModal = styled(VStack)`
-  width: 300px;
```
kaitoo1, line 243: this was making the content artificially narrow on small mobile screens. on mobile, the popover width is dictated by the parent containers so this component doesn't have to set a width

src/components/core/Button/Button.tsx
```diff
@@ -49,7 +49,6 @@ const StyledButton = styled.button<StyledButtonProps>`
 
   .Button-label {
     opacity: 1;
-    transition: all ${transitions.cubic};
```
kaitoo1, line 52: there is already a transition defined on the parent so this was making the button-label text's transition occur after the parent transition, creating a 2-step transition effect when hovering on the button

src/components/ManageWallets/PrimaryWalletRow.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <Button variant="warning" disabled>
+            Disconnect
+          </Button>
+        </div>
+        <NewTooltip
```
kaitoo1, line 35: trying out @tbezman 's new tooltip

src/components/ManageWallets/ManageWalletsRow.tsx
```diff
@@ -20,36 +24,23 @@ type Props = {
...
-    setShowTooltip(false);
-  }, []);
+  const [isPending, setIsPending] = useState(false);
 
   const handleDisconnectClick = useCallback(async () => {
     ReactTooltip.hide();
     try {
       setErrorMessage('');
       setIsDisconnecting(true);
+      setIsPending(true);
```
tbezman, line 38: The underlying `usePromisifiedMutation` returns `[mutate, isMutating]`. Can we use the underlying flag here instead of creating our own state or is this necessary?

src/components/ManageWallets/ManageWalletsRow.tsx
```diff
@@ -20,36 +24,23 @@ type Props = {
...
-    setShowTooltip(true);
-  }, []);
-
-  const handleMouseExit = useCallback(() => {
-    setShowTooltip(false);
-  }, []);
+  const [isPending, setIsPending] = useState(false);
 
   const handleDisconnectClick = useCallback(async () => {
     ReactTooltip.hide();
```
tbezman, line 34: still need?

src/components/ManageWallets/PrimaryWalletRow.tsx
tbezman: 🤔

src/components/ManageWallets/PrimaryWalletRow.tsx
tbezman: i think we can make this API more ergo. thinking we can combine the `style` and `ref` into `getFloatingProps` somehow

src/components/ManageWallets/ManageWalletsRow.tsx
kaitoo1: ooh I didn't realize. sounds promising

kaitoo1: uhh no pun intended

src/components/ManageWallets/ManageWallets.tsx
```diff
@@ -93,13 +111,20 @@ function ManageWallets({
       <VStack gap={16}>
         <VStack>
           <SettingsRowDescription>
-            Add more wallets to access your other NFTs. You&apos;ll also be able to sign in using
-            any connected wallet.
+            Add wallets to access your pieces. You&apos;ll also be able to sign in using any
+            connected wallet.
```
Robinnnnn, line 115: nit: toss the word "also"

Robinnnnn (approved):
idk how you do it ser

src/components/ManageWallets/ManageWalletsRow.tsx
kaitoo1: worked like a charm. updated both the `useUpdatePrimaryWallet` and `useRemoveWallet` hooks so I was able to remove `useState` entirely from this file, thank you ser

src/components/ManageWallets/ManageWalletsRow.tsx
kaitoo1: good catch, removed

PR #1302, opened 2023-01-30
New Feed Pages

tbezman:
This is the PR to enable the hot feed. This part was pretty simple, we just had to swap out `globalFeed` with `trendingFeed`

To enable the `globalFeed` to appear after the `trendingFeed`, we did the following.
- Fetch the first page of both in the initial query.
- Paginate through the `trendingFeed` until `hasPrevious` is false
- Then we start appending the `globalFeed` to a joined list
- `hasMore` is now `trendingFeed.hasPrevious || globalFeed.hasPrevious`
- `loadMore` now triggers the next page based on which dataset has more results.


Robinnnnn (approved):
pulled this down locally, works like a charm

PR #1308, opened 2023-02-02
Fix double creations of collections

tbezman:
Prior to this PR, we were not resetting the collection state in the context after a save.

I've updated the logic to make sure
- Collections from the server are shown to the user after hitting save
- Deleted collectionIds don't get sent twice
- Created collections get reset after saving

Bonus:
- Clicking on another section no longer causes the editor to think you have unsaved changes.

### DIFF
--- a/src/components/GalleryEditor/getInitialCollectionsFromServer.ts
+++ b/src/components/GalleryEditor/getInitialCollectionsFromServer.ts
@@ -13,46 +13,33 @@ import { generate12DigitId } from '~/utils/generate12DigitId';
 import { removeNullValues } from '~/utils/removeNullValues';
 
 export function getInitialCollectionsFromServer(
-  queryRef: getInitialCollectionsFromServerFragment$key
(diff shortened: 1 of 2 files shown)


src/components/GalleryEditor/getInitialCollectionsFromServer.ts
```diff
@@ -13,47 +13,34 @@ import { generate12DigitId } from '~/utils/generate12DigitId';
...
-  const query = readInlineData(
+  const gallery = readInlineData(
     graphql`
-      fragment getInitialCollectionsFromServerFragment on Query @inline {
-        galleryById(id: $galleryId) {
-          __typename
-          ... on Gallery {
-            dbid
-            collections {
+      fragment getInitialCollectionsFromServerFragment on Gallery @inline {
```
tbezman, line 20: this whole block is me just making the fragment on the Gallery type instead of Query so we can reuse it in the mutation response

PR #1312, opened 2023-02-02
No unsafe indexed access

tbezman:
This PR enables the `noUncheckedIndexedAccess` compiler option in TypeScript. 

Consider the following
```ts
function add(a: number, b: number) {
    return a + b;
}

const x: Record<string, number> = {};

const sum = add(x['fake1'], x['fake2']);

console.log(sum)
```

TypeScript thinks this is fine, but you get `NaN` in the console.

### My Approach
Fix everything except for the old gallery editor stuff. I just ts-expect-error'd those

### Outcome?
I actually found a few bugs
- If you only have one collection and then delete it, another empty collection did not get created
- If you only had one section and then delete it, you have no sections with no way to create a new one.
- We weren't checking the right collection for unsaved changes when moving a collection
- I'm sure there were other spicy cases but I couldn't reason about them


src/components/GalleryEditor/getInitialCollectionsFromServer.ts
```diff
@@ -125,3 +107,24 @@ export function getInitialCollectionsFromServer(
...
+    name: '',
+    collectorsNote: '',
+    hidden: false,
+
+    sections: {
+      [generatedSectionId]: { columns: 3, items: [] },
+    },
+    activeSectionId: generatedSectionId,
+  };
+}
```
Robinnnnn, line 130: image

should we also use this func here? in `GalleryEditorContext`

Robinnnnn (approved):

PR #1323, opened 2023-02-07
Editor improvement

jakzaizzat:
TODO

- [x] if click save, dont popup caption  
- [x]  if click done  
	- [x] popup caption if there is any changes  
	- [x] back to previous state  
	- [x] same with `all galleries`  
- [x] add event tracking  
- [x] check move collection  
- [x] check onboarding flow  
- [ ] check feed

**Notes**

Required backend PR to merge first: https://github.com/gallery-so/go-gallery/pull/795

**Demo**

**New flow**

https://user-images.githubusercontent.com/4480258/217436278-24d519dc-e7eb-46f3-ae86-15418f802a64.mp4

**Silent Done**

https://user-images.githubusercontent.com/4480258/217436768-6cd8612d-4577-4693-85d9-1d60092810e2.mp4

**Different session for feed event**

https://user-images.githubusercontent.com/4480258/217437016-a5ef4963-fe3e-4902-a3c1-4d3f902d12db.mp4


apps/web/src/contexts/globalLayout/EditGalleryNavbar/EditGalleryNavbar.tsx
```diff
@@ -64,6 +69,10 @@ export function EditGalleryNavbar({
   const isMobile = useIsMobileOrMobileLargeWindowWidth();
 
   const handleAllGalleriesClick = useGuardEditorUnsavedChanges(() => {
+    if (doneAction === 'saved') {
+      onDone('');
+    }
+
```
tbezman, line 75: can we leave a comment here explaining that this auto saves the user's gallery? if that's what it does

tbezman (approved):
PR is good, I think there's some product holes here that I called out in slack. Hopefully that gets resolved

apps/web/src/contexts/globalLayout/EditGalleryNavbar/EditGalleryNavbar.tsx
Robinnnnn: +1

@tbezman when this hook is instantiated will it lock in the value of `doneAction`? in other words can we be confident that the callback will be up-to-date when `doneAction` changes

Robinnnnn (approved):
jakz, thanks for crushing this so quickly 🔥 🔥 🔥 🔥 🔥 🔥 

noticed this small issue in the video, not sure if regression

image

also not sure why there's a scroll bar that appears on the right- is it because of the placeholder? or because you have a mouse plugged in? 🤔 

image

apps/web/src/contexts/globalLayout/EditGalleryNavbar/EditGalleryNavbar.tsx
tbezman: nah it's always up to date

jakzaizzat: Got it! Added the comment

jakzaizzat:
@Robinnnnn 

1. Regarding the caption box, its because of the placeholder. I've increase the caption textarea so its not scrollable by default. 

CleanShot 2023-02-09 at 10 10 29

2. For regression I opened another PR since its happens on prod - https://github.com/gallery-so/gallery/pull/1329

PR #1334, opened 2023-02-09
Bye bye old gallery editor

tbezman:
Alright, here's what I did
- Deleted the old directory
- Updated the gallery edit page to remove the feature flag
- Used JetBrains to find all unused code (deleted it)
- Renamed `XNew` to `X` since they're not new anymore

I did some basic testing
- Edited a gallery
- Signed up as a new user

Random changes
- My node version kept getting overridden by Moon so I hard set our node version to `16.18.1` which is the version that runs in Vercel


.moon/toolchain.yml
```diff
@@ -8,6 +8,8 @@ $schema: 'https://moonrepo.dev/schemas/toolchain.json'
 # instead of relying on a version found on the host machine. This ensures deterministic
 # and reproducible builds across any machine.
 node:
+  node: "16.18.1"
```
Robinnnnn, line 11: i think this is supposed to be `version`

Robinnnnn (approved):

PR #1340, opened 2023-02-14
Yarn 1 => Yarn 3

tbezman:
This PR upgrades us to Yarn 3! This is a step in the right direction. We are on the latest yarn version with some backwards compatibility options turned on (See `.yarnrc`)

In the future there won't be a `node_modules` directory 😱. The ecosystem needs some time to catch up so we can disable these options.


.moon/toolchain.yml
```diff
@@ -25,7 +25,7 @@ node:
   aliasPackageNames: "name-and-scope"
 
   # Dedupe dependencies after the lockfile has changed.
-  dedupeOnLockfileChange: true
```
tbezman, line 28: There is a bug in yarn dedupe that doesn't understand our `nohoist` settings for yarn workspaces. 

This disabled `yarn dedupe` which isn't the worst thing in the world

apps/mobile/metro.config.js
```diff
@@ -17,6 +17,6 @@ config.resolver.nodeModulesPaths = [
   path.resolve(workspaceRoot, "node_modules"),
 ];
 // 3. Force Metro to resolve (sub)dependencies only from the `nodeModulesPaths`
-// config.resolver.disableHierarchicalLookup = false;
+config.resolver.disableHierarchicalLookup = true;
```
tbezman, line 20: This is how this was intended to be and finally works now that we have nohoist setup properly

apps/web/moon.yml
```diff
@@ -38,8 +38,8 @@ tasks:
     - concurrently
     - -n Next,Relay-Compiler,GraphQL-Codegen
     - yarn next dev
-    - yarn moon run web:relay-watch
-    - yarn moon run web:graphql-codegen-watch
+    - node_modules/.bin/moon run web:relay-watch
+    - node_modules/.bin/moon run web:graphql-codegen-watch
```
tbezman, line 42: the new yarn doesn't let you run binary files which is super annoying 

There's even a thread [here](https://github.com/yarnpkg/berry/issues/882) with Miles on it

apps/web/package.json
```diff
@@ -10,6 +10,7 @@
     "@gallery-so/visitor-plugin-common": "^0.0.2",
     "@graphql-codegen/cli": "^2.11.6",
     "@graphql-codegen/typescript": "^2.7.3",
+    "@moonrepo/cli": "^0.23.4",
```
tbezman, line 13: I want to look into a way to make sure that we don't let each app fall out of sync with it's siblings' / parent's dependencies.

Syncpack seems interesting. Will look into this some more

package.json
```diff
@@ -2,10 +2,11 @@
   "name": "gallery-monorepo",
   "private": true,
   "workspaces": [
-    "apps/*",
+    "apps/mobile",
+    "apps/web",
```
tbezman, line 6: just opted to be consistent / explicit here

Robinnnnn (approved):

PR #1344, opened 2023-02-15
Replace FadeTransitioner with framer motion, simplify global layout context

Robinnnnn:
Original PR containing context for some of the stuff that's being removed: #743 

**Handled**
- [x] replace `FadeTransitioner` with framer motion
- [x] navbar remains intact when moving between Galleries <> Followers <> Activity

**Known bugs**
- [x] when scrolled down with the navbar out of view, then clicking into a new gallery, the navbar flashes
- [x] when clicking back to the scrolled down position, the navbar flashes

**Still need to test**
- [x] Test route transitions with reproducible Suspense
- [x] Test moving in and out of editor
- [x] Test mobile
- [x] Test onboarding

**Questions**
- [x] is [GlobalNavbar](https://github.com/gallery-so/gallery/pull/1344/files#diff-48fbb5907331ce5feaebca7e1313ebc53fd838182acfd94102674d119b758c11) okay to toss?
- [x] What was the purpose of [useGlobalNavbarHeight](https://github.com/gallery-so/gallery/pull/1344/files#diff-5c091bbbe18c1ca3d5cbe3c964faea02bde5ecc2066b23d37e414172e8f2b3f8)? can we straight use 64? seems to be fine


apps/web/pages/gallery/[galleryId]/edit.tsx
```diff
@@ -103,6 +103,13 @@ function EditGalleryPageInner({ queryRef }: EditGalleryPageInnerProps) {
     throw new Error('Username does not exist');
   }
 
+  // ummm shouldn't need to do this but sometimes this component
+  // renders when on another route?! this occurs when navigating
+  // away from the editor
+  if (route !== '/gallery/[galleryId]/edit') {
+    return null;
+  }
```
Comment (tbezman) [inline apps/web/pages/gallery/[galleryId]/edit.tsx:111]: This seems a little sus to me. WIth `AnimatedPresence` every route stays mounted for a short while while the new page is animating in.

We should be able to handle that gracefully and not do stuff like this. What was the user facing issue when this stays mounted?

PR #1345, opened 2023-02-15
add sign up button next to sign in button

kaitoo1:
## Description
This PR adds a Sign Up button next to the Sign In button on the home page.
It also makes it possible to show slightly different content on the auth modal depending on whether its sign up vs sign in, like showing the Email Sign In option, and showing custom text.

## Changes
- Add a Sign Up button to home nav
- Add `authType` to `AuthModal` to display different content based on sign in vs sign up
   - don't show Magic Link option if sign up, because it's only for existing users
   - show different subheadings that are appropriate for each flow to improve user friendliness
   -`authType` doesn't affect flow logic, just the content of the modal. If an existing user connects their wallet via the "Sign Up" modal, they'll still be able to sign in (this is existing behavior)
   - Defaults to "Sign In", so other places the Auth Modal is used will say "Sign In" (such as when the user clicks the Admire button when signed out)
- On mobile, only show the Sign In button because there isn't enough horizontal space for both buttons. We also don't currently support mobile onboarding, so it wouldn't lead to a good experience.

https://user-images.githubusercontent.com/80802871/218967511-79e41442-a03c-4cac-996c-f2ead732121f.mp4

### DIFF
--- a/apps/web/src/contexts/globalLayout/GlobalNavbar/HomeNavbar/HomeNavbar.tsx
+++ b/apps/web/src/contexts/globalLayout/GlobalNavbar/HomeNavbar/HomeNavbar.tsx
@@ -6,6 +6,7 @@ import { graphql, useFragment } from 'react-relay';
 import { HStack } from '~/components/core/Spacer/Stack';
 import { useTrack } from '~/contexts/analytics/AnalyticsContext';
 import { HomeNavbarFragment$key } from '~/generated/HomeNavbarFragment.graphql';
+import { useIsMobileWindowWidth } from '~/hooks/useWindowSize';
 
 import { NavbarLink } from '../NavbarLink';
 import { ProfileDropdown } from '../ProfileDropdown/ProfileDropdown';
@@ -58,6 +59,8 @@ export function HomeNavbar({ queryRef }: Props) {
   const latestFollowingRoute: Route = { pathname: '/latest/following', query: {} };
   const featuredRoute: Route = { pathname: '/featured', query: {} };
 
+  const isMobile = useIsMobileWindowWidth();
+
   return (
     <StandardNavbarContainer>
       <NavbarLeftContent>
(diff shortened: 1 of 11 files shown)


jakzaizzat:
@kaitoo1 LGTM. Tested on both mobile/desktop and its working great!

Some minor UI stuff when connecting the wallet. Should we vertically center it or let it be?

CleanShot 2023-02-16 at 08 54 50@2x

jakzaizzat (approved):

kaitoo1:
> @kaitoo1 LGTM. Tested on both mobile/desktop and its working great!
> 
> 
> 
> Some minor UI stuff when connecting the wallet. Should we vertically center it or let it be?
> 
> 
> 
> CleanShot 2023-02-16 at 08 54 50@2x
> 
> 

oh great catch that's cus I wrapped the existing stuff in a new stack. Need to center it for sure

apps/web/src/contexts/globalLayout/GlobalNavbar/HomeNavbar/HomeNavbar.tsx
```diff
@@ -94,7 +98,14 @@ export function HomeNavbar({ queryRef }: Props) {
...
       {/* Strictly here to keep spacing consistent */}
-      <NavbarRightContent>{isLoggedIn ? null : <SignInButton />}</NavbarRightContent>
+      <NavbarRightContent>
+        {isLoggedIn ? null : (
+          <HStack gap={8} align="center">
+            <SignInButton />
+            {!isMobile && <SignUpButton />}
+          </HStack>
+        )}
+      </NavbarRightContent>
```
Robinnnnn, line 108: nit: could we leave a comment here explaining why we're disabling signup on mobile

Robinnnnn (approved):
nit: non-blocking, but could we vertically center the buttons when the email option is missing? not sure how annoying that is to do

image

apps/web/src/contexts/globalLayout/GlobalNavbar/HomeNavbar/HomeNavbar.tsx
kaitoo1: added! good call

kaitoo1:
fixed centering issues!

Screen Shot 2023-02-16 at 16 48 31

Screen Shot 2023-02-16 at 16 48 34

Robinnnnn (approved):

PR #1349, opened 2023-02-16
Mobile Relay Setup + Shared Network Layer

tbezman:
- Factor out our network layer to shared package so we can share across web / mobile
- Clean up our Github Actions
  - Web actions get prefixed with `web-`
  - Hoist up the typecheck to a single job that runs against all of our projects


apps/mobile/relay.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
```
Robinnnnn, line 1: share this + schema with web

Robinnnnn (approved):

PR #1360, opened 2023-02-21
Setup GraphQL Defer support on mobile

tbezman:
The goal of this PR is to enable support for the `@defer` directive on mobile. 

- I've completely separated this new network layer from the web one to make sure web stays stable until we've figured this out.
- I've manually tested this about 1000x locally to make sure things work
- Automatic persisted queries still work.

### Polyfills / `fetch-multipart-graphql`
React Native does not support streaming responses out of the box. To get this to work I had to add a few polyfills. You can read more about that [here](https://github.com/acostalima/react-native-polyfill-globals).

The `fetch-multipart-graphql` library does a lot of the heavy lifting for us to actualy parse the streaming bytes of the response and turn them into streaming objects which we can use to send to Relay.

### Demo
It's really hard to see but if you look closely, `b_ez_man` should appear slightly after `robin` because I set it up such that the `Terence` component requests a bunch more data.

#### Example Code
```tsx
const QueryNode = graphql`
  query AppDeferredDataQuery {
    ...AppRobinFragment @defer
    ...AppTerenceFragment @defer
  }
`;

function DeferredData() {
  const query = useLazyLoadQuery<AppDeferredDataQuery>(QueryNode, {});

  return (
    <SafeAreaView>
      <Suspense fallback={<Text>Loading Robin...</Text>}>
        <RobinUser queryRef={query} />
      </Suspense>

      <Suspense fallback={<Text>Loading Robin...</Text>}>
        <TerenceUser queryRef={query} />
      </Suspense>
    </SafeAreaView>
  );
}
```

#### Video
https://user-images.githubusercontent.com/6754223/220417024-f954662a-376f-4650-84ac-fa6397e3695a.mov

### DIFF
--- a/apps/mobile/App.tsx
+++ b/apps/mobile/App.tsx
@@ -2,16 +2,14 @@ import "expo-dev-client";
 
 import {
   graphql,
-  GraphQLTaggedNode,
   RelayEnvironmentProvider,
   useFragment,
   useLazyLoadQuery,
-  useRelayEnvironment,
 } from "react-relay";
-import { Suspense, useEffect, useState } from "react";
+import { Suspense, useState } from "react";
 import { createRelayEnvironment } from "./src/contexts/relay/RelayProvider";
 import { AppDeferredDataQuery } from "~/generated/relay/AppDeferredDataQuery.graphql";
-import { SafeAreaView, Text, View } from "react-native";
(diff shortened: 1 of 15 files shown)


Robinnnnn (approved):
this is unbelievably cool man

PR #1388, opened 2023-03-01
Left Hand Nav

kaitoo1:
## Description
This PR adds the Left Hand Nav to replace the Profile Dropdown menu in the navbar today.
Screen Shot 2023-03-02 at 0 04 57

### Structure
**Sidebar**
In the code, the Left Hand Nav is called "Sidebar".
`GlobalSidebar` is the sidebar container included in the Global Layout that contains various flavors of the sidebar such as  `StandardSidebar`. `StandardSidebar` contains all the functionality described in the spec for the Left Hand Nav. In the future, we might have other types of sidebars that are rendered instead of `StandardSidebar` inside `GlobalSidebar`.

**Sidebar Drawer**
`SidebarDrawerContext` controls the drawers that slide out of the sidebar. For this PR, the two drawers are "Notifications" and "Settings". The drawer acts very similar to the Modal component and can be opened and closed with `showDrawer` and `hideDrawer`, respectively from most places of the app. These actions take whatever component you want to display within the drawer, which allows us to re-use the components we displayed in the Notifications and Settings modal.
Screen Shot 2023-03-02 at 0 09 50

**Sidebar Icons/Buttons**
The sidebar contains clickable elements, which live in `SidebarIcon`. These can have Next Link navigation behavior or onClick handler behavior depending on the props passed. Some icons in the Left Hand Nav take the user to a different route, while others open the drawer (Notifications, Settings). The icon component also takes props that populate the tooltip.

Screen Shot 2023-03-02 at 0 07 35

### Responsiveness
The same sidebar component is displayed for both desktop and mobile. On Desktop, it is rendered on the left side of the screen and the drawers slide out to the right. On mobile, it is at the bottom and the drawers slide up.
**Position**
The sidebar should always be visible and fixed on both desktop and mobile, as confirmed with Design.

### Misc
Added the `useDetectOutsideClick` hook that triggers a callback if a click is detected outside the ref element.

Added all the icons we have to the components page so that it's easy to see which icons we already have in the repo.
Screen Shot 2023-03-01 at 23 38 38

## Test Cases
There's a lot of logic that was added or moved from the existing Nav. Below is a list of some notable cases and behaviors we want to ensure are present.

- [x] sidebar is added to all relevant routes
- [x] blue dot appears over notifications icon if there are unread items
- [x] blue notification dot disappears if notifications are read
- [x] sidebar icon has an active state (dark outline) if the drawer is open
- [x] tooltips on sidebar icons are not visible on mobile
- [ ] scrolling within the sidebar drawer doesn't scroll the page (this works if the drawer itself is scrollable - see known issues below)
- [x] drawer can be closed by clicking on the X or outside the drawer
- [x] opening another drawer while a drawer is already open doesn't trigger an animation. it just replaces the content
- [x] all existing places where we trigger the Settings Modal now trigger the Settings sidebar drawer
- [x] gallery.so/settings opens the Settings sidebar
- [ ] clicking edit Twitter handle on user profile opens the Settings sidebar
- [ ] x-browser compatibility - safari and firefox (haven't been able to sign in on safari and firefox to test yet)
- [x] pages without the sidebar look normal

### Known Issues
- [x] The drawers reset when they're closed, so if you load more notifications and close+re-open the drawer, you'll see the initial list of notifications. I think this is acceptable for now.
- [ ] same issue as above when clicking into "X and Y users viewed your gallery" notification, then going back to the Notifications view. It resets and you lose your scroll place
- [ ] If you attempt to scroll inside a drawer that is not scrollable because the content fits on the screen, this scrolls the whole page. This is because the `overscroll-behavior` css property only prevents scrolling outside the element if the element itself is scrollable


apps/web/src/contexts/globalLayout/GlobalNavbar/CollectionNavbar/CollectionNavbar.tsx
```diff
@@ -91,30 +87,27 @@ export function CollectionNavbar({ queryRef, username, collectionId }: Collectio
...
-                <SlashText>/</SlashText>
-
-                <CollectionNameText title={unescapedCollectionName}>
-                  {unescapedCollectionName}
-                </CollectionNameText>
-              </RightContentWrapper>
-            )
-          }
-        />
+        <Wrapper gap={4} align="center">
```
kaitoo1, line 90: The change here is removing `ProfileDropdown`, so we can now directly render what used to be `rightContent`

apps/web/src/contexts/globalLayout/GlobalNavbar/CollectionNavbar/CollectionNavbar.tsx
```diff
@@ -157,3 +155,10 @@ const MobileUsernameText = styled(Paragraph)`
 
   color: ${colors.shadow};
 `;
+
+const SlashText = styled(Paragraph)`
```
kaitoo1, line 159: moved `SlashText` from another file I deleted, bc this is the only place it's used now

apps/web/src/icons/CogIcon.tsx
```diff
@@ -1,22 +1,15 @@
...
-          stroke="black"
-          strokeMiterlimit="10"
-        />
-      </g>
-      <defs>
-        <clipPath id="clip0_997_14860">
-          <rect width="16" height="16" fill="white" />
-        </clipPath>
-      </defs>
+    <svg width="25" height="24" viewBox="0 0 25 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kaitoo1, line 3: the old cog icon was too small. I confirmed we aren't using it anywhere else, so there should be no regressions

apps/web/src/icons/index.tsx
```diff
@@ -0,0 +1,82 @@
+import AlertIcon from './AlertIcon';
```
kaitoo1, line 1: pls advise if there's a way to automate this/ import+export all the icons from the directory at once

apps/web/pages/_/components.tsx
```diff
@@ -2,7 +2,10 @@ import React, { useState } from 'react';
 import styled from 'styled-components';
```
kaitoo1, line 2: this is the page that displays all icons on the same page for dev use

apps/web/src/components/Notifications/NotificationUserList.tsx
```diff
@@ -1,17 +1,18 @@
...
+import { NotificationUserListTitle } from '~/components/Notifications/NotificationUserListTitle';
 import { BackButton } from '~/contexts/globalLayout/GlobalNavbar/BackButton';
-import { useModalActions } from '~/contexts/modal/ModalContext';
+import { useDrawerActions } from '~/contexts/globalLayout/GlobalSidebar/SidebarDrawerContext';
 import { NotificationUserListModalQuery } from '~/generated/NotificationUserListModalQuery.graphql';
 
 import Loader from '../core/Loader/Loader';
+import { Notifications } from './Notifications';
 
 type NotificationUserListModalProps = {
```
kaitoo1, line 17: this is the component that shows the full list of users who viewed your gallery when you click on that notification.

main change here is to show the Notifications drawer upon clicking back instead of closing a modal.

apps/web/src/contexts/globalLayout/GlobalLayoutContext.tsx
```diff
@@ -249,40 +252,50 @@ const GlobalLayoutContextProvider = memo(({ children }: Props) => {
...
-                delay: NAVIGATION_TRANSITION_TIME_SECONDS,
-                duration: FADE_TRANSITION_TIME_SECONDS,
-              }}
-            >
-              <Suspense fallback={<FullPageLoader />}>{children}</Suspense>
-            </motion.div>
-          </AnimatePresence>
-        </GlobalLayoutActionsContext.Provider>
-      </GlobalLayoutStateContext.Provider>
+      <MainContentWrapper isSidebarPresent={isSidebarPresent}>
```
kaitoo1, line 260: the only change here is wrapping everything in `MainContentWrapper` which just adds extra whitespace on the left of the content to account for when the sidebar is present

apps/web/src/contexts/globalLayout/GlobalSidebar/AnimatedSidebarDrawer.tsx
```diff
@@ -0,0 +1,157 @@
+import { motion } from 'framer-motion';
```
kaitoo1, line 1: This is the sidebar drawer's equivalent of AnimatedModal

apps/web/src/components/Notifications/NotificationList.tsx
```diff
@@ -5,15 +5,15 @@ import styled from 'styled-components';
...
+import { SeeMore } from '~/components/Notifications/SeeMore';
 import { NotificationListFragment$key } from '~/generated/NotificationListFragment.graphql';
 import useExperience from '~/utils/graphql/experiences/useExperience';
 
 import { Notification } from './Notification';
 import { NotificationEmailAlert } from './NotificationEmailAlert';
 import { NotificationTwitterAlert } from './NotificationTwitterAlert';
 
-export const NOTIFICATIONS_PER_PAGE = 10;
+export const NOTIFICATIONS_PER_PAGE = 15;
```
kaitoo1, line 16: this fills the Notifications container better because it's full vh now

apps/web/src/contexts/globalLayout/GlobalSidebar/GlobalSidebar.tsx
```diff
@@ -1,17 +1,74 @@
...
+  const drawerState = useDrawerState();
+
+  if (!content) {
+    return null;
+  }
+
+  return (
+    <StyledGlobalSidebar>
+      <StyledGlobalSidebarContent>{content}</StyledGlobalSidebarContent>
+      <AnimatePresence>
```
tbezman, line 30: Any shot we can make this feel better? Everything is feeling so nice and this just throws my brain way off

[Screen Recording 2023-03-02 at 10.00.01 AM.mov](https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/icIMHMJMNkXxGoH3BEkt/c541fb98-137e-42b7-a6df-bec25fb64aae/Screen%20Recording%202023-03-02%20at%2010.00.01%20AM.mov)

tbezman: I guess ideally they would transition at the same time. Right now it feels like two separate things are happening

apps/web/pages/[username]/[collectionId]/index.tsx
```diff
@@ -66,6 +68,7 @@ export default function CollectionGallery({ collectionId, username }: Collection
           <CollectionGalleryPage queryRef={query} username={username} />
         </>
       }
+      sidebar={<StandardSidebar queryRef={query} />}
```
Comment (tbezman) [inline apps/web/pages/[username]/[collectionId]/index.tsx:71]: I really appreciate the consistency in API. Thanks @kaitoo1 ❤️

apps/web/pages/_/components.tsx
```diff
@@ -82,18 +85,40 @@ export default function DesignPage() {
...
+            const Icon = icons[iconKey];
+            return (
+              <IconContainer key={iconKey} align="center" justify="center" gap={8}>
+                <Icon />
+                <BaseM color={colors.shadow}>{iconKey}</BaseM>
+              </IconContainer>
+            );
+          })}
+        </Examples>
+      </Section>
```
tbezman, line 102: ❤️

apps/web/src/components/Notifications/Notification.tsx
```diff
@@ -108,16 +107,16 @@ export function Notification({ notificationRef, queryRef }: NotificationProps) {
...
-      showModal({
+      showDrawer({
         content: (
           <NotificationUserListModal notificationId={notification.id} fullscreen={isMobile} />
         ),
-        isFullPage: isMobile,
-        hideClose: true,
-        isPaddingDisabled: true,
-        headerVariant: 'standard',
+        drawerName: 'notificationsUserList',
```
tbezman, line 118: Is there anyway we can make this type safe?

apps/web/src/components/Notifications/Notification.tsx
```diff
@@ -108,16 +107,16 @@ export function Notification({ notificationRef, queryRef }: NotificationProps) {
...
 
+  const { showDrawer } = useDrawerActions();
+
   const handleNotificationClick = useMemo((): NotificationClick => {
     function showUserListModal() {
-      showModal({
+      showDrawer({
         content: (
           <NotificationUserListModal notificationId={notification.id} fullscreen={isMobile} />
         ),
```
tbezman, line 117: It feels a little bit strange to have both a `name` and a `content` field. In my head it would be on or the other, like maybe you can infer `content` from the `drawerName` but I'm not sure. 

Are there cases where you show `notificationUserList` with different content?

apps/web/src/components/Notifications/NotificationUserList.tsx
```diff
@@ -37,7 +38,15 @@ export function NotificationUserListModal({
...
-  const { hideModal } = useModalActions();
+  const { showDrawer } = useDrawerActions();
+
+  const handleBackClick = useCallback(() => {
+    showDrawer({
+      content: <Notifications />,
+      headerText: 'Notifications',
+      drawerName: 'notifications',
+    });
+  }, [showDrawer]);
```
tbezman, line 49: Love this

apps/web/src/components/Notifications/useClearNotifications.tsx
```diff
@@ -23,11 +23,7 @@ export function useClearNotifications() {
         const connectionIds = [
           ConnectionHandler.getConnectionID(
             `client:Viewer:${userId}`,
-            'ProfileDropdownFragment_notifications'
-          ),
-          ConnectionHandler.getConnectionID(
-            `client:Viewer:${userId}`,
-            'ProfileDropdownContentFragment_notifications'
+            'StandardSidebarFragment_notifications'
```
tbezman, line 26: Good find, not easy to find this shit, I wish it was type safe some how

apps/web/src/contexts/globalLayout/GlobalNavbar/CollectionNavbar/CollectionNavbar.tsx
```diff
@@ -91,30 +87,27 @@ export function CollectionNavbar({ queryRef, username, collectionId }: Collectio
...
-                <CollectionNameText title={unescapedCollectionName}>
-                  {unescapedCollectionName}
-                </CollectionNameText>
-              </RightContentWrapper>
-            )
-          }
-        />
+        <Wrapper gap={4} align="center">
+          {isMobile ? null : (
+            <RightContentWrapper shrink gap={4}>
```
tbezman, line 92: I think we've got a little too much `Wrapper` business going on here. Maybe we can consolidate. 

We should at least rename `RightContentWrapper` since that's definitely confusing.

apps/web/src/contexts/globalLayout/GlobalSidebar/AnimatedSidebarDrawer.tsx
```diff
@@ -0,0 +1,157 @@
...
+    hideDrawer();
+  }, [hideDrawer]);
+
+  const handleDoneClick = useCallback(() => {
+    hideDrawer();
+  }, [hideDrawer]);
+
+  const isMobile = useIsMobileWindowWidth();
+
+  const motionSettings = useMemo(() => {
```
tbezman, line 58: nit: declaring a return type on the function inside the useMemo would get us better autocomplete inside the function

apps/web/src/contexts/globalLayout/GlobalSidebar/SidebarDrawerContext.tsx
```diff
@@ -0,0 +1,83 @@
...
+  headerText?: string;
+  drawerName: string;
+  showDoneFooter?: boolean;
+};
+
+type DrawerState = {
+  activeDrawer: ActiveDrawer | null;
+};
+
+const DrawerStateContext = createContext<DrawerState | undefined>(undefined);
```
tbezman, line 22: Should we have a discussion about splitting up all of our contexts into two? I don't think it's worth it and just adds extra complexity. I really don't think we'll take a perf hit if we don't do it.

apps/web/src/contexts/globalLayout/GlobalSidebar/SidebarDrawerContext.tsx
```diff
@@ -0,0 +1,83 @@
+import {
```
tbezman, line 1: Any reason to keep this separate from `GlobalLayoutContext`? Seems like they're related enough to just shove the state in there

apps/web/src/icons/index.tsx
tbezman: You could try this https://webpack.js.org/guides/dependency-management/#requirecontext but I haven't used it 🤷

apps/web/pages/_/components.tsx
jakzaizzat: This is so helpful 😍

apps/web/src/contexts/globalLayout/GlobalSidebar/SidebarDrawerContext.tsx
kaitoo1: will defer to you/ @Robinnnnn  @jakzaizzat  on this one since it seems like a larger decision!

apps/web/package.json
```diff
@@ -36,6 +36,7 @@
     "@types/relay-runtime": "^14.1.6",
     "@types/styled-components": "5.1.9",
     "@types/uuid": "^8.3.4",
+    "@types/webpack-env": "^1.18.0",
```
kaitoo1, line 39: needed this to do require.context

apps/web/src/contexts/globalLayout/GlobalSidebar/SidebarDrawerContext.tsx
```diff
@@ -0,0 +1,101 @@
...
+    throw new Error('Attempted to use DrawerActionsContext without a provider!');
+  }
+
+  return context;
+};
+
+type Props = { children: ReactNode };
+
+function isDrawerType(type: string): type is DrawerType {
+  console.log({ type });
```
kaitoo1, line 52: temporarily logging some stuff to debug why type guard returns false on preview deploy

apps/web/src/components/Notifications/Notification.tsx
tbezman: This is so much better. One last thing I'll say is using the `type` for the header text seems like it might break down quickly, but I'm fine with it for now.

apps/web/src/components/Notifications/Notifications.tsx
```diff
@@ -0,0 +1,115 @@
...
+            exit={{
+              opacity: 0,
+              x: subView
+                ? ANIMATED_COMPONENT_TRANSLATION_PIXELS_SMALL
+                : -ANIMATED_COMPONENT_TRANSLATION_PIXELS_SMALL,
+            }}
+            transition={{
+              ease: rawTransitions.cubicValues,
+              duration: FADE_TRANSITION_TIME_SECONDS,
+            }}
```
tbezman, line 82: This animation looks gorgeous. Good work dude

apps/web/src/components/Notifications/useClearNotifications.tsx
tbezman: Starting up a thread with some Relay folks about this

apps/web/src/contexts/globalLayout/GlobalSidebar/GlobalSidebar.tsx
tbezman: This is significantly better!!!

apps/web/src/contexts/globalLayout/GlobalSidebar/SidebarDrawerContext.tsx
tbezman: Fair enough. Let's keep it this way for now. Can revisit later.

tbezman (approved):

PR #1390, opened 2023-03-01
Polish refresh section

jakzaizzat:
**Changes**

- Add gap between two button
- Remove tooltips
- Move refresh icon to the left

**Demo**

https://user-images.githubusercontent.com/4480258/222073438-6f68cd1f-2559-43e7-aa49-74f766ea3820.mp4


Robinnnnn (approved):
danggg the gap makes the text two lines? mind if we make the text a tiny bit smaller?

afterwards down to yeet

PR #1410, opened 2023-03-06
Fix empty space on mobile if no bio

jakzaizzat:
**Notes**

This bug only happens on **mobile** and if the user don't have any **bio**

Tested to avoid any regression

- [x] mobile with no bio
- [x] mobile with bio
- [x] desktop with no bio
- [x] desktop with bio

**Before**
CleanShot 2023-03-06 at 15 49 04

**After**
CleanShot 2023-03-06 at 15 47 35

**Desktop**

CleanShot 2023-03-06 at 15 47 27
CleanShot 2023-03-06 at 15 47 18

### DIFF
--- a/apps/web/src/scenes/UserGalleryPage/UserGallery.tsx
+++ b/apps/web/src/scenes/UserGalleryPage/UserGallery.tsx
@@ -45,6 +45,8 @@ function UserGallery({ queryRef }: Props) {
               ...UserGalleryLayoutFragment
             }
 
+            bio
+
             ...UserNameAndDescriptionHeader
             ...UserTwitterSectionFragment
           }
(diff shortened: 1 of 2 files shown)


kaitoo1 (approved):

Robinnnnn (approved):

apps/web/src/scenes/UserGalleryPage/UserGallery.tsx
```diff
@@ -92,7 +94,7 @@ function UserGallery({ queryRef }: Props) {
 
   return (
     <VStack gap={isMobile ? 12 : 24}>
-      <VStack gap={12}>
+      <VStack gap={user?.bio || !isMobile ? 12 : 0}>
```
tbezman, line 97: I think a better approach here  would be to have the `UsernameAndDescriptionHeader` just return null if it has no content. That way there's no extra gap and we can keep this number constant.

jakzaizzat: big brain 🧠 . Updated the component!

PR #1417, opened 2023-03-07
GAL-2507 temp fix for baseurl

kaitoo1:
## Description
`baseUrl` isn't working as expected on prod, and causes the link copied by the Share button to have localhost instead of gallery.so

getting the baseurl from the next env vars didn't seem to be working as expected on deployed environments, so this is a temp solution to use window.origin until we figure out why


kaitoo1:
legendary

PR #1421, opened 2023-03-08
Switch to Relay's supported preloaded query pattern to fix dedupes.

tbezman:
I haven't quite figured out what was causing the duplicate queries to happen earlier but this definitely fixes it.

Preloaded Queries are Relay's solution to the problem we had before with waterfalls. I'm pretty sure my original solution was relying on a race condition to work. With Preloaded Queries, we'll definitely never run into this problem again.

The code behind this is quite annoying (it's the same as what we were doing before). I want to callout that all of this nonsense should get better with the new Next.js app directory stuff.

### What is a preloaded query?
1. Start loading query
2. Get reference to this preloaded query
3. Pass that reference to the page  / component that needs it
4. That component now does `usePreloadedQuery`.

No room for race conditions here, very transparently passing around a reference to an in flight query.

### DIFF
--- a/apps/web/tsconfig.json
+++ b/apps/web/tsconfig.json
@@ -6,78 +6,30 @@
     "emitDeclarationOnly": false,
     "noEmit": true,
     "paths": {
-      "~/abis/*": [
-        "src/abis/*"
-      ],
-      "~/assets/*": [
-        "src/assets/*"
-      ],
-      "~/components/*": [
-        "src/components/*"
-      ],
-      "~/connectors/*": [
-        "src/connectors/*"
-      ],
-      "~/constants/*": [
-        "src/constants/*"
-      ],
-      "~/contexts/*": [
-        "src/contexts/*"
-      ],
-      "~/errors/*": [
-        "src/errors/*"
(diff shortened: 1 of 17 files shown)


apps/web/tsconfig.json
```diff
@@ -6,78 +6,30 @@
     "emitDeclarationOnly": false,
     "noEmit": true,
     "paths": {
-      "~/abis/*": [
```
tbezman, line 9: I'm pretty sure Moon has stopped formatting these files, which puts prettier back in control. This has happened on a few PRs now

package.json
```diff
@@ -15,6 +15,7 @@
     "@types/react": "^18.0.0",
     "@taquito/taquito": "^14.0.0",
     "react": "18.2.0",
+    "relay-runtime": "14.1.0",
```
tbezman, line 18: Forcing a resolution here since there's a bad peer dep from graphql-codegen. 

This isn't related but thought I'd fix while I was debugging.

Robinnnnn (approved):

PR #1423, opened 2023-03-09
Show bio instead of collection # in Explore page user cards

kaitoo1:
## Description

Changes are based on discussion around replacing collection # with bio on Explore page user cards.

Bio is truncated to one line to visually balance cards with and without bios.

before
Screen Shot 2023-03-09 at 12 08 19

after

Screen Shot 2023-03-08 at 19 31 54

### DIFF
--- a/apps/web/src/components/Explore/ExploreUserCard.tsx
+++ b/apps/web/src/components/Explore/ExploreUserCard.tsx
@@ -8,6 +8,7 @@ import { ExploreUserCardFragment$key } from '~/generated/ExploreUserCardFragment
 import { useLoggedInUserId } from '~/hooks/useLoggedInUserId';
 import { useIsMobileWindowWidth } from '~/hooks/useWindowSize';
 import { removeNullValues } from '~/shared/relay/removeNullValues';
+import unescape from '~/utils/unescape';
 
 import Badge from '../Badge/Badge';
 import breakpoints from '../core/breakpoints';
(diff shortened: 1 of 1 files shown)


apps/web/src/components/Explore/ExploreUserCard.tsx
```diff
@@ -98,6 +87,15 @@ export default function ExploreUserCard({ userRef, queryRef }: Props) {
     return badges;
   }, [user.badges]);
 
+  const isMobile = useIsMobileWindowWidth();
+
+  const bioFirstLine = useMemo(() => {
+    if (!user.bio) {
```
Robinnnnn, line 93: could we make sure to use the [unescape util](https://github.com/gallery-so/gallery/blob/main/apps/web/src/scenes/UserGalleryPage/UserNameAndDescriptionHeader.tsx#L79) here?

side note, in the future would be great to auto-unescape all string fields in the relay cache automatically. @tbezman do you know a simple way to do that? we use it in ~25 locations in the codebase and i think it'd be relatively harmless to do across the board

Robinnnnn (approved):

apps/web/src/components/Explore/ExploreUserCard.tsx
kaitoo1: ooh good catch. added unescape !

tbezman: Relay provides no mechanism to do things like this globally. Thinking of a few options here.

1. Have the backend escape these markdown fields for us.
2. Instead of using a `String` type, we use a custom scalar type to make it a little bit more obvious to the engineer that it has to be escaped although I'm not entirely sure what this would look like in practice.
3. Add a custom Relay Resolver. This is like adding a field to our schema that's only available on the client. Something like `escapedBio` that would just run the function for us. I don't think people would remember to use this though.

PR #1425, opened 2023-03-09
Mobile release automation

tbezman:
It's time to automate our release process. You can read more about this PR in this notion doc https://www.notion.so/gallery-so/Mobile-Release-Management-e7579fb89bb447f5b0a8d2b7ebc39af1

I'll tag up some sections of code in this PR but otherwise I suggest you read the notion doc.


.github/workflows/mobile-release.yml
```diff
@@ -0,0 +1,65 @@
+name: mobile-release
```
tbezman, line 1: only difference between this file and `mobile-build.yml` is the `--auto-submit` flag.

apps/mobile/app.json
```diff
@@ -17,12 +16,13 @@
     },
     "assetBundlePatterns": ["**/*"],
     "ios": {
-      "buildNumber": "5",
+      "config": {
+        "usesNonExemptEncryption": false
+      },
```
tbezman, line 21: If I don't add this here, I have to manually accept something in AppStoreConnect. This is totally chill.

apps/mobile/eas.json
```diff
@@ -1,28 +1,33 @@
...
+  "submit": {
+    "development": {
+      "ios": {
+        "ascAppId": "1672735556"
       }
     },
     "production": {
       "ios": {
-        "resourceClass": "m1-medium"
+        "ascAppId": "1672735556"
```
tbezman, line 29: this is chill and not secret.

apps/mobile/index.js
```diff
@@ -1,10 +1,7 @@
 import registerRootComponent from 'expo/build/launch/registerRootComponent';
```
tbezman, line 1: The `index` file has to be `.js` or Android builds don't work. This doesn't super bother me but is weird.

Robinnnnn (approved):
🚀🚀🚀

PR #1426, opened 2023-03-10
Show shared followers + communities on user profiles

kaitoo1:
## Description

This PR adds a section to the user profile that displays the following
- "Also owns XYZ": Communities that you both own pieces from
- "Followed by XYZ": Their followers that you also follow

## Details

Logic for displayed copy
- [x] If one shared communities, show "Also holds Community1"
Screen Shot 2023-03-10 at 19 24 27

- [x] If two shared communitiess, show "Also holds Community1 and Community2"
Screen Shot 2023-03-10 at 19 24 35

- [x] If three shared communitiess, show "Also holds Community1, Community2 and Community3"
Screen Shot 2023-03-10 at 19 24 40

- [x] If more than three shared communitiess, show "Also holds Community1, Community2, and [x] more"
Screen Shot 2023-03-10 at 19 24 45

- [x] Same logic applies for Shared Followers

Misc

- [x] If not signed in, don't show anything
- [x] if there are 0 mutual followers or communities, don't show that category
- [x] Popover works on mobile
- [x] Handles cases where long names might cause the line to wrap
Screen Shot 2023-03-10 at 19 29 09

- [x] for communities, the link to the community page should work for any chain
- [x] for communities that don't have a community page, don't link


apps/web/src/scenes/UserGalleryPage/UserSharedCommunities.tsx
```diff
@@ -0,0 +1,127 @@
+import { useCallback, useMemo } from 'react';
```
kaitoo1, line 1: a lot of the logic is duplicated from `UserSharedCommunities`

Ideally we can share some pieces, but I found that because the data type being displayed is different (user vs community) and there are enough small differences between the two, making the logic shareable made it more difficult to follow and didn't provide a clear benefit. 

Since we only have 2 versions, I think this is acceptable. maybe if we add a 3rd similar category we can reconsider consolidating the shared logic.

apps/web/src/components/WalletSelector/multichain/WalletSelectorWrapper.tsx
```diff
@@ -5,7 +5,6 @@ import { VStack } from '~/components/core/Spacer/Stack';
 
 export const WalletSelectorWrapper = styled(VStack)`
   text-align: center;
-  width: 320px;
```
kaitoo1, line 8: adjusted the modal padding in this PR, and this was no longer necessary

apps/web/src/utils/getCommunityUrlForToken.ts
```diff
@@ -12,6 +12,20 @@ export const DISABLED_CONTRACTS = [
   '0xdfde78d2baec499fe18f2be74b6c287eed9511d7', // Braindrops
 ];
 
+export function getUrlForCommunity(contractAddress: string, chain: string): Route | null {
```
kaitoo1, line 15: made this piece of `getCommunityUrlForToken` reusable elsewhere

apps/web/src/scenes/UserGalleryPage/UserSharedCommunities.tsx
kaitoo1: that being said if anyone sees any obvious things that could be reused pls lmk!

tbezman: totally chill, way easier to understand this way when they're just explicitly split up

apps/web/src/scenes/Modals/PaginatedList/PaginatedUsersList.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+              username
+              bio
+            }
+          }
+        }
+      }
+    `,
+    queryRef
+  );
```
tbezman, line 39: This file is in the `src/scenes/Modals` which makes me think you'd want to re-use this in a few places. However, we're fetching "sharedFollowers", which seems awfully specific.

We already have `FollowListUsers`, maybe that could be adapted. I feel like we could do some consolidation there. 

Takeaway here is, we should either figure out how to consolidate these, or just rename this component since it's not generic.

apps/web/src/scenes/UserGalleryPage/UserSharedFollowers.tsx
```diff
@@ -0,0 +1,130 @@
...
+    // But if there are exactly 3 shared followers, we display all 3 usernames. ie "username1, username2 and username3"
+    const maxUsernamesToDisplay = totalSharedFollowers === 3 ? 3 : 2;
+    return sharedFollowers.slice(0, maxUsernamesToDisplay);
+  }, [sharedFollowers, totalSharedFollowers]);
+
+  const content = useMemo(() => {
+    // Display up to 3 usernames
+    const result = followersToDisplay.map(
+      (user) =>
+        user && (
```
tbezman, line 78: I feel like some of this code might be more logically correct if we `removeNullValues` before we `slice` above. That way we know for sure that we're actually displaying the number of users we think we are. 

I would add it right before the slice in the `sharedFollowers` `useMemo` bit. Then, all of your downstream usages shouldn't have to deal with the null values.

apps/web/src/scenes/UserGalleryPage/UserSharedCommunities.tsx
```diff
@@ -0,0 +1,136 @@
...
+        ...PaginatedCommunitiesListFragment
+      }
+    `,
+    queryRef
+  );
+
+  const sharedCommunities = useMemo(
+    () => query.sharedCommunities?.edges?.map((edge) => edge?.node) ?? [],
+    [query.sharedCommunities?.edges]
+  );
```
tbezman, line 54: same thing here with the removeNullValues

apps/web/src/contexts/modal/AnimatedModal.tsx
```diff
@@ -84,16 +84,10 @@ function AnimatedModal({
   return (
     <_ToggleFade isActive={isActive}>
       <Overlay onClick={hideModal} />
-      <StyledContentContainer isFullPage={isFullPage}>
+      <StyledModal isFullPage={isFullPage}>
```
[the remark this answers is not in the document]
tbezman: Nice to see this, do you think this broke other modals? The surface area here is quite large.

apps/web/src/scenes/UserGalleryPage/UserSharedFollowers.tsx
```diff
@@ -0,0 +1,130 @@
...
+type Props = {
+  queryRef: UserSharedFollowersFragment$key;
+};
+
+export default function UserSharedFollowers({ queryRef }: Props) {
+  const query = useFragment(
+    graphql`
+      fragment UserSharedFollowersFragment on GalleryUser {
+        __typename
+        sharedFollowers(first: $sharedFollowersFirst, after: $sharedFollowersAfter)
```
jarrel-b, line 24: Think this should be `last` and `before` instead

apps/web/src/scenes/UserGalleryPage/UserSharedCommunities.tsx
```diff
@@ -0,0 +1,136 @@
...
+  queryRef: UserSharedCommunitiesFragment$key;
+};
+
+export const COMMUNITIES_PER_PAGE = 20;
+export default function UserSharedCommunities({ queryRef }: Props) {
+  const query = useFragment(
+    graphql`
+      fragment UserSharedCommunitiesFragment on GalleryUser {
+        __typename
+        sharedCommunities(first: $sharedCommunitiesFirst, after: $sharedCommunitiesAfter)
```
jarrel-b, line 26: Think this should be `last` and `before` instead

apps/web/src/scenes/UserGalleryPage/UserSharedFollowers.tsx
kaitoo1: good point - I added `removeNullValues` to the `sharedCommunities` memo

kaitoo1: good point - I added removeNullValues to the sharedCommunities memo

apps/web/src/scenes/UserGalleryPage/UserSharedCommunities.tsx
kaitoo1: 👍 added removeNullValues to the sharedCommunities memo here too

apps/web/src/scenes/UserGalleryPage/UserSharedCommunities.tsx
kaitoo1: commenting for posterity: chatted with Jarrel and we're gonna reverse the order of the data on the backend to avoid having to paginate backwards and display the data in reverse order

so we can keep `first` and `after` here

apps/web/src/scenes/Modals/PaginatedList/PaginatedUsersList.tsx
kaitoo1: moved these files out of `/Modals` and renamed named them to be specific to the Shared Info piece of User Profiles

PR #1449, opened 2023-03-21
Fix galleries position

jakzaizzat:
**Changes**
- Fix regression on full page component. 
- Affected `galleries` page

**Before**

CleanShot 2023-03-21 at 15 40 31

**After**

CleanShot 2023-03-21 at 15 40 05


tbezman:
Closing this for a less hacky fix here https://github.com/gallery-so/gallery/pull/1450

PR #1456, opened 2023-03-27
Feed Speed Improvements + Infinite Scroll

tbezman:
Just did some work to optimize the loading time of the feed screen.

THIS IS WITH SIMULATED 3G Network conditions, things should be way faster on 4G / 5G / Wifi

**New (Left), Old (Right)**

1. Only load the first 3 feed events
2. Immediately start loading the next X events so the user can keep scrolling
3. Add fallbacks for images so the screen doesn't feel broken while things are loading
4. Add fallbacks for the whole screen when the GraphQL request is in flight so the user can see that things are loading
  i. We can play with this to make it feel nicer but this is a huge upgrade in UX IMO

https://user-images.githubusercontent.com/6754223/228025302-b422696b-9d2d-408c-abfe-ea2857c6419a.mp4

https://user-images.githubusercontent.com/6754223/228025304-9d608a7d-392d-41df-820e-b010c866b048.mp4


Robinnnnn (approved):
so gorgeous

PR #1473, opened 2023-04-06
NftPreview Popup Menu

tbezman:
### Demo

**Note**: In the demo, pressing the menu items does not work. I fixed that right after I took the recording.

https://user-images.githubusercontent.com/6754223/230241734-3334763c-0a9f-4741-ad69-5facef4bf745.mp4

### Summary
This actually ended up being pretty simple since it's supported natively on iOS. We just won't support this on Android until there is native support. If the platform does not support this, it just doesn't work and the preview renders normally.

### Refactor
I refactored the `NftPreview` into two components as recommended by @Robinnnnn . Now `NftPreview` is a preview + a long press handler. and `NftPreviewAsset` is just the raw display of the asset.


apps/mobile/src/screens/NftDetailScreen/NftDetailAsset/NftDetailAssetImage.tsx
```diff
@@ -12,7 +12,6 @@ type Props = {
 
 export function NftDetailAssetImage({ imageUrl, outputDimensions, onLoad }: Props) {
   if (imageUrl.includes('svg')) {
-    console.log(imageUrl);
```
tbezman, line 15: Embarrassing. I thought I removed this in the other PR

Robinnnnn: lint rule? we can eslint-ignore when needed

Robinnnnn (approved):
this looks so freaking sexy. any way we can reduce the time for the context menu to pop up by like 100ms?

apps/mobile/src/components/NftPreview/NftPreviewContextMenuPopup.tsx
```diff
@@ -0,0 +1,192 @@
...
+                      font={{ family: 'ABCDiatype', weight: 'Medium' }}
+                    >
+                      CREATED BY
+                    </Typography>
+
+                    <Typography
+                      className="text-shadow text-sm"
+                      font={{ family: 'ABCDiatype', weight: 'Regular' }}
+                    >
+                      {token.owner.username}
```
kaitoo1, line 162: im realizing CREATED BY doesn't fully make sense to have yet before we launch Creator Support. The idea would be to show the artist name here.


And we probably don't need to show the owner username here either because we already know the owner from the context of the feed event or gallery page, and it's also available on the detail page.

So, I'm thinking maybe we can simplify the metadata display here to only show the title and community name.

This way we can keep the preview super clean and intentional until we have a better reason to display additional fields, like when we have Creator Support and can display the artist name.
Screenshot 2023-04-06 at 16 13 17

kaitoo1: As a bonus this should also help resolve some other edge cases I'm seeing where the community name is long and doesn't fit.
the updated design above will create more space to display the community name. We can cap at 1 line and truncate with ellipses for anything longer than that.

Screenshot 2023-04-06 at 15 48 27

kaitoo1 (commented):
unbelievably excited for this. going to change the viewing experience for sure.

pulled it down and ran it locally to look for edge cases and potential ux improvements, so I left a few comments regarding those.

good shit!!

tbezman:
@Robinnnnn 

> this looks so freaking sexy. any way we can reduce the time for the context menu to pop up by like 100ms?

I looked into it and we cannot change the animation timing. 

I added an eslint rule to make sure we don't leave dangling console logs.

PR #1480, opened 2023-04-09
[GAL-2688] Mobile Notifications

jakzaizzat:
**Changes**

Added notification screen. For now, its only can view all the notification and mark it as read. Will continue on the inner notification page whenever we already have a design for it. 

**Demo**

https://user-images.githubusercontent.com/4480258/231036840-7a67675f-3d85-4153-a0f3-7aaf4c6d5bf2.mp4

**Notification fallback**

https://user-images.githubusercontent.com/4480258/231036544-23961491-077b-4fb1-89c5-4af37bd9c7bb.mp4


apps/mobile/src/navigation/MainTabNavigator/MainTabNavigator.tsx
```diff
@@ -11,14 +11,18 @@ import { HomeScreen } from '../../screens/HomeScreen/HomeScreen';
...
+
+export function MainTabNavigator({ hasUnreadNotifications }: MainTabNavigatorProps) {
   const { top } = useSafeAreaInsets();
 
   return (
     <Tab.Navigator
       tabBarPosition="bottom"
       initialRouteName="Home"
-      tabBar={TabBar}
+      tabBar={(props) => <TabBar {...props} hasUnreadNotifications={hasUnreadNotifications} />}
```
tbezman: We should try to avoid this pattern https://reactnavigation.org/docs/troubleshooting/#screens-are-unmountingremounting-during-navigation

This is quite annoying since the approach w/ Relay is to pass fragment refs down as props and we don't have a way to pass props to the component. I think the right thing to do here is to have the leaf component (NotificationTabItem) use a lazy loaded query.

While the query is loading, we can show a notification icon w/o the blue dot, and then when it loads, the user may or may not see the blue dot based on whether or not they have unread notifications.

tbezman: I pushed up a change for this since I had a very specific idea in mind

PR #1523, opened 2023-05-04
Mobile Mixpanel setup

tbezman:
Same pattern as the `ErrorReportingContext`. Just import `useTrack` and you're golden


apps/mobile/src/contexts/MobileAnalyticsProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+import AnalyticsProvider, {
+  IdentifyFunction,
+  TrackFunction,
+} from '~/shared/contexts/AnalyticsContext';
+
+let instance: Mixpanel | undefined = undefined;
+
+// TODO env
+const token = '39358d4e4b8aeb0ee10e4ddbcb7a53eb';
+const apiUrl = 'https://analytics.gallery.so';
```
Robinnnnn, line 13: did you mean to handle this TODO in a later PR?

Robinnnnn (approved):
love the shared context!

kaitoo1:
BIG

apps/mobile/src/contexts/MobileAnalyticsProvider.tsx
tbezman: yeah I added some more of these around the codebase. just so I have something to grep when we finally support multiple environments

PR #1524, opened 2023-05-04
Dark Mode Splash Screen & Fix white flash

tbezman:
I could not figure out why the white flash was happening. A good workaround for now is to have a proper background view with the theme color.


kaitoo1:
yessssss big QOL improvement

kaitoo1 (approved):

kaitoo1:
clever workaround

PR #1536, opened 2023-05-09
Fix bug w/ login having stale data

tbezman:
Relay being too smart here. It knows there's a `null` viewer in the store from the first load (since we're not logged in). After we log in, it's like "Right the viewer is null so let's use that until the network request has actually finished". 

We DO NOT want that to happen. We want full network only for the top level screens that use viewer data.

### DIFF
--- a/apps/mobile/src/screens/Login/EnterEmailScreen.tsx
+++ b/apps/mobile/src/screens/Login/EnterEmailScreen.tsx
@@ -91,6 +91,7 @@ export function EnterEmailScreen() {
             )}
 
             <FadedInput
+              autoFocus
               placeholder="Email"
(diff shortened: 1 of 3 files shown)


apps/mobile/src/screens/Login/EnterEmailScreen.tsx
```diff
@@ -91,6 +91,7 @@ export function EnterEmailScreen() {
             )}
 
             <FadedInput
+              autoFocus
```
Robinnnnn, line 94: noice

Robinnnnn (approved):
great catch 🔥

Robinnnnn (approved):
great catch 🔥

PR #1538, opened 2023-05-09
[GAL-2874] Mobile tracking

jakzaizzat:
**Changes**

- add tracking to login flow and bottom navigation
- create wrapped `TouchableOpacity` component with tracking - `GalleryTouchableOpacity`
- track page view


apps/mobile/src/contexts/MobileAnalyticsProvider.tsx
```diff
@@ -19,7 +19,7 @@ if (token && apiUrl) {
 
 const track: TrackFunction = (eventName, eventProps, userId) => {
   try {
-    instance?.track(eventName, { userId, ...eventProps });
+    instance?.track(eventName, { userId, platform: 'mobile', ...eventProps });
```
tbezman, line 22: nice catch.

apps/mobile/.eslintrc.json
```diff
@@ -24,6 +29,13 @@
         // Okay to use require in these files
         "@typescript-eslint/no-var-requires": "off"
       }
+    },
+    {
+      "files": ["src/components/GalleryTouchableOpacity.tsx"],
+      "rules": {
+        // Only allow TouchableOpacity import in this file
+        "no-restricted-imports": "off"
+      }
```
tbezman, line 38: I would just inline ignore this in the file. This way we're not ignoring other restricted imports.

apps/mobile/src/navigation/RootStackNavigator.tsx
```diff
@@ -25,8 +32,20 @@ export function RootStackNavigator() {
...
+    const unsubscribe = navigationContainerRef.addListener('state', () => {
+      track('Page View', {
+        page: navigationContainerRef.getCurrentRoute()?.name,
+        params: navigationContainerRef.getCurrentRoute()?.params,
+      });
+    });
+
+    return unsubscribe;
+  }, [navigationContainerRef, track]);
+
```
tbezman, line 48: I think instead of passing the ref down, we could use `useNavigationState` for this.

```ts
const { page, params } = useNavigationState(state => {
  return {
    ... stuff here
  }
});

useEffect(()=> {
  track here
}, [page, params]);
```

tbezman: something like that, I'm pretty sure we don't have to pass a ref down though.

apps/mobile/src/navigation/MainTabNavigator/TabBar.tsx
```diff
@@ -20,25 +22,37 @@ type TabItemProps = {
   navigation: MaterialTopTabBarProps['navigation'];
 };
 
+const TAB_NAMES = {
```
tbezman, line 25: We should add a type here so we never forget to keep this updated. 

Something like this
```ts
const TAB_NAMES: { [key in keyof MainTabNavigatorParams]: string } = {...mappingsHere}
```

apps/mobile/src/components/GalleryTouchableOpacity.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useCallback } from 'react';
+import { GestureResponderEvent, TouchableOpacity, TouchableOpacityProps } from 'react-native';
+
+import { useTrack } from '~/shared/contexts/AnalyticsContext';
+
+export type GalleryTouchableOpacityProps = {
+  // unique identifier for the element
+  id?: string;
+  eventName?: string;
+  onPress?: (event: GestureResponderEvent) => void;
```
tbezman, line 11: Nit: prefer relying on the lib as the source of truth. 

```ts
  onPress?: TouchableOpacityProps['onPress'];
```

apps/mobile/src/components/Button.tsx
```diff
@@ -1,32 +1,34 @@
...
   className?: string;
-  style?: TouchableOpacityProps['style'];
   loading?: boolean;
   disabled?: boolean;
   text: string;
   icon?: ReactNode;
   variant?: Variant;
-  onPress?: TouchableOpacityProps['onPress'];
-};
+  trackProperties?: Record<string, unknown>;
```
tbezman, line 18: do we need this prop if the intersection already has `properties`? If I understand this correctly, now `Button` takes both `properties` and `trackProperties`

tbezman (commented):
a few small comments

tbezman (approved):

PR #1545, opened 2023-05-10
Dismiss announcement notifications when moving away from panel

Robinnnnn:
Currently, the only way to dismiss an announcement – and get rid of the blue dot in the sidebar – is by actively clicking on the announcement and going to its CTA.

This may be annoying for new users who see a bunch of notifications especially as our announcements grow over time.

image

### DIFF
--- a/apps/web/src/components/Announcement/AnnouncementList.tsx
+++ b/apps/web/src/components/Announcement/AnnouncementList.tsx
@@ -1,6 +1,6 @@
 import { useRouter } from 'next/router';
 import { Route } from 'nextjs-routes';
-import { useCallback } from 'react';
+import { useCallback, useEffect } from 'react';
 import { graphql, useFragment } from 'react-relay';
 import styled from 'styled-components';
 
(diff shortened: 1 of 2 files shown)


apps/web/src/components/Announcement/AnnouncementList.tsx
```diff
@@ -56,23 +56,25 @@ export default function AnnouncementList({ queryRef }: Props) {
...
+        dismissExperience(announcement.key);
       }
 
       hideDrawer();
     },
-    [hideDrawer, router, track, update]
+    [dismissExperience, hideDrawer, router, track]
   );
 
+  // TODO [GAL-3033]: come up with a system to "dismiss all announcement-type experiences"
```
Robinnnnn, line 67: https://linear.app/galleryso/issue/GAL-3033/come-up-with-a-system-to-dismiss-all-announcement-type-experiences

apps/web/src/components/Announcement/AnnouncementList.tsx
```diff
@@ -56,23 +56,25 @@ export default function AnnouncementList({ queryRef }: Props) {
...
+  useEffect(() => {
+    return () => {
+      for (const announcement of announcements) {
+        if (!announcement.experienced) {
+          dismissExperience(announcement.key);
+        }
+      }
+    };
+  }, [announcements, dismissExperience]);
```
Robinnnnn, line 76: run when navigating away

tbezman (approved):
i dig this

PR #1559, opened 2023-05-12
Fix gif playback rate bug

tbezman:
https://github.com/DylanVann/react-native-fast-image/pull/842

A bit silly but we have to patch `react-native-fast-image` for this.


Robinnnnn:
sick find

PR #1565, opened 2023-05-15
Base Dark Mode

tbezman:
Super hacky, we just export a different set of colors from `colors.js` if the user is on the base gallery and they have dark mode enabled. Other than that, I just had to update a few icons

localhost_3000_base (2)


Robinnnnn (approved):
LOOKS SICK

PR #1569, opened 2023-05-16
[GAL-3053] Multi-Env and associate commits w/ releases

tbezman:
## Environment Variables

The big idea here is to use the `zod` package as a schema for what our environment variables contain. This way, you cannot forget to add an environment variable to all environments.

`app.json` was moved to `app.config.ts` so we can dynamically put stuff in. We simply read from the appropriate `.env.x` file and throw those in.

The pipeline is `.env.x` => `app.config.ts` => `expo-constants` => `Constants.expoConfig.extra` => `env.YOUR_ENV_VAR_HERE`.

### Usage

If you want to use an env var.
```ts
import { env } from '~/env/runtime'`

console.log(env.FULLY_TYPED_ENV_VAR_HERE)
```

If you want to create a new variable

In `env.js`, add a new string to the environment schema, and then update all of the env files!

## Sentry Commit Tracking

This should happen automatically for us now that we have `setCommits: true` in the Sentry Expo config. We already have the integration setup on Sentry so we should be golden here.

**If this doesn't end up working, we can inject the commit has as a property on each report. But I want to try the paved road first.**

### Links
- https://docs.expo.dev/guides/using-sentry/#app-configuration
- https://docs.sentry.io/product/sentry-basics/integrate-frontend/configure-scms/
- https://docs.expo.dev/build-reference/variables/#built-in-environment-variables


apps/mobile/app.config.ts
```diff
@@ -0,0 +1,93 @@
...
+    return result.data;
+  } else if (result.success === false) {
+    throw new Error(`Could not validate your ${file}:\n\n${result.error.message}`);
+  }
+}
+
+const environmentVariablePath = `./env/.env.${process.env.ENV ?? 'prod'}`;
+// eslint-disable-next-line no-console
+console.log(`Loading Environment Variables from: ${environmentVariablePath}`);
+const environmentVariables = readEnvironmentFromFile(environmentVariablePath);
```
tbezman, line 22: This is the new stuff in this file.

apps/mobile/app.config.ts
```diff
@@ -0,0 +1,93 @@
...
+    },
+  },
+  web: {
+    favicon: './assets/favicon.png',
+  },
+  extra: {
+    eas: {
+      projectId: '67c952bc-7ade-408c-a95d-72c36ce9252c',
+    },
+    ...environmentVariables,
```
tbezman, line 74: also this

apps/mobile/app.config.ts
```diff
@@ -0,0 +1,93 @@
...
+        config: {
+          organization: 'usegallery',
+          project: 'gallery-mobile',
+          authToken: 'b2ea4a67f0ed4409968c4725d550df82d5187c12908d441cbf4a43da145934b1',
+        },
+      },
+    ],
+  },
+  plugins: [
+    ['sentry-expo', { setCommits: true, release: process.env.EAS_BUILD_GIT_COMMIT_HASH }],
```
tbezman, line 90: also this

apps/mobile/index.js
```diff
@@ -17,7 +17,13 @@ function filterOutAnnoyingWarnings() {
   console.warn = (...args) => {
     const joined = args.join(' ');
 
-    if (joined.includes('The provided value')) {
+    if (
+      joined.includes('The provided value') ||
+      joined.includes('Constants.manifest has been deprecated in favor of Constants.expoConfig.') ||
+      joined.includes(
+        'There was a problem sending log messages to your development environment [RangeError: Maximum call stack size exceeded (native stack depth)]'
```
tbezman, line 24: Some annoying logs coming from expo-constants. I couldn't find anything w/ a quick google.

apps/mobile/app.config.ts
```diff
@@ -0,0 +1,90 @@
...
+        file: 'sentry-expo/upload-sourcemaps',
+        config: {
+          organization: 'usegallery',
+          project: 'gallery-mobile',
+          authToken: 'b2ea4a67f0ed4409968c4725d550df82d5187c12908d441cbf4a43da145934b1',
+        },
+      },
+    ],
+  },
+  plugins: [['sentry-expo', { setCommits: true }], 'expo-barcode-scanner'],
```
tbezman, line 89: also this

apps/mobile/env/env.js
```diff
@@ -0,0 +1,14 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { z } = require('zod');
```
Robinnnnn, line 2: this is dope

Robinnnnn (approved):

PR #1581, opened 2023-05-18
[GAL-3095] Support ability to display list of followers in feed

tbezman:
Data is loaded lazily so we don't make the feed slower. But we should also make that followed list paginated.

https://github.com/gallery-so/gallery/assets/6754223/61543fa1-813b-4712-95bc-403ae2aa6fae


Robinnnnn (commented):
excellent idea to lazy load 🧠

Robinnnnn (approved):

kaitoo1 (approved):
this looks great - pretty cool that this is a one file change

PR #1583, opened 2023-05-18
Simplify tracking

Robinnnnn:
Explanations in comments

### DIFF
--- a/apps/mobile/src/contexts/MobileAnalyticsProvider.tsx
+++ b/apps/mobile/src/contexts/MobileAnalyticsProvider.tsx
@@ -17,18 +17,16 @@ if (token && apiUrl) {
   instance.init(false, {}, apiUrl);
 }
 
-const track: TrackFunction = (eventName, eventProps, userId) => {
+const track: TrackFunction = (eventName, eventProps) => {
   try {
-    instance?.track(eventName, { userId, platform: 'mobile', ...eventProps });
+    instance?.track(eventName, eventProps);
   } catch (error: unknown) {
     // Handle error here
   }
 };
 
 const identify: IdentifyFunction = (userId) => {
-  if (userId) {
-    instance?.identify(userId);
-  }
+  instance?.identify(userId);
 };
 
 export function MobileAnalyticsProvider({ children }: PropsWithChildren) {
--- a/apps/web/src/contexts/analytics/WebAnalyticsProvider.tsx
+++ b/apps/web/src/contexts/analytics/WebAnalyticsProvider.tsx
@@ -17,14 +17,11 @@ if (process.env.NEXT_PUBLIC_MIXPANEL_TOKEN && process.env.NEXT_PUBLIC_ANALYTICS_
 }
 
 // raw tracking. use this sparingly if you can't use hooks, or are outside the analytics context.
-export const _track: TrackFunction = (eventName, eventProps, userId) => {
+export const _track: TrackFunction = (eventName, eventProps) => {
   if (!mixpanelEnabled) return;
 
   try {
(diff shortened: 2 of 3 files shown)


apps/mobile/src/contexts/MobileAnalyticsProvider.tsx
```diff
@@ -17,18 +17,16 @@ if (token && apiUrl) {
   instance.init(false, {}, apiUrl);
 }
 
-const track: TrackFunction = (eventName, eventProps, userId) => {
+const track: TrackFunction = (eventName, eventProps) => {
   try {
-    instance?.track(eventName, { userId, platform: 'mobile', ...eventProps });
+    instance?.track(eventName, eventProps);
```
Robinnnnn, line 22: 1) `userId` tracking is unnecessary as the `identify` function will attach a `User ID` key to all track events out of the box. from looking at the data, `User ID` is much more reliable than our own `userId`

2) remove `platform: mobile` tracking as a more reliable piece is the `Mixpanel Library` key, which again is attached out of the box

apps/mobile/src/contexts/MobileAnalyticsProvider.tsx
```diff
@@ -17,18 +17,16 @@ if (token && apiUrl) {
...
   } catch (error: unknown) {
     // Handle error here
   }
 };
 
 const identify: IdentifyFunction = (userId) => {
-  if (userId) {
-    instance?.identify(userId);
-  }
+  instance?.identify(userId);
```
Robinnnnn, line 29: we should never call this function without `userId`

apps/web/src/contexts/analytics/WebAnalyticsProvider.tsx
```diff
@@ -17,14 +17,11 @@ if (process.env.NEXT_PUBLIC_MIXPANEL_TOKEN && process.env.NEXT_PUBLIC_ANALYTICS_
...
-export const _track: TrackFunction = (eventName, eventProps, userId) => {
+export const _track: TrackFunction = (eventName, eventProps) => {
   if (!mixpanelEnabled) return;
 
   try {
-    mixpanel.track(eventName, {
-      userId, // field will be `null` if no user ID exists
-      ...eventProps,
-    });
+    mixpanel.track(eventName, eventProps);
```
Robinnnnn, line 24: no need to manually track `userId` for reasons listed above

packages/shared/src/contexts/AnalyticsContext.tsx
```diff
@@ -32,8 +32,8 @@ const AnalyticsContextQueryNode = graphql`
   }
 `;
 
-export type TrackFunction = (eventName: string, eventProps: EventProps, userId?: string) => void;
-export type IdentifyFunction = (userId?: string) => void;
```
Robinnnnn, line 36: we'll never call this without `userId`

packages/shared/src/contexts/AnalyticsContext.tsx
```diff
@@ -66,20 +66,9 @@ const AnalyticsProvider = memo(({ children, identify, track }: Props) => {
...
-        {},
-        { fetchPolicy: 'store-or-network' }
-      )
-        .toPromise()
-        .then((query) => {
-          const userId = query?.viewer?.user?.dbid;
-
-          track(eventName, eventProps, userId);
-        });
+      track(eventName, eventProps);
```
Robinnnnn, line 69: no longer sending `userId` with track; a successful `identify` call will automatically attach `User ID`s

tbezman (approved):
luv 2 c it. thanks man

PR #1586, opened 2023-05-19
[GAL-3106] Magic Link Email Check

tbezman:
Make sure the email is valid before heading into the magic link flow.

The error copy in the video is outdated, new copy is  `We don't recognize the email you entered. Login with QR code instead, or verify your email at gallery.so/settings`

https://github.com/gallery-so/gallery/assets/6754223/873629ef-63e5-4eb1-af32-c3d1b7ce8f4a

### DIFF
--- a/apps/mobile/src/screens/Login/EnterEmailScreen.tsx
+++ b/apps/mobile/src/screens/Login/EnterEmailScreen.tsx
@@ -44,6 +44,8 @@ export function EnterEmailScreen() {
       message: string;
       underlyingError?: Error;
     }) {
+      track('Sign In Failure', { 'Sign in method': 'Email', error: message });
+
       if (underlyingError) {
         reportError(underlyingError);
       } else {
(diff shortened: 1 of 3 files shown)


Robinnnnn (approved):

apps/mobile/src/screens/Login/EnterEmailScreen.tsx
```diff
@@ -24,48 +27,81 @@ export function EnterEmailScreen() {
...
-          "Something unexpected went wrong while logigng in. We've been notified and are looking into it"
-        );
-        return;
+        return handleLoginError({ message: FALLBACK_ERROR_MESSAGE });
       }
 
       const result = await login({ magicLink: { token } });
 
       if (result.kind === 'failure') {
         track('Sign In Failure', { 'Sign in method': 'Email', error: result.message });
```
Robinnnnn, line 90: let's make sure this is called in earlier cases of failure / early returns. maybe just roll into `handleLoginError`.

tbezman: great callout

PR #1592, opened 2023-05-22
Handle missing collection on NftDetail screen more gracefully

tbezman:
Okay back story

- The web nft detail page requires the collection id since we let you slide between pieces in a collection.
- Since we have a share button on mobile that links to the web version, I thought it would make sense to thread down a `CollectionToken` type and not allow navigation if there wasn't a collection id.
- Sometimes, a feed event token's collection no longer exists, which means we couldn't navigate to the nft detail screen as it required a `collectionId`

Now we just optionally require a collectionId, and the Share button just doesn't work if there's not a collection associated w/ the token.

Tested by setting the resolver on the backend to always return `nil` for the collection and observed that the nft detail and popup menu still work as expected.


Robinnnnn (approved):
let's gooooo amazing catch

PR #1609, opened 2023-05-31
Use token processor dimensions

tbezman:
Use the width / height from the server to avoid layout shift on the client.

| Before | After |
|--------|--------|
| <video src="https://github.com/gallery-so/gallery/assets/6754223/bb9475b2-8f63-4303-ace8-a50dd433fcf7" /> | <video src="https://github.com/gallery-so/gallery/assets/6754223/5e3f41e6-a452-4003-8bd5-5dec0f04c857" /> |


apps/mobile/src/contexts/GalleryTokenDimensionCacheContext.tsx
tbezman, line 1: Much better to keep this state local to the component that needs it.

apps/mobile/src/components/ProfileView/Tabs/ProfileViewFeaturedTab.tsx
```diff
@@ -56,7 +54,9 @@ export function ProfileViewFeaturedTab({ queryRef }: ProfileViewFeaturedTabProps
   return (
     <View style={contentContainerStyle}>
       <Tabs.FlashList
-        estimatedItemSize={300}
+        getItemType={(item) => item.kind}
+        keyExtractor={(item) => item.key}
+        estimatedItemSize={100}
```
tbezman, line 59: should have the `getItemType` and `keyExtractor` on all of these. `estimatedItemSize` helps with funky mounting behavior

apps/mobile/src/components/ProfileView/Tabs/ProfileViewFeaturedTab.tsx
```diff
@@ -39,13 +35,15 @@ export function ProfileViewFeaturedTab({ queryRef }: ProfileViewFeaturedTabProps
...
+      return { items: [], stickyIndices: [] };
+    }
 
-  const { items, stickyIndices } = createVirtualizedGalleryRows({
-    galleryRef: user.featuredGallery,
-  });
+    return createVirtualizedGalleryRows({
+      galleryRef: user.featuredGallery,
+    });
+  }, [user?.featuredGallery]);
```
tbezman, line 46: had anxiety about this not being memoized.

apps/mobile/src/components/ProfileView/ProfileView.tsx
```diff
@@ -77,42 +76,40 @@ export function ProfileView({ queryRef, shouldShowBackButton }: ProfileViewProps
       </View>
 
       <View className="flex-grow">
-        <GalleryTokenDimensionCacheProvider>
```
tbezman, line 80: just removed the provider

apps/mobile/src/components/Gallery/GalleryTokenPreview.tsx
```diff
@@ -40,36 +48,50 @@ export function GalleryTokenPreview({ tokenRef, containerWidth }: GalleryTokenPr
...
       });
+    }
 
-      return result;
+    if (measuredDimensions) {
+      return fitDimensionsToContainerContain({
+        container: { width: containerWidth, height: screenDimensions.height / 2 },
+        source: measuredDimensions,
+      });
     }
```
tbezman, line 85: In this part, we're just trying to use the server dimensions as a primary source of truth, and then fallback to the measured dimensions.

kaitoo1 (approved):

PR #1612, opened 2023-06-05
[GAL-3159] Display Creator Name on NFT Preview State

jakzaizzat:
**Changes**
- Update the nft preview on web and mobile

**Web**

https://github.com/gallery-so/gallery/assets/4480258/d336d130-39cc-4848-99b9-1b29b1944fd0

**Mobile**

CleanShot 2023-06-05 at 14 13 59@2x

**Notes**

Right now, the UI to display artist is disabled. Will enabled it whenever the backend ready


apps/web/src/components/NftPreview/NftPreviewLabel.tsx
```diff
@@ -20,6 +21,8 @@ type Props = {
   interactive?: boolean;
 };
 
+const ENABLED_ARTIST = false;
```
jakzaizzat, line 24: Disabled the UI to display artist detail

apps/mobile/src/components/NftPreview/NftPreviewContextMenuPopup.tsx
```diff
@@ -27,6 +27,8 @@ type NftPreviewContextMenuPopupProps = PropsWithChildren<{
   collectionTokenRef: NftPreviewContextMenuPopupFragment$key;
 }>;
 
+const ENABLED_ARTIST = false;
```
jakzaizzat, line 30: Disabled the UI to display artist detail

kaitoo1:
nice!! looking good

just a few visual comments

- +1 to @tbezman 's comment that we don't need the gradient anymore

- also looks like the padding on the label is slightly off, top/bottom padding should be 4px
Screenshot 2023-06-05 at 16 26 58

Screenshot 2023-06-05 at 16 27 16


- looks like longer text doesn't get truncated ( should be truncated with ellipses instead of going over the edge)


- this wasn't spec'd but I think the icon/badge can be a bit smaller so that it matches the height of the text
so here the badge height is 24px but text is 20px, so it makes the banner seem a bit too tall. can we try making the badge 20px height?
Screenshot 2023-06-05 at 16 26 58

tbezman:
@jakzaizzat is this pr ready to review? I didn't see any of the container query stuff?

jakzaizzat:
@tbezman not yet. Seems like the container query doesn't work well on non chromium browser

tbezman (approved):

PR #1641, opened 2023-06-21
[GAL-3362] Double tap to admire

tbezman:
Pretty simple feature, just a lot of prop plumbing to do.

The behavior is as follows
- First tap: Queue single tap behavior to run in 200ms
- Second tap
  - If the second tap happened in that 200ms window, it's a double tap
  - If the second tap happened after that 200ms window, start over
- If there is no second tap, the single tap behavior will run w/ a 200ms delay.

Testing this on the simulator is not a great experience for some reason. Testing on my phone it feels great.

### DIFF
--- a/apps/mobile/src/components/Feed/EventTokenGrid.tsx
+++ b/apps/mobile/src/components/Feed/EventTokenGrid.tsx
@@ -14,12 +14,15 @@ type EventTokenGridProps = {
   imagePriority: Priority;
   allowPreserveAspectRatio: boolean;
   collectionTokenRefs: EventTokenGridFragment$key;
+
+  onDoubleTap: () => void;
 };
 
 export function EventTokenGrid({
   collectionTokenRefs,
   allowPreserveAspectRatio,
   imagePriority,
+  onDoubleTap,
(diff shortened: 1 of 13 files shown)


apps/mobile/src/components/Feed/EventTokenGrid.tsx
```diff
@@ -40,6 +53,59 @@ export function EventTokenGrid({
...
+  const singleTapTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const handlePress = useCallback(
+    (
+      collectionToken: EventTokenGridFragment$data[number],
+      cachedPreviewAssetUrl: string | null | undefined
+    ) => {
+      // ChatGPT says 200ms is at the fast end for double tapping.
+      // I want the single tap flow to feel fast, so I'm going for speed here.
+      // Our users better be nimble af.
+      const DOUBLE_TAP_WINDOW = 200;
```
Robinnnnn, line 83: great call 👌

Robinnnnn (approved):
this is so phenomenal

PR #1644, opened 2023-06-22
[GAL-3244] Update Admires/Comments on Feed and Display PFPs [Web]

jakzaizzat:
**Changes**
- Add new admire flow
- Add new comment flow
- Added FF to toggle it

**Demo**

https://github.com/gallery-so/gallery/assets/4480258/21ab8c69-d1aa-42af-9a6b-76ffdf6e5564


kaitoo1:
looking great!


just a heads up, I pushed a commit to use the `hasInset` prop on the pfp component - this adds the 2px padding to match the designs for the Stacked pfps
before
Screenshot 2023-06-30 at 19 17 07

after
Screenshot 2023-06-30 at 19 16 57


It looks like there's a hover cursor state for the whole section - let's remove that so only the clickable items have the hover cursor. 

https://github.com/gallery-so/gallery/assets/80802871/149a0a6b-1de1-40a0-b6e8-bb212057d065

kaitoo1:
seems like sometimes there's still an issue displaying the shorter height version of the modal 👀

Screenshot 2023-06-30 at 19 22 00

kaitoo1:
on the Admires modal, the user preview card is hidden inside the borders - would be awesome if it could show on top
Screenshot 2023-06-30 at 19 26 48

tbezman:
Echoing the admires modal thing being small

apps/web/src/components/Feed/Socialize/NotesModal/NotesModal.tsx
tbezman, line 1: Should we rename this to CommentsModal now?

apps/web/src/components/Feed/Socialize/AdmireLine.tsx
```diff
@@ -47,37 +58,70 @@ export function AdmireLine({ admireRef, eventRef, queryRef, totalAdmires }: Comm
...
     `,
     queryRef
   );
 
-  const event = useFragment(
-    graphql`
-      fragment AdmireLineEventFragment on FeedEvent {
-        dbid
-        ...NoteModalOpenerTextFragment
+  const isPfpEnabled = isFeatureEnabled(FeatureFlag.PFP, query);
```
kaitoo1, line 68: maybe worth it to `useMemo`?

apps/web/src/components/ProfilePictureStack.tsx
```diff
@@ -1,45 +1,77 @@
...
+          __typename
+        }
+        ...ProfilePictureFragment
+      }
+    `,
+    usersRef
+  );
+
+  // Sort by user that has a profile image
+  // Priority is given to the user that has a profile image
```
kaitoo1, line 35: 🧠🧠🧠🧠

kaitoo1 (approved):

apps/web/src/components/Feed/Socialize/AdmireLine.tsx
jakzaizzat: @kaitoo1 Im considering too but other places not wrapped with `useMemo` 🤔

kaitoo1:
psa: for the Admire + Comments modal, we decided it's not a dealbreaker if the modal is always a fixed size. so Jakz will timebox to find a solution (~1hr) and we don't find a solution within that time, we are good to ship with a fixed height (modal is always the same height regardless of # of comments)

apps/web/src/components/Feed/Socialize/NotesModal/NotesModal.tsx
jakzaizzat: Renamed it!

kaitoo1:
> psa: for the Admire + Comments modal, we decided it's not a dealbreaker if the modal is always a fixed size. so Jakz will timebox to find a solution (~1hr) and we don't find a solution within that time, we are good to ship with a fixed height (modal is always the same height regardless of # of comments)

update o the modal, Jakz was able to figure it out! so we have dynamic height modals

PR #1662, opened 2023-07-04
Support for more chains in editor sidebar

Robinnnnn:


apps/web/src/components/GalleryEditor/PiecesSidebar/SidebarChainDropdown.tsx
```diff
@@ -56,17 +56,12 @@ export default function SidebarChainDropdown({
 
   const availableChains = useMemo(() => {
     if (selectedView === 'Created') {
-      return chains.filter((chain) => chain.name === 'Ethereum');
+      return chains.filter((chain) => chain.hasCreatorSupport);
```
kaitoo1, line 59: moved chain-specific logic to be controlled by the chain metadata in chains.ts so we can easily see what the behavior is for each chain

PR #1673, opened 2023-07-07
fix animated modal height

kaitoo1:
before
Screenshot 2023-07-07 at 12 50 22

after
Screenshot 2023-07-07 at 12 50 15

admire/comment modals should still be ok after this change
Screenshot 2023-07-07 at 12 51 03
Screenshot 2023-07-07 at 12 49 43


jakzaizzat (approved):
LGTM! Thanks for the fix!

PR #1682, opened 2023-07-10
[GAL-3611] Post Composer [mobile]

jakzaizzat:
**Changes**

- Add Post composer screen to the mobile
- Hide it behind the feature flag `KOALA`
- Add new component for `<Toast/>`. It will have similar API as in web codebase

**Demo**

**Overall Workflow**
https://github.com/gallery-so/gallery/assets/4480258/e08e2386-0e5c-4b1c-b7c4-ca20ed2fa1a2

**Post Input Warning Message**
https://github.com/gallery-so/gallery/assets/4480258/1f51e299-9580-43be-95ac-a08fc0b2a955


apps/mobile/src/components/Toast/transition.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const rawTransitions = {
+  cubicValues,
+};
+
+const transitions = {
+  cubic,
+};
+
+export default transitions;
```
jakzaizzat, line 23: Might be worth to move it to `shared` folder if the animation for web and mobile is similar in future

apps/mobile/src/navigation/PostStackNavigator.tsx
```diff
@@ -0,0 +1,26 @@
...
+function Empty() {
+  return null;
+}
+
+export function PostStackNavigator() {
+  return (
+    <Stack.Navigator screenOptions={{ header: Empty }} initialRouteName="NftSelector">
+      <Stack.Screen
+        name="NftSelector"
+        component={ProfilePicturePickerScreen}
```
kaitoo1, line 18: let's rename `ProfilePicturePickerScreen` to be more general since we're using it for the Post flow now as well

kaitoo1: maybe `NftSelectorScreen`

kaitoo1: same for some components inside `ProfilePicturePickerScreen` like `ProfilePicturePickerFilterBottomSheet` and `ProfilePicturePickerGrid`

apps/mobile/src/screens/ProfilePicturePickerScreen/ProfilePicturePickerSingularAsset.tsx
```diff
@@ -37,44 +38,30 @@ export function ProfilePicturePickerSingularAsset({
...
+      setProfileImage(token.dbid).then(() => {
         onProfilePictureChange();
-      })
-      .catch((error) => {
-        setError(error);
-        reportError(error);
       });
-  }, [onProfilePictureChange, reportError, setProfileImage, token.dbid]);
+    } else {
+      navigation.navigate('PostComposer', {
```
kaitoo1, line 60: since we're using this `ProfilePicturePickerSingularAsset.tsx` for the Posts flow we should also rename this to be more generic like "NftSelectorSingularAsset`, along with all other references to profile pics in this file

kaitoo1: below on lines 71 and 72 there are some analytics events specific to Profile Pictures which shouldn't be sent if this component is being used for the Posts flow

kaitoo1: I did see that you isolated the pfp setting function and moved it out of this file though 🔥🔥 so renaming the other stuff will just help complete the separation

apps/mobile/src/components/Post/PostInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  return (
+    <View
+      className="relative border bg-offWhite dark:bg-black-900 border-porcelain dark:border-black-500 pb-10"
+      style={{
+        height: 117,
+      }}
+    >
+      <TextInput
```
kaitoo1, line 31: oo could you make the text input auto-focus when it's mounted? so that the keyboard is open and ready to go

apps/mobile/src/components/Post/PostInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+import colors from '~/shared/theme/colors';
+
+import { Typography } from '../Typography';
+
+export function PostInput() {
+  const [value, setValue] = useState('');
+
+  const { colorScheme } = useColorScheme();
+
+  const isTextTooLong = value.length >= 600;
```
kaitoo1, line 15: 600 could be a const to ensure it's the same value everywhere

apps/mobile/src/screens/ProfilePicturePickerScreen/ProfilePicturePickerScreen.tsx
```diff
@@ -41,6 +47,10 @@ export function ProfilePicturePickerScreen() {
   const [filter, setFilter] = useState<'Collected' | 'Created'>('Collected');
   const [networkFilter, setNetworkFilter] = useState<NetworkChoice>('all');
 
+  const screenTitle = useMemo(() => {
+    return currentScreen === 'ProfilePicture' ? 'Select a profile picture' : 'Select item to post';
```
kaitoo1, line 51: same comment about renaming from `ProfilePicturePickerScreen` : feels out of place for `ProfilePicturePickerScreen` to contain logic about showing a header saying "Select item to post"

apps/mobile/src/navigation/RootStackNavigator.tsx
```diff
@@ -53,7 +60,17 @@ export function RootStackNavigator({ navigationContainerRef }: Props) {
       initialRouteName={isLoggedIn ? 'MainTabs' : 'Login'}
     >
       <Stack.Screen name="Login" component={LoginStackNavigator} />
-      <Stack.Screen name="MainTabs" component={MainTabNavigator} />
+
+      <Stack.Screen name="Post" component={ProfilePicturePickerScreen} />
+      <Stack.Screen
+        name="ProfilePicturePickerContract"
+        component={ProfilePicturePickerContractScreen}
```
kaitoo1, line 67: question since I'm less familiar with the RootStackNavigator - how come ProfilePicturePickerContractScreen was moved from MainTabStackNavigator to here?

apps/mobile/src/navigation/types.ts
```diff
@@ -39,6 +49,9 @@ export type MainTabStackNavigatorParamList = {
 
   Search: undefined;
   Settings: undefined;
+  // Post: {
```
kaitoo1, line 52: can remove?

kaitoo1 (commented):
niiiiice Jakz!!! this looks great, feels nice to use. 

had a few comments about making the separation between Profile Picture and Post flows more defined - seeing some places where the line is a bit blurry about what a component is responsible for, which might lead to confusion in the future.

other than that LGTM, we can get a design review early next week, hook it up with backend and we should be good to go 🚀

apps/mobile/src/navigation/RootStackNavigator.tsx
jakzaizzat: The `ProfilePicturePickerContractScreen` was not moved from `MainTabStackNavigator` to `RootTabStackNavigator` but I was adding a new one. 

The reason is in post user flow, we want it to fullscreen (without any tab bar)

apps/mobile/src/components/Post/PostInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export function PostInput({ value, onChange }: Props) {
+  const { colorScheme } = useColorScheme();
+
+  const isTextTooLong = value.length >= 600;
+  const characterCount = value.length;
+
+  const handleChangeText = (newText: string) => {
+    if (newText.length <= 600) {
```
kaitoo1, line 21: I think there's a `maxlength` that we can set on the input / text area directly that will prevent the user from typing more 👀

jakzaizzat: Damnn. Nice found @kaitoo1! Didn't know it exist

apps/mobile/src/components/Post/PostInput.tsx
kaitoo1: bump @jakzaizzat

kaitoo1: oo also noticed a potential bug if the keyboard is open:
if I tap back or "x" while I'm typing, the keyboard remains open so I can't see the confirmation bottom sheet. 
I think we should unfocus the keyboard . not 100% if this is the same behavior on device but I think so


https://github.com/gallery-so/gallery/assets/80802871/81534438-36bc-41f2-a2a9-0eff9f736e35

jakzaizzat:
@kaitoo1 Updated the 2 comments regarding `TextInput` and latest dark mode styling

https://github.com/gallery-so/gallery/assets/4480258/d7231cfa-3e37-4461-acbe-291afce3e1e2


CleanShot 2023-07-19 at 10 48 03@2x

PR #1684, opened 2023-07-11
GAL-3568: View on Mirror and fix OpenSea link for certain chains

Rohan-cp:


packages/shared/src/utils/getOpenseaExternalUrl.ts
```diff
@@ -12,11 +12,12 @@ export const hexHandler = (str: string) => {
   return d;
 };
 
-export const getOpenseaExternalUrl = (contractAddress: string, tokenId: string) => {
+export const getOpenseaExternalUrl = (contractAddress: string, tokenId: string, chainStr: string = "ethereum") => {
```
Robinnnnn, line 15: looks like typescript linter is complaining about a small thing: when you default to a string literal, the type can be inferred, and therefore you could just do `contractAddress: string, tokenId: string, chainStr = 'ethereum'`, without specifying the type for the final param

image

that said, i suggest _not_ adding a default parameter and keeping it `chainStr: string`, and forcing the function caller to pass in a chain explicitly

also this is purely aesthetics, but i think the order of parameters should be:

`chainStr: string, contractAddress: string, tokenId: string`

Robinnnnn: oh and to check locally if you have any typescript issues, you can run:
`moon run web:typecheck`

packages/shared/src/utils/extractMirrorXyzUrl.tsx
```diff
@@ -0,0 +1,10 @@
+export const extractMirrorXyzUrl = (tokenMetadata: string ) => {
+  const metadataObj = JSON.parse(tokenMetadata);
+  const tokenDesc = metadataObj.description;
+  const startsWithMirrorXYZ = tokenDesc?.startsWith("https://mirror.xyz");
+  const hasWhitespaceInMiddle = /\s/.test(tokenDesc);
+  if (startsWithMirrorXYZ && !hasWhitespaceInMiddle) {
+    return tokenDesc
+  }
+  return "";
+};
```
Robinnnnn, line 10: nice! could we make sure to unit test this?

PR #1699, opened 2023-07-13
hide remove pfp option if no pfp set

kaitoo1:
## Description
hide remove pfp option if no pfp set

Screenshot 2023-07-13 at 21 56 45


kaitoo1:
will need to do this for mobile too

Robinnnnn (approved):

PR #1701, opened 2023-07-13
GAL-3582: add optimism support for community page

SouravInsights:
## Description
Added a new page like this: `pages/community/optimism/[contractAddress]/index.tsx` this means we would have to new pages like this for all kinds of L2s, we could probably think of a better way to do this instead of copy pasting the same code. Let me know your thoughts @Robinnnnn on this.

Updated the `getExternalAddressLink` to account for etherscan link of optimism.

`Note:` still need to make changes accordingly for mobile

Demo: https://www.loom.com/share/8147b7f00cc44b4b9db4043c4e53ff82?sid=1013b59f-13b7-46ab-9c16-e5899ce93c95


Robinnnnn (commented):
@SouravInsights this looks fantastic! 🔥 

and you're right, we're going to eventually want to do this for the other chains we're planning to support: Zora, Arbitrum, and Polygon. we'll likely reuse a lot of code there.

the main reusable code seems to live in `community/X/[contractAddress]/index.ts`. 99% is the same across different chains, with only a few differences.

i wonder what this would look like if we renamed the file and used a routing structure like: `apps/web/pages/community/[chain]/[contractAddress]/index.ts`. then use the `chain` from `getServerSideProps` params to feed into the graphql query. want to try experimenting with that?

Robinnnnn:
closing in favor of #1703

PR #1710, opened 2023-07-18
[GAL-3613] Post Composer [web]

kaitoo1:
## Description
Add Post Composer on Web

Screenshot 2023-07-19 at 19 42 21

This enables creating a post from the Sidebar as well as the NFT Detail Page for any NFT you own.

Adds a `PostComposerModalWithSelector` modal that enables a user to select an item with the NftSelector, then compose a post with the PostComposer. There is an alternate modal, `PostComposerModal`, that only has the PostComposer (no selector) - this is used when creating a post directly from an nft detail page.

TODOs
- [x] mobile responsiveness
- [x] feature flag for composer
- [x] feature flag for nft detail page
- [x] catch all error handling for composer
- [ ] error handling when posting (need error state design) (follow up ticket: GAL-3841)
- [ ] close modal confirmation  (follow up ticket GAL-3844)
- [ ] handle vertical, horizontal display (follow up ticket: GAL-3843)

Screenshot 2023-07-19 at 19 42 42
Screenshot 2023-07-19 at 19 42 19

### DIFF
--- a/apps/web/src/components/CommunityHolderGrid/CommunityHolderGrid.tsx
+++ b/apps/web/src/components/CommunityHolderGrid/CommunityHolderGrid.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
-import { graphql, usePaginationFragment } from 'react-relay';
+import { graphql, useFragment, usePaginationFragment } from 'react-relay';
 import styled from 'styled-components';
 
 import breakpoints from '~/components/core/breakpoints';
@@ -9,14 +9,16 @@ import { TitleS } from '~/components/core/Text/Text';
 import { GRID_ITEM_PER_PAGE } from '~/constants/community';
 import { GLOBAL_FOOTER_HEIGHT } from '~/contexts/globalLayout/GlobalFooter/GlobalFooter';
 import { CommunityHolderGridFragment$key } from '~/generated/CommunityHolderGridFragment.graphql';
+import { CommunityHolderGridQueryFragment$key } from '~/generated/CommunityHolderGridQueryFragment.graphql';
 
 import CommunityHolderGridItem from './CommunityHolderGridItem';
 
 type Props = {
   communityRef: CommunityHolderGridFragment$key;
+  queryRef: CommunityHolderGridQueryFragment$key;
 };
 
-export default function CommunityHolderGrid({ communityRef }: Props) {
+export default function CommunityHolderGrid({ communityRef, queryRef }: Props) {
   const {
     data: community,
     loadNext,
@@ -51,6 +53,15 @@ export default function CommunityHolderGrid({ communityRef }: Props) {
       }
     `,
     communityRef
+  );
+
+  const query = useFragment(
+    graphql`
+      fragment CommunityHolderGridQueryFragment on Query {
+        ...CommunityHolderGridItemQueryFragment
+      }
+    `,
+    queryRef
   );
 
   const [isFetching, setIsFetching] = useState(false);
@@ -109,7 +120,9 @@ export default function CommunityHolderGrid({ communityRef }: Props) {
           <TitleS>Collectors on Gallery</TitleS>
           <StyledCommunityHolderGrid>
             {galleryMemberTokens.map((holder) =>
-              holder ? <CommunityHolderGridItem key={holder.id} holderRef={holder} /> : null
+              holder ? (
+                <CommunityHolderGridItem key={holder.id} holderRef={holder} queryRef={query} />
+              ) : null
             )}
           </StyledCommunityHolderGrid>
         </VStack>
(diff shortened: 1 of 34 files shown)


apps/web/src/components/CommunityHolderGrid/CommunityHolderGrid.tsx
```diff
@@ -53,6 +55,15 @@ export default function CommunityHolderGrid({ communityRef }: Props) {
     communityRef
   );
```
kaitoo1, line 57: only change in file is prop-drilling query

apps/web/src/components/CommunityHolderGrid/CommunityHolderGridItem.tsx
```diff
@@ -8,6 +8,7 @@ import { VStack } from '~/components/core/Spacer/Stack';
 import { BaseM } from '~/components/core/Text/Text';
```
kaitoo1, line 8: only change in file is prop-drilling query

apps/web/src/components/Feed/Events/CollectorsNoteAddedToTokenFeedEvent.tsx
```diff
@@ -11,6 +11,7 @@ import HoverCardOnUsername from '~/components/HoverCard/HoverCardOnUsername';
 import { ProfilePicture } from '~/components/ProfilePicture/ProfilePicture';
```
kaitoo1, line 11: only change in file is prop-drilling query

apps/web/src/components/Posts/PostComposer.tsx
```diff
@@ -0,0 +1,145 @@
...
+    track('Clicked Post in Post Composer', {
+      added_description: !!description,
+    });
+    try {
+      await createPost({
+        tokenIds: [token.dbid],
+        caption: description,
+      });
+    } catch (error) {
+      // TODO add error state
```
kaitoo1, line 63: this can be a follow up PR , made ticket GAL-3841

apps/web/src/scenes/CommunityPage/CommunityPageView.tsx
```diff
@@ -16,6 +16,7 @@ import MemberListFilter from '~/components/TokenHolderList/TokenHolderListFilter
 import { GRID_ENABLED_COMMUNITY_ADDRESSES } from '~/constants/community';
```
kaitoo1, line 16: only change in this file is drilling query

apps/web/src/components/Posts/PostComposerModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { NftSelector } from '../NftSelector/NftSelector';
+import PostComposer from './PostComposer';
+
+type Props = {
+  tokensRef: PostComposerModalWithSelectorFragment$key;
+  queryRef: PostComposerModalWithSelectorQueryFragment$key;
+};
+
+// Modal with multiple steps: the NFT Selector -> then Post Composer
+export function PostComposerModalWithSelector({ tokensRef, queryRef }: Props) {
```
Robinnnnn, line 26: nice!! we could prob consolidate into a single component with an option (`withSelector: boolean`) but i think this is cleaner / easier to reason about, esp if the modals end up having diverging use cases

Robinnnnn (commented):
for this view, could we just have the number value in red and ditch `Max text reached`? i think the number will be self-explanatory. could we also disable the POST button (this is essentially the same logic as the bio modal)

image

thoughts on removing this X?
1. it's a bit distracting, only shows up when you hover, and almost entices you to click on it because it appears so smoothly lol
2. user already has the option to click Back or X out of the modal
3. we can keep this functionality for later when we have multi-NFT posting

Screenshot 2023-07-25 at 5 18 04 PM

apps/web/src/components/NftSelector/NftSelectorToken.tsx
```diff
@@ -35,13 +37,13 @@ export function NftSelectorToken({ tokenRef, isInGroup = false }: Props) {
           }
         }
         ...RawNftSelectorPreviewAssetFragment
+        # eslint-disable-next-line relay/must-colocate-fragment-spreads
+        ...PostComposerFragment
```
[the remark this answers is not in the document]
kaitoo1: was able to remove this thanks to your comment below!

apps/web/src/components/NftSelector/NftSelector.tsx
```diff
@@ -33,11 +34,13 @@ import { NftSelectorView } from './NftSelectorView';
 type Props = {
   tokensRef: NftSelectorFragment$key;
   queryRef: NftSelectorQueryFragment$key;
+  onSelectToken: (token: NftSelectorTokenFragment$data) => void;
```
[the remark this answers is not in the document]
kaitoo1: yess agree. thanks for the callout, I thought I needed it but ended up not having to do this and pass `tokenId` as suggested instead!

apps/web/src/components/Posts/PostComposerModal.tsx
kaitoo1: yeahh exactly - if we find we're repeating code a bunch we can definitely consolidate. I was deep trying to separate the nft selector for pfp and post use cases so it also felt clean to separate these two modal flows

kaitoo1:
address text input and X button comments based on input from Robin and Fraser ✅

Robinnnnn (approved):

PR #1737, opened 2023-07-27
[GAL-3847] Mobile NFT details regression

Rohan-cp:
Screenshot 2023-07-28 at 3 54 14 PM Screenshot 2023-07-28 at 8 14 28 PM

 Tezos Piece with no ID
Screenshot 2023-07-28 at 3 54 41 PM Screenshot 2023-07-28 at 4 49 55 PM
Screenshot 2023-07-28 at 7 36 22 PM

### DIFF
--- a/apps/mobile/src/screens/NftDetailScreen/DetailSection.tsx
+++ b/apps/mobile/src/screens/NftDetailScreen/DetailSection.tsx
@@ -75,18 +75,3 @@ export function DetailExternalLink({
     </InteractiveLink>
   );
 }
-
-export function DetailMoreInfoLink({ link }: { link: string }) {
-  return (
-    <InteractiveLink href={link} type="NFT Detail More Info URL">
-      <Typography
-        numberOfLines={1}
-        ellipsizeMode="middle"
-        className="text-sm"
-        font={{ family: 'ABCDiatype', weight: 'Regular' }}
-      >
-        More Info
-      </Typography>
-    </InteractiveLink>
-  );
-}
--- a/apps/mobile/src/screens/NftDetailScreen/DetailSection.tsx
+++ b/apps/mobile/src/screens/NftDetailScreen/DetailSection.tsx
@@ -4,6 +4,7 @@ import { View, ViewProps } from 'react-native';
 import { InteractiveLink } from '~/components/InteractiveLink';
 
 import { Typography } from '../../components/Typography';
+import { ExternalLinkIcon } from '../../icons/ExternalLinkIcon';
 
 export function DetailSection({
   children,
@@ -54,7 +55,7 @@ export function DetailExternalLink({
 }: DetailExternalLinkProps) {
   return (
     <InteractiveLink href={link} type={trackingLabel}>
-      <View className="flex flex-row space-x-1">
+      <View className="flex flex-row">
         <Typography
           numberOfLines={1}
           ellipsizeMode="middle"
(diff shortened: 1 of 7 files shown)


apps/mobile/src/screens/NftDetailScreen/DetailSection.tsx
```diff
@@ -19,7 +20,7 @@ export function DetailSection({
 export function DetailLabelText({ children }: PropsWithChildren) {
   return (
     <Typography
-      className="text-xs text-metal dark:color-metal"
+      className="text-xs text-[#707070] dark:color-metal"
```
Rohan-cp, line 23: @Robinnnnn need to find out how to not hardcore color here

Robinnnnn: i would try `shadow` or `faint`

image

Robinnnnn: i think you want shadow since this is in light mode

apps/mobile/src/screens/NftDetailScreen/NftAdditionalDetailsEth.tsx
```diff
@@ -79,42 +84,56 @@ export function NftAdditionalDetailsEth({ tokenRef }: NftAdditionalDetailsEthPro
...
-        )}
+          {externalUrl ? (
+            <DetailSection>
+              <DetailMoreInfoLink link={externalUrl} />
+            </DetailSection>
+          ) : (
+            <DetailSection>
+              <DetailLabelText>{'    '}</DetailLabelText>
+            </DetailSection>
+          )}
```
Robinnnnn, line 135: interesting, we add one anyway? feels like it might cause other issues

have you tried playing around with different permutations of elements?

Rohan-cp: yeah I don't think it will cause any problems because we render in this row a max of 3 things, and it works for all cases with elements <= 3

Screenshot 2023-07-28 at 8 26 18 AM
Screenshot 2023-07-28 at 8 29 11 AM
Screenshot 2023-07-28 at 12 45 31 PM

apps/mobile/src/screens/NftDetailScreen/NftAdditionalDetailsEth.tsx
```diff
@@ -59,16 +53,23 @@ export function NftAdditionalDetailsEth({ tokenRef }: NftAdditionalDetailsEthPro
     return null;
   }, [tokenMetadata]);
 
+  const numOfLinks = [openSeaExternalUrl, externalUrl, mirrorXyzUrl].filter((value) =>
+    value ? value : null
+  ).length;
```
Robinnnnn, line 58: nit:
1. should `useMemo` whenever we operate on arrays to cache computation
2. for something like this you could probably use [Array.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

```js
const numLinks = useMemo(() => {
  return [openSeaExternalUrl, externalUrl, mirrorXyzUrl].some((el) => Boolean(el))
}, [openSeaExternalUrl, externalUrl, mirrorXyzUrl])
```

Robinnnnn (approved):
So it looks like the prohibition URL detection logic was introduced into this file, but could probably exist outside (similar to OS and mirror detection)

I have an idea for consolidating the url extraction logic that I'll run by you later.

Not a blocker so will approve!!!

PR #1753, opened 2023-08-01
GAL-3706: auto period on comment input

SouravInsights:
## Description:
This seems to be a React Native's limitation. Please [refer](https://github.com/facebook/react-native/issues/22078) this. There is an ongoing [work](https://github.com/facebook/react-native/issues/37437) to fix this issue. 

As of now, the workaround is to set the `multiline` prop to `true`. And I don't presume making it multiline affects our CommentBox comment.

## Demo
https://www.loom.com/share/7ac2a321c718400ab03a23b090a927cc?sid=beaa36aa-9069-432b-b2c3-4dcf7895907a

### DIFF
--- a/apps/mobile/src/components/Feed/Socialize/CommentBox.tsx
+++ b/apps/mobile/src/components/Feed/Socialize/CommentBox.tsx
@@ -97,6 +97,7 @@ export function CommentBox({
         <BottomSheetTextInput
           value={value}
           onChangeText={setValue}
+          multiline={true}
           className="text-sm h-5"
           selectionColor={colorScheme === 'dark' ? colors.white : colors.black['800']}
           autoCapitalize="none"
--- a/apps/mobile/src/components/Feed/Socialize/CommentBox.tsx
+++ b/apps/mobile/src/components/Feed/Socialize/CommentBox.tsx
@@ -1,7 +1,13 @@
 import { BottomSheetTextInput } from '@gorhom/bottom-sheet';
 import { useColorScheme } from 'nativewind';
 import { useCallback, useLayoutEffect, useMemo, useState } from 'react';
-import { Text, View, ViewStyle } from 'react-native';
+import {
+  NativeSyntheticEvent,
+  Text,
+  TextInputContentSizeChangeEvent,
+  View,
+  ViewStyle,
+} from 'react-native';
 import Animated, {
   AnimatedStyleProp,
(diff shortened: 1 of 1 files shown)


Robinnnnn (approved):

apps/mobile/src/components/Feed/Socialize/CommentBox.tsx
```diff
@@ -97,6 +97,7 @@ export function CommentBox({
         <BottomSheetTextInput
           value={value}
           onChangeText={setValue}
+          multiline={true}
```
kaitoo1, line 100: nit: can just be `multiline` (don't need to pass true)

kaitoo1:
nice find @SouravInsights !! the period is showing up as expected ✅ 

I noticed that when I type on a new line, sometimes the vertical position glitches a bit and resets to showing the line above. attached a video as an example. 
would you be able to take a look? looks like it's new behavior due to it being multiline


https://github.com/gallery-so/gallery/assets/80802871/cb96c4df-6251-4e5b-baa4-ffca4b5778dd

SouravInsights:
Oh yeah I know why it's happening, I read somewhere that this is a possible issue we need to handle when we set `multiline` to true. I'll take a look! Sorry I missed it initially.

apps/mobile/src/components/Feed/Socialize/CommentBox.tsx
SouravInsights: yeah totally! 👍

SouravInsights:
hey, @kaitoo1 turns out we could adjust the height of the comment box dynamically based on the content size.. i think it fixed the glitch but let me know your feedback! here's [demo](https://www.loom.com/share/e68326911d9341a9ad2502f967ee0d69?sid=c7595ba9-3670-43f7-9ed5-86e94f2f51f3)

kaitoo1:
hey @SouravInsights thanks for looking for a solution. it looks like unfortunately it doesn't work when the keyboard is actually open on the screen- attached a recording.

https://github.com/gallery-so/gallery/assets/80802871/cb0be74f-b991-4cac-be83-66b1923f59f2


IMO the original issue of not supporting the auto-period iOS feature isn't a huge deal so let's table this for now - it's not a game changer and I don't think it's worth investing in a workaround since it's not supported by react native. thanks for researching alternatives 🙏

PR #1756, opened 2023-08-02
[GAL-3891] reduce the user bio and social pills for own profile

Rohan-cp:
Screenshot 2023-08-01 at 10 03 24 PM Screenshot 2023-08-01 at 9 30 21 PM Screenshot 2023-08-01 at 9 30 12 PM

### DIFF
--- a/apps/mobile/src/components/ProfileView/ProfileViewHeader.tsx
+++ b/apps/mobile/src/components/ProfileView/ProfileViewHeader.tsx
@@ -93,7 +93,7 @@ export function ProfileViewHeader({ queryRef, selectedRoute, onRouteChange }: Pr
         </View>
       )}
       {!isLoggedInUser && <ProfileViewSharedInfo userRef={user} />}
-      <View className={`flex flex-row mx-4 space-x-12 ${isLoggedInUser ? 'mt-2.5' : 'mt-4'}`}>
+      <View className={`flex flex-row mx-4 space-x-12 mt-4`}>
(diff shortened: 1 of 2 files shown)


apps/mobile/src/components/ProfileView/ProfileViewHeader.tsx
```diff
@@ -88,13 +88,12 @@ export function ProfileViewHeader({ queryRef, selectedRoute, onRouteChange }: Pr
...
       {user.bio && (
-        <View className="mb-4 px-4">
+        <View className="px-4">
           <Markdown>{user.bio}</Markdown>
         </View>
       )}
       {!isLoggedInUser && <ProfileViewSharedInfo userRef={user} />}
-
-      <View className="flex flex-row mx-4 space-x-12 ">
+      <View className={`flex flex-row mx-4 space-x-12 ${isLoggedInUser ? 'mt-2.5' : 'mt-4'}`}>
```
kaitoo1, line 96: instead of using `isLoggedInUser` to adjust the margin on this element, I think the absence of `ProfileViewSharedInfo` if `!isLoggedInUser` on line 95 should dictate the amount of extra space

so in other words the extra space should be on the conditional ProfileViewSharedInfo , or be set as a gap on the parent so that there's automatically more space if `ProfileViewSharedInfo` is present, and less space if `ProfileViewSharedInfo` is not present

Robinnnnn (approved):

kaitoo1 (approved):

PR #1774, opened 2023-08-03
[GAL-3954] Community Post

jakzaizzat:
**Changes**
- Added new variant (disabled) to the Button
- Allow user to post from community screen

**Demo**
**The user not belongs to the community**

https://github.com/gallery-so/gallery/assets/4480258/12b62549-8eed-43d9-848a-a73dd35a0de2

**The user belongs to the community**

https://github.com/gallery-so/gallery/assets/4480258/fcb5e90f-3793-419c-83c4-0a866dd5b806


apps/mobile/src/components/Community/CommunityMeta.tsx
```diff
@@ -70,39 +61,83 @@ export function CommunityMeta({ communityRef, queryRef }: Props) {
...
+          }
+        }
+      }
+    `,
+    {
+      communityID: community.dbid,
+    }
+  );
+
+  const isMemberOfCommunity = isMemberOfCommunityQuery.viewer?.user?.isMemberOfCommunity ?? false;
```
jakzaizzat, line 93: Anyone know what's the best way to refetch this query? 🤔

Context: Would love to refetch this value whenever the sync is completed

apps/mobile/src/navigation/MainTabNavigator/PostIcon.tsx
```diff
@@ -1,18 +1,25 @@
...
   size?: number;
 } & SvgProps;
 
 export const PostIcon = ({ color, size = 24, ...props }: Props) => {
+  const height = size;
+  const originalHeight = 24;
+  const scale = height / originalHeight;
+  const width = height;
+
   return (
```
kaitoo1, line 14: if we define a fixed viewBox here we should be able to change the size of the svg by just using width and height attributes. see `TwitterIcon` as an example

apps/mobile/src/components/Community/Tabs/CommunityViewPostsTab.tsx
```diff
@@ -57,6 +63,25 @@ export function CommunityViewPostsTab({ communityRef, queryRef }: Props) {
     queryRef
   );
 
+  const isMemberOfCommunityQuery = useLazyLoadQuery<CommunityViewPostsTabQuery>(
```
kaitoo1, line 66: are we able to put this on the existing `query`?

apps/mobile/src/components/Community/CommunityPostBottomSheet.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    onRefresh();
+    pushToast({
+      message: 'Successfully refreshed your collection',
+    });
+  }, [closeBottomSheet, community.chain, onRefresh, pushToast, syncTokens]);
+
+  return (
+    <GalleryBottomSheetModal
+      ref={(value) => {
```
kaitoo1, line 68: I see we repeat this function a lot, mostly every where that we use `GalleryBottomSheetModal`... do you know what is is for? I wonder if we can reduce the repetition some how

jakzaizzat: I hate this too 😔 

Seems like the latest version bottom sheet provide different types for `animatedSnapPoints`. Upgraded it because of keyboard crash on comment

apps/mobile/src/components/Community/Tabs/CommunityViewPostsTab.tsx
jakzaizzat: Good one! Merge it to existing query

kaitoo1 (approved):
LFG! LGTM

PR #1789, opened 2023-08-07
community pfp regression

SouravInsights:
### Summary of Changes
The issue was I hardcoded the the text color of `<InteractiveLink />` component like this 👇 to match up to Fraser's design:
```
<InteractiveLink
  onPress={onSeeAll}
  textStyle={{ fontSize: 12, color: 'text-black-800' }} // hardcoded color value
  font={{ family: 'ABCDiatype', weight: 'Bold' }}
  type="Shared Communities See All"
 >
  {children}
</InteractiveLink>
 ```
 So what I did now is I added a new prop called `isSecondary` that applies the `text-shadow` & `dark:text-white` class conditionally.

#### Before:
image

#### After
image
image

### Edge Cases
N/A

### Testing Steps
Just toggle the dark mode to see if the `SharedInfo` texts are looking fine.. 

### Checklist

- [ ] WEB: The changes have been tested on various desktop screen sizes to ensure responsiveness.

### DIFF
--- a/apps/mobile/src/components/InteractiveLink.tsx
+++ b/apps/mobile/src/components/InteractiveLink.tsx
@@ -10,6 +10,7 @@ import { Typography, TypographyProps } from './Typography';
 export type InteractiveLinkProps = PropsWithChildren<{
   href?: string;
   showUnderline?: boolean;
+  isSecondary?: boolean; // if true, the text color renders as normal Typography..
   onPress?: () => void;
   style?: TouchableOpacityProps['style'];
   // for tracking
@@ -24,6 +25,7 @@ export function InteractiveLink({
   style,
   onPress,
   showUnderline = false,
+  isSecondary = false,
   children,
   type,
   textStyle,
@@ -51,7 +53,8 @@ export function InteractiveLink({
       }}
     >
       <Typography
-        className={clsx(`text-shadow dark:text-white text-sm`, {
+        className={clsx(`text-sm`, {
+          'text-shadow dark:text-white': !isSecondary,
           underline: showUnderline,
         })}
         style={[textStyle, { lineHeight: undefined }]}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: apps/mobile/src/components/InteractiveLink.tsx [hunk 1]
```diff
@@ -10,6 +10,7 @@ import { Typography, TypographyProps } from './Typography';
 export type InteractiveLinkProps = PropsWithChildren<{
   href?: string;
   showUnderline?: boolean;
+  isSecondary?: boolean; // if true, the text color renders as normal Typography..
```
kaitoo1, line 13: I think `isSecondary` is a misleading prop name because it suggests that there's are primary and secondary versions of InteractiveLinks, which we don't have.  (unlike Button.tsx where we do have primary and secondary buttons that are designed to look unique)

instead of adding a prop here for this use case let's stick to our current pattern of conditionally passing a different color to textStyles depending on dark/light mode

example
https://github.com/gallery-so/gallery/blob/a50d634bb05c31a1c888c97d16bca74d56048601/apps/mobile/src/screens/NftDetailScreen/NftAdditionalDetailsTezos.tsx#L53

apps/mobile/src/components/ProfileView/ProfileViewSharedInfo/ProfileViewSharedCommunities.tsx
```diff
@@ -174,9 +181,13 @@ function HoldsText({ communityRefs, onSeeAll, style }: HoldsTextProps) {
 
           <InteractiveLink
             onPress={onSeeAll}
-            textStyle={{ fontSize: 12, color: 'text-black-800' }}
             font={{ family: 'ABCDiatype', weight: 'Bold' }}
             type="Shared Communities See All"
+            isSecondary
```
kaitoo1, line 186: can remove now

PR #1792, opened 2023-08-08
GAL-3892: fix shared following regression

SouravInsights:
### Summary of Changes

So the problem was we were using the `length` of the `users` data returned by the `ProfileViewSharedFollowersFollowingTextFragment` query which actually paginated or lazy loaded.. That's why it was showing the false count in the first render but once you open bottom sheet and scroll, the count gets updated to the correct number. 

So what I did is just used the `totalSharedFollowers` or `totalSharedCommunities` that refers to `pageInfo.total` which actually has the total number of shared followers or communities.

#### Before
https://www.loom.com/share/d5718de1d03249558df7350a63174bbb?sid=d3e18ab5-5a4c-402b-a5e5-7850501c3027

#### After
https://www.loom.com/share/e9e3511f754d4d18b86f1b56be72baf8?sid=e608cda7-5e8c-4b41-a81a-4d4ef8ac214d

### Edge Cases
N/A

### Testing Steps

Note: Since this issue only happens when you have more than 20 shared followers or communities, we need to make sure we are testing against those kind of users.

### Checklist

- [ ] WEB: The changes have been tested on various desktop screen sizes to ensure responsiveness.

### DIFF
--- a/apps/mobile/src/components/ProfileView/ProfileViewSharedInfo/ProfileViewSharedFollowers.tsx
+++ b/apps/mobile/src/components/ProfileView/ProfileViewSharedInfo/ProfileViewSharedFollowers.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import { useNavigation } from '@react-navigation/native';
 import clsx from 'clsx';
 import { useColorScheme } from 'nativewind';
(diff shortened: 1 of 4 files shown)


apps/mobile/src/components/ProfileView/ProfileViewSharedInfo/ProfileViewSharedFollowers.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
kaitoo1, line 1: can toss

kaitoo1 (approved):
LGTM thank you sourav! can remove `eslint-disable no-console` but otherwise good to go

kaitoo1 (approved):

PR #1796, opened 2023-08-10
add notification setting + prompt

kaitoo1:
### Summary of Changes

This PR adds a Notification Settings screen to the Settings menu.
Currently this is an admin-only feature since we want to add a bit more design polish before making it available to all users.

The primary goal is to give a user a chance to opt in to Push Notifications if they find that they're not receiving any. These users likely fall into 2 groups
1. they've never been prompted with the Enabled Notifications native iOS prompt
2. they've seen in and denied permission / turned off push notifications since

In the first case, pressing the new Enable Push Notifications button will trigger the prompt so that they can opt in. By showing the prompt, the Gallery app will also be listed in their iOS Settings app, so that they can toggle these settings on their own later. Without seeing this prompt, the app doesn't appear to be listed in the iOS settings app.
CleanShot 2023-08-10 at 16 50 38

In the 2nd case, pressing the new Enable Push Notifications button will show a different prompt that gives them an option to navigate to the Settings app. this is because iOS does not allow re-prompting the user with the direct notification permission dialog. This will take the user directly to the Gallery app setting in the native Settings app.
CleanShot 2023-08-10 at 16 50 44

 

This is role-flagged behind the ADMIN role.

### Demo or Before and After

Case 1: prompting for permission directly
https://github.com/gallery-so/gallery/assets/80802871/6fcecadb-d1e3-4615-a4f3-9afa3fe20b7b

Case 2: prompting to navigate to Settings app

https://github.com/gallery-so/gallery/assets/80802871/10f47820-012a-49fb-8d2d-092c0bd0b302

If Push Notifications are enabled, this is what they see
CleanShot 2023-08-10 at 16 52 05

### Edge Cases

- [x] Push notifications are enabled
- [x] Push notifications aren't enabled. user has never seen the prompt
- [x] Push notifications aren't enabled, but user has seen the prompt
- [x] Not an admin user

### Testing Steps

Toggle your push notifications for Gallery on/off and go to the Notification Settings screen in app.

### Checklist

- [x] WEB: The changes have been tested on various desktop screen sizes to ensure responsiveness.
- [ ] MOBILE APP: The changes have been tested on both light and dark modes.

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -17,6 +17,7 @@
     "date-fns": "^2.29.3",
     "dotenv": "^16.0.3",
     "expo": "~48.0.18",
+    "expo-application": "~5.1.1",
     "expo-asset": "^8.9.1",
     "expo-av": "~13.2.1",
     "expo-barcode-scanner": "~12.3.2",
--- a/apps/mobile/src/navigation/MainTabStackNavigator.tsx
+++ b/apps/mobile/src/navigation/MainTabStackNavigator.tsx
@@ -15,6 +15,7 @@ import { PostScreen } from '~/screens/PostScreen/PostScreen';
 import { ProfileScreen } from '~/screens/ProfileScreen/ProfileScreen';
 import { SearchScreen } from '~/screens/SearchScreen';
 import { SettingsProfileScreen } from '~/screens/SettingsProfileScreen';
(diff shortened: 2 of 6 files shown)


jakzaizzat (approved):
Looks good!

PR #1803, opened 2023-08-15
Fix comment didn't show up

jakzaizzat:
### Summary of Changes

Fix the comment didn't show up after submitting comment in the bottom sheet

### Demo or Before and After

**Before**

https://github.com/gallery-so/gallery/assets/4480258/8a0ef6e1-cb31-4f0f-9683-c15e1b67f50c

**After**

https://github.com/gallery-so/gallery/assets/4480258/2d278910-cb71-4094-8716-3b6e4880def3

### Edge Cases

1. Submit comment while opening bottom sheets 
2. Submit comment without opening bottom sheets

### Testing Steps

1. Open a comment bottom sheets (should be able to open if there is more than 3 comments)
2. Submit a comment and it should appear in the bottom sheets

### Checklist

### DIFF
--- a/apps/mobile/index.js
+++ b/apps/mobile/index.js
@@ -46,6 +46,7 @@ LogBox.ignoreLogs([
   'Overriding previous layout animation',
   'There was a problem sending log message',
   'Constants.manifest has been deprecated in favor of Constants.expoConfig',
+  "CommentsBottomSheet_comments_connection' doesn't exist.",
 ]);
 
 polyfillEncoding();
--- a/apps/mobile/index.js
+++ b/apps/mobile/index.js
@@ -46,6 +46,7 @@ LogBox.ignoreLogs([
   'Overriding previous layout animation',
   'There was a problem sending log message',
   'Constants.manifest has been deprecated in favor of Constants.expoConfig',
+  // Warning only occurs when commenting before opening the bottom sheet because it uses lazyLoadQuery, which initializes the connection
   "CommentsBottomSheet_comments_connection' doesn't exist.",
 ]);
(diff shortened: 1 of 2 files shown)


apps/mobile/index.js
```diff
@@ -46,6 +46,7 @@ LogBox.ignoreLogs([
   'Overriding previous layout animation',
   'There was a problem sending log message',
   'Constants.manifest has been deprecated in favor of Constants.expoConfig',
+  "CommentsBottomSheet_comments_connection' doesn't exist.",
```
jakzaizzat, line 49: Ignore the connection warning. Seems like the reason the comment didn't show up because of missing connection. 

At some cases, the connection didn't initialise yet and we get the warning. The reason is the bottom sheet modal is using `lazyLoadQuery` so the connection will init when the bottom sheet is open. 

The warning only appear when we submit the comment without open the bottom sheet

kaitoo1: gotcha! this wouldn't happen in a normal use case then right? would you be able to add a comment to the file so we can keep the context?

maybe something like 
`Warning only occurs when commenting before opening the bottom sheet because it uses `lazyLoadQuery`, which initializes the connection` ?

jakzaizzat: @kaitoo1 Yup! Normal use case's working fine. 

Adding the comment above the log

kaitoo1 (approved):

apps/mobile/index.js
kaitoo1: nice! tyyy

PR #1807, opened 2023-08-16
[GAL-3926] use isMemberOfCommunity to show different states

kaitoo1:
### Summary of Changes

This PR adds the use of isMemberOfCommunity to show different states of the Community Page depending on whether you own a token.

If a user is signed in and owns a token,
- they see a primary Post button that opens the Post flow
- they see the same button lower in the page if there are 0 posts

If a user is signed in and doesn't own a token
- they see a "disabled" Post button that opens a modal explaining why they can't post and and option to refresh their collection

if a user is not signed in, 
- they don't see any Post buttons 

### Demo or Before and After

User owns a token:

https://github.com/gallery-so/gallery/assets/80802871/dc80e12a-2f68-4547-a9e3-2ee8b9c556e2

User doesn't own: (signed in)

https://github.com/gallery-so/gallery/assets/80802871/c6a3c271-89e6-46ca-ba57-b1b04554381a

User is not signed in

https://github.com/gallery-so/gallery/assets/80802871/0b03632c-10a3-4162-81b3-9d2b8235e387

### Edge Cases
Main edge cases are covered above

moweb
CleanShot 2023-08-16 at 23 49 30@2x

### Testing Steps
Go to various community pages that you own and don't own tokens for and verify the state of the page

### Checklist

- [ ] The changes have been tested and all tests pass.
- [x] WEB: The changes have been tested on various desktop screen sizes to ensure responsiveness.
- [ ] MOBILE APP: The changes have been tested on both light and dark modes.

### DIFF
--- a/apps/web/src/contexts/communityPage/IsMemberOfCommunityContext.tsx
+++ b/apps/web/src/contexts/communityPage/IsMemberOfCommunityContext.tsx
@@ -10,6 +10,9 @@ type IsMemberOfCommunityState = {
   refetchIsMemberOfCommunity: () => void;
 };
 
+// Using a context because isMemberOfCommunity is used in different areas of the Community Page that are separate from each other.
+// Wanted to avoid them making redundant separate requests to retrieve this value.
+// isMemberOfCommunity is its own query because it requires a community id which we don't have until the first query for community data is resolved.
 export const IsMemberOfCommunityContext = createContext<IsMemberOfCommunityState | undefined>(
   undefined
 );
--- a/apps/web/src/scenes/CommunityPage/CommunityPageOwnershipRequiredModal.tsx
+++ b/apps/web/src/scenes/CommunityPage/CommunityPageOwnershipRequiredModal.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react';
+import { useCallback } from 'react';
 import { graphql, useFragment } from 'react-relay';
 import styled from 'styled-components';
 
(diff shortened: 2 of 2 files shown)


apps/web/src/contexts/communityPage/IsMemberOfCommunityContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return context;
+};
+
+type IsMemberOfCommunityProviderProps = {
+  children: ReactNode;
+  communityDbid: string;
+};
+
+export const IsMemberOfCommunityProvider = ({
```
kaitoo1, line 31: I decided to use a context because `isMemberOfCommunity` is used in different areas of the Community Page that are a bit separate from one another, and I wanted to avoid them having to make redundant separate requests to retrieve this value. isMemberOfCommunity is its own query because it requires a community id which we don't have until our first query for community data is  resolved.

jakzaizzat: That make sense! No need to do prop drilling too

apps/web/src/scenes/CommunityPage/CommunityPageOwnershipRequiredModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+      return;
+    }
+    setIsRefreshing(true);
+    await syncTokens(community.chain);
+    setIsRefreshing(false);
+    refetchIsMemberOfCommunity();
+    hideModal();
+  }, [community.chain, hideModal, refetchIsMemberOfCommunity, syncTokens]);
+
+  const [isRefreshing, setIsRefreshing] = useState(false);
```
jakzaizzat, line 50: nit: I think we have `isLocked` in the `useSyncTokens` hooks. Not really sure should we used the local state or state in the hooks 🤔

apps/web/src/contexts/communityPage/IsMemberOfCommunityContext.tsx
kaitoo1: sweet, I just added my comment as a comment to the file lol

jakzaizzat (approved):

apps/web/src/scenes/CommunityPage/CommunityPageOwnershipRequiredModal.tsx
kaitoo1: updated

PR #1842, opened 2023-08-25
GAL-4102 make Curated the default screen on mobile app

kaitoo1:
### Summary of Changes
Change the default screen on mobile app from Latest to Curated.
Also move the Beta Welcome bottom sheet from Latest to Curated so we can continue to show it.

### Demo or Before/After Pics
Before

https://github.com/gallery-so/gallery/assets/80802871/68b9c833-0665-49eb-823d-43b3ce5195a5

After
https://github.com/gallery-so/gallery/assets/80802871/8a48bde1-30da-4121-b792-a6cec1ce749e

### Edge Cases
- for first time sign in users we show a Welcome to the Beta bottom sheet. made sure we keep showing that

### Testing Steps
reboot the app, the first thing you should see is the Curated Screen

### Checklist

### DIFF
--- a/apps/mobile/src/screens/HomeScreen/CuratedScreen.tsx
+++ b/apps/mobile/src/screens/HomeScreen/CuratedScreen.tsx
@@ -107,7 +107,7 @@ function CuratedScreenInner({ queryRef }: CuratedScreenInnerProps) {
         feedEventRefs={events}
         queryRef={query.data}
       />
-      {showWelcome && <WelcomeToBeta username={query.viewer?.user?.username ?? ''} />}
+      {showWelcome && <WelcomeToBeta username={query.data.viewer?.user?.username ?? ''} />}
     </>
   );
 }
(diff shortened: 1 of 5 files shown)


apps/mobile/src/screens/HomeScreen/CuratedScreen.tsx
```diff
@@ -73,15 +83,32 @@ function CuratedScreenInner({ queryRef }: CuratedScreenInnerProps) {
...
+    const shown = await AsyncStorage.getItem('welcomeMessageShown');
+    if (shown !== 'true') {
+      setShowWelcome(true);
+      await AsyncStorage.setItem('welcomeMessageShown', 'true');
+    }
+  }, [setShowWelcome]);
+
+  useEffect(() => {
+    checkShouldShowWelcome();
+  }, [checkShouldShowWelcome]);
```
Robinnnnn, line 98: nit: this looks like an extremely useful hook to make generic, like [usePersistedState](https://github.com/gallery-so/gallery/blob/main/apps/web/src/hooks/usePersistedState.ts)

Robinnnnn (approved):

apps/mobile/src/screens/HomeScreen/CuratedScreen.tsx
kaitoo1: yesss True!! I think Rohan is adding one in his PR for debugger mode , do you think it's chill to wait to switch it out until that's merged? We already have this code in prod (just moving files) so should be minimal downside

Robinnnnn: 💯

PR #1845, opened 2023-08-25
GAL-4071 fix tiny typo in refresh bottom sheet and increase font size 

kaitoo1:
### Summary of Changes

 fix tiny typo in refresh bottom sheet (its -> it's)
and increase font size (sm -> md)

### Demo or Before/After Pics
before
CleanShot 2023-08-25 at 19 11 46@2x

after
CleanShot 2023-08-25 at 19 10 36@2x

### Edge Cases
n/a

### Testing Steps
go to a community page you don't have a token for and tap the post button

### Checklist

### DIFF
--- a/apps/mobile/src/components/Community/CommunityPostBottomSheet.tsx
+++ b/apps/mobile/src/components/Community/CommunityPostBottomSheet.tsx
@@ -96,7 +96,7 @@ function CommunityPostBottomSheet(
               font={{ family: 'ABCDiatype', weight: 'Regular' }}
             >
               Only {community.name} owners can post about {community.name}. If you own this item but
-              its not displaying try
+              it's not displaying try
             </Typography>{' '}
             <Typography
               className="text-sm text-black-900 dark:text-offWhite"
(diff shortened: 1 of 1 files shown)


Robinnnnn (commented):
a bit confused about the wording:
1. what does "refreshing your collection" refer to? do we expect the user to know this? not sure where i would need to trigger the refresh from (on the community page vs. some other view)
2. not sure if the text needs to be bolded

kaitoo1:
> a bit confused about the wording:
> 
> 1. what does "refreshing your collection" refer to? do we expect the user to know this? not sure where i would need to trigger the refresh from (on the community page vs. some other view)
> 
> 2. not sure if the text needs to be bolded

Ohh it's a bit cut off from the screenshot but there's a secondary button below OK that says "refresh collection"! So it should be pretty straightforward in this context

So given that, the text is bolded as a reference to the action you can take via the button in the bottom sheetimage

Robinnnnn (approved):

PR #1858, opened 2023-08-29
[GAL-4059] Remove All Networks view from Selector for now

Rohan-cp:
### Summary of Changes
on mobile app, we default to Ethereum in the NFT Selector similar to web as we can't show a refresh button for 'all' option only

### Demo
| Before | Before 2 | After | After 2 |
|--------|--------|--------|--------|
| Screenshot 2023-08-29 at 3 20 48 PM | Screenshot 2023-08-29 at 3 20 59 PM | Screenshot 2023-08-29 at 5 54 47 PM | Screenshot 2023-08-29 at 3 15 39 PM |

### Edge Cases
Hmm not sure about what else to consider

### Testing Steps
Can test locally on branch by trying to compose a post and see the default filters

### Checklist


Robinnnnn (commented):
oh man, i think there was some miscommunication here 😅 

we DO want to show the refresh button for each network. the problem was there's no way to refresh "all networks", only single ones. so the play was to just toss the all networks option (but **keep** the refresh button for the other networks)

Robinnnnn (approved):

PR #1878, opened 2023-09-03
fix unicode rendering in comment notification

motzko:
### Summary of Changes

The notification feed was rendering unicode characters in comments as strings. This PR adds dangerouslySetInnerHTML to the comment notification component to fix that issue.

### Demo or Before/After Pics

Screenshot 2023-09-03 at 7 57 31 AM

### Testing Steps

1. Click on notifications
2. Find a comment notification where the comment includes a special character

### Checklist

### DIFF
--- a/apps/web/src/components/Notifications/notifications/SomeoneCommentedOnYourPost.tsx
+++ b/apps/web/src/components/Notifications/notifications/SomeoneCommentedOnYourPost.tsx
@@ -10,6 +10,7 @@ import { SomeoneCommentedOnYourPostFragment$key } from '~/generated/SomeoneComme
 import { useReportError } from '~/shared/contexts/ErrorReportingContext';
 import getVideoOrImageUrlForNftPreview from '~/shared/relay/getVideoOrImageUrlForNftPreview';
 import colors from '~/shared/theme/colors';
+import unescape from '~/shared/utils/unescape';
 type Props = {
   notificationRef: SomeoneCommentedOnYourPostFragment$key;
   onClose: () => void;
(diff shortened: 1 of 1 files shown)


apps/web/src/components/Notifications/notifications/SomeoneCommentedOnYourPost.tsx
```diff
@@ -70,7 +70,7 @@ export default function SomeoneCommentedOnYourPost({ notificationRef, onClose }:
               commented on your <strong>post</strong>
             </BaseM>
           </StyledTextWrapper>
-          <StyledCaption>{comment.comment}</StyledCaption>
+          <StyledCaption dangerouslySetInnerHTML={{ __html: comment.comment }}></StyledCaption>
```
Robinnnnn, line 73: thank you so much for doing this man!

a few things:
1. we generally try to avoid `dangerouslySetInnerHTML` as it could result in dangerous code execution if a malicious user is able to encode a script via comments
2. we have a helper called `unescape` that accomplishes what you're trying to do. example usage can be found [here](https://github.com/gallery-so/gallery/blob/main/apps/web/src/scenes/UserGalleryPage/UserGalleryCollection.tsx#L86). so in this PR, you could replace with `{unescape(comment.comment)}`

on a higher level note, our eventual plan is to actually handle processing these strings on the backend before they're sent down to the client, so we don't have to sprinkle `unescape` throughout the web + mobile apps; it's been easy to forget to do this esp as we increase our product surface area.

TL;DR – client devs shouldn't have to worry about doing this in the future!

motzko: makes sense! was just copying what you all were already doing in the social feed https://github.com/gallery-so/gallery/blob/607b31148223324b44655a6a43e8c4d057ea058d/apps/web/src/components/Feed/Socialize/CommentLine.tsx#L55

pushed change using unescape

Robinnnnn: bruh, @tbezman is in trouble lol

a couple requests:
1. do you mind making the same changes in `CommentLine.tsx` and `CommentNote.tsx` where it's also used? (can be a separate PR)
2. could you re-open this PR directly from our repo, as opposed to your fork? this should allow the tests to run. let me know if you need any help!

image

PR #1902, opened 2023-09-11
[GAL-4259] include isBetaTester in events

kaitoo1:
### Summary of Changes
Include in events whether or not a user is a beta tester.
there are two methods: including the flag in every event, and setting the flag on the user profile.
there are pros and cons to each method, so we'll start by doing both to minimize gaps or incomplete data. we can evaluate if one turns out to be more reliable or accurate. for example, one difference is user role state - we may add or remove Beta Tester roles in the future, and including it in every event maintains historical context because you can see more directly whether a user was a beta tester at the time the event was triggered. 

### Demo or Before/After Pics

- If this PR makes visual changes, include before and after screenshots. You can type `/table` to quickly come up with a before/after table.

### Edge Cases

### Testing Steps

### Checklist

### DIFF
--- a/apps/web/src/contexts/analytics/WebAnalyticsProvider.tsx
+++ b/apps/web/src/contexts/analytics/WebAnalyticsProvider.tsx
@@ -4,6 +4,7 @@ import { memo, ReactNode } from 'react';
 
 import AnalyticsProvider, {
   IdentifyFunction,
+  RegisterSuperPropertiesFunction,
   TrackFunction,
 } from '~/shared/contexts/AnalyticsContext';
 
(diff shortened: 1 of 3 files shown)


packages/shared/src/contexts/AnalyticsContext.tsx
```diff
@@ -66,22 +78,27 @@ const AnalyticsProvider = memo(({ children, identify, track }: Props) => {
...
         if (!userId) {
           return;
         }
-
         identify(userId);
+
+        if (user.roles?.includes('BETA_TESTER')) {
+          setIsBetaTester(true);
+          registerSuperProperties({ isBetaTester: true });
+        }
```
jakzaizzat, line 93: beautiful 😍

apps/web/src/contexts/analytics/WebAnalyticsProvider.tsx
```diff
@@ -41,11 +42,26 @@ export const _identify: IdentifyFunction = (userId) => {
...
 
+export const _registerSuperProperties: RegisterSuperPropertiesFunction = (eventProps) => {
+  if (!mixpanelEnabled) return;
+
+  try {
+    mixpanel.register(eventProps);
+  } catch (error: unknown) {
+    // mixpanel errors shouldn't disrupt app
+    captureException(error);
+  }
```
jakzaizzat, line 53: Nice catch!

jakzaizzat (approved):
LGTM!

PR #1905, opened 2023-09-12
[GAL-4263] [GAL-4264] Make links in Post description on mobile and web app are clickable

Rohan-cp:
### Summary of Changes

### Demo
https://www.loom.com/share/1e39ba4c68914d28867ec863620b3939
### After on Mobile (Light mode)
| 1 | 2 | 3 |
|--------|--------|--------|
| Screenshot 2023-09-14 at 8 17 55 PM | Screenshot 2023-09-14 at 2 44 33 AM| Screenshot 2023-09-14 at 2 44 40 AM

### After on Mobile (Dark mode)
| 1 | 2 | 3 |
|--------|--------|--------|
|  Screenshot 2023-09-14 at 8 18 34 PM | Screenshot 2023-09-14 at 2 43 51 AM | Screenshot 2023-09-14 at 2 43 59 AM |

### Demo wrap issue resolved
| Before | After |
|--------|--------|
| Screenshot 2023-09-13 at 8 13 44 AM | Screenshot 2023-09-13 at 8 13 49 AM | 

### Edge Cases
Tested to see if changes to Markdown component still works elsewhere in app

### Testing Steps
Can test locally on branch by looking at a post with a valid URL on web or mobile

### Checklist


apps/mobile/src/components/Feed/Posts/WarningLinkBottomSheet.tsx
```diff
@@ -0,0 +1,84 @@
...
+        onLayout={handleContentLayout}
+        style={{ paddingBottom: bottom }}
+        className="p-4 flex flex-col space-y-6"
+      >
+        <View className="flex flex-col space-y-4">
+          <Typography
+            className="text-lg text-black-900 dark:text-offWhite"
+            font={{ family: 'ABCDiatype', weight: 'Bold' }}
+          >
+            Leaving gallery.so
```
kaitoo1, line 56: O lets change to Leaving Gallery since we're not on the site

kaitoo1, line 56:
```suggestion
            Leaving Gallery
```

apps/mobile/src/components/Feed/Posts/WarningLinkBottomSheet.tsx
```diff
@@ -0,0 +1,84 @@
...
+        </View>
+
+        <Button
+          variant="secondary"
+          onPress={handleCancel}
+          text="CANCEL"
+          eventElementId={null}
+          eventName={null}
+        />
+        <Button onPress={handleOpen} text="OPEN" eventElementId={null} eventName={null} />
```
kaitoo1, line 76: could you add 

eventElementId="External URL Confirmation Open Button"
eventName="Pressed External URL Confirmation Open Button"

apps/mobile/src/components/Feed/Posts/WarningLinkBottomSheet.tsx
```diff
@@ -0,0 +1,84 @@
...
+              {props.redirectUrl}
+            </Typography>
+          </Typography>
+        </View>
+
+        <Button
+          variant="secondary"
+          onPress={handleCancel}
+          text="CANCEL"
+          eventElementId={null}
```
kaitoo1, line 73: could you add

eventElementId="External URL Confirmation Cancel Button"
eventName="Pressed External URL Confirmation Cancel Button"

packages/shared/src/utils/convertToMarkdownLinks.test.ts
```diff
@@ -0,0 +1,13 @@
+import { convertToMarkdownLinks } from './convertToMarkdownLinks';
+
+describe('convertToMarkdownLinks', () => {
```
kaitoo1, line 3: lets gooo love tests thank you

kaitoo1:
nice!! this looks great, really big quality of life improvement 🔥 🔥 

one edge case I think we could handle is if the caption already is a markdown formatted url. we don't have a markdown url shortcut in the caption editor today so it's not super likely, but I've seen some users post with this format - maybe they're used to using markdown for collection and gallery descriptions. currently the valid markdown would get re-markdown'd

CleanShot 2023-09-13 at 11 42 31@2x

CleanShot 2023-09-13 at 11 42 35@2x

CleanShot 2023-09-13 at 11 42 40@2x

apps/mobile/src/components/Markdown.tsx
```diff
@@ -43,6 +44,7 @@ const markdownItOptions = MarkdownIt({ typographer: true, linkify: false }).disa
 
 export function Markdown({
   children,
+  onBypassLinkPress,
```
kaitoo1, line 47: what does bypass mean in this context? I might be missing something but its not immediately to me clear what type of function should be passed here , wondering if another name might make it more apparent

packages/shared/src/utils/convertToMarkdownLinks.ts
```diff
@@ -0,0 +1,6 @@
+// Regular expression to match URLs
+const urlRegex = /(?:https|ftp):\/\/[^\s/$.?#].[^\s]*[^\s.,!?#$]/g;
```
kaitoo1, line 2: how come we wanna check for / allow ftp specifically?

kaitoo1: maybe we can use the `VALID_URL` regex pattern defined in util/regex.ts in web (can move here), we use that one right now to detect urls when they get pasted into the collection description/bio editor I believe

that one is a bit more flexible by allowing urls that start with `www` instead of `https://` but also bit more strict about disallowing special characters in the domain

apps/mobile/src/components/Markdown.tsx
Rohan-cp: Yeah, so when you click on a link in `Markdown` it opens the link in the browser (this is the default `onLinkPress` behaviour) but if the `onBypassLinkPress` is provided to the `Markdown` tag  then when you click on a link `onBypassLinkPress` runs instead of directly opening the link in the browser.

If you provide `onBypassLinkPress` and still wish to redirect after the link is pressed or something you can add it at the end of the `onBypassLinkPress` function provided using the `Linking` import from react-native). 

I added a comment later in the file to kind of explain this, lmk if it makes sense and if I should move it up to the prop section of the file.

packages/shared/src/utils/convertToMarkdownLinks.ts
Rohan-cp: @kaitoo1 updated the regex and test to allow for links starting with "www" and disallow "ftp" but did not use the `VALID_URL` because I think @Robinnnnn mentioned on slack thread that we should allow p much any link and that logic actually disallows some of the links in the test which seem like common use cases.

apps/mobile/src/components/Markdown.tsx
```diff
@@ -93,14 +100,28 @@ export function Markdown({
...
   }, []);
 
+  const handleLinkPress = (url: string) => {
+    if (url && onBypassLinkPress) {
+      onBypassLinkPress(url);
+      return false;
+    }
+
+    return true;
+  };
```
Robinnnnn, line 110: useCallback

packages/shared/src/utils/convertToMarkdownLinks.test.ts
Robinnnnn: +1!!

packages/shared/src/utils/convertToMarkdownLinks.ts
```diff
@@ -0,0 +1,8 @@
+// Regular expression to match URLs
+const urlRegex =
+  /(?:https:\/\/(?:www\.[^\s/$.?#].[^\s]*[^\s.,!?#$]|(?!www\.)[^\s/$.?#].[^\s]*[^\s.,!?#$])|www\.[^\s/$.?#].[^\s]*[^\s.,!?#$])/g;
+
+export const convertToMarkdownLinks = (text: string) => {
+  const unformattedStr = text.replace(/\[([^\]]+)\]\([^)]+\)/g, '$1');
+  return unformattedStr.replace(urlRegex, (url: string) => `[${url}](${url})`);
+};
```
Robinnnnn, line 8: nit: could we rename this function to be `replaceUrlsWithMarkdownFormat`? otherwise the function sounds like it'll convert _any_ arbitrary string to markdown links

apps/mobile/src/components/Markdown.tsx
Robinnnnn: yup makes sense, can we move the comment up above L40 where the prop is defined?

can also say something like: "if provided:, this function will be executed instead of navigating to the underlying link"

kaitoo1: ah ok gotcha, thanks for explaining! makes sense, comment is super helpful and agree with robin putting the comment where the prop is defined 🙏

Robinnnnn (approved):

PR #1910, opened 2023-09-13
[GAL-4227] truncate string in hover card

ka17068:
### Summary of Changes

Truncate long strings in community hover cards and check for user hover cards.

### Before
image

### After
Screenshot 2023-09-14 at 2 22 29 AM
Screenshot 2023-09-14 at 2 21 37 AM
Screenshot 2023-09-14 at 2 22 11 AM

### Testing Steps

Try with ETH addresses and long user names

### DIFF
--- a/apps/web/src/components/HoverCard/HoverCardCommunityInner.tsx
+++ b/apps/web/src/components/HoverCard/HoverCardCommunityInner.tsx
@@ -138,6 +138,13 @@ export function HoverCardCommunityInner({
     return null;
   }
 
+  const formatEthAddress = (e: string) => {
+    return e.slice(0, 4) + '...' + e.slice(-4);
+  };
+
+  // check if name is eth address and truncate
+  const hasAddress = Boolean(community.name.startsWith('0x') && community.name.length > 20);
+
   const hasDescription = Boolean(community.description);
   return (
     <VStack gap={6}>
@@ -145,9 +152,11 @@ export function HoverCardCommunityInner({
         <StyledLink href={communityProfileLink}>
           <CommunityProfilePicture communityRef={community} size={64} />
         </StyledLink>
-        <VStack gap={2}>
+        <Section gap={2}>
           <StyledLink href={communityProfileLink}>
-            <StyledCardTitle>{community.name}</StyledCardTitle>
+            <StyledCardTitle>
+              {hasAddress ? formatEthAddress(community.name) : community.name}
+            </StyledCardTitle>
           </StyledLink>
           {community.description && (
             <StyledCardDescription>
@@ -156,7 +165,7 @@ export function HoverCardCommunityInner({
               </BaseM>
             </StyledCardDescription>
           )}
-        </VStack>
+        </Section>
       </HStack>
       <HStack align="center" gap={4}>
         <ProfilePictureStack usersRef={owners} total={totalOwners} />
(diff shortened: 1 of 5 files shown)


apps/web/src/components/HoverCard/HoverCardUsernameInner.tsx
```diff
@@ -126,13 +126,17 @@ export function HoverCardUsernameInner({ preloadedQuery }: Props) {
...
 }
 
+const Section = styled(VStack)`
+  width: 100%;
+`;
+
 const StyledCardHeader = styled(HStack)`
   display: flex;
-
+  gap: 2px;
```
Robinnnnn, line 139: nit: the component that this extends, `HStack`, takes `gap` as a property so you can pass this directly into `StyledCardHeader`!

```
<StyledCardHeader gap={2} ... />
```

apps/web/src/components/HoverCard/HoverCardOnUsername.tsx
```diff
@@ -157,7 +157,7 @@ const StyledCardContainer = styled.div`
   width: 375px;
   min-height: 128px;
   max-width: calc(100vw - ${pageGutter.mobile * 2}px);
-  display: grid;
```
Robinnnnn, line 160: we also have a corresponding component for communities, `HoverCardOnCommunity`. should we update this there as well?

apps/web/src/components/HoverCard/HoverCardCommunityInner.tsx
```diff
@@ -175,16 +184,20 @@ const StyledLink = styled(Link)`
   min-width: 0;
 `;
 
+const Section = styled(VStack)`
+  max-width: 250px;
+`;
```
Robinnnnn, line 189: in `HoverCardUesrnameInner`, the width is set to `100%`, but here it's set to 250px. any reason for that difference?

want to be wary of these components diverging!

apps/web/src/components/HoverCard/HoverCardCommunityInner.tsx
```diff
@@ -138,16 +138,25 @@ export function HoverCardCommunityInner({
     return null;
   }
 
+  const formatEthAddress = (e: string) => {
+    return e.slice(0, 4) + '...' + e.slice(-4);
+  };
```
Robinnnnn, line 143: when defining functions within a react component, we want to make sure to wrap it in `useCallback` for performance reasons.

that said, this function is generic enough where it's defined elsewhere. you should be able to do `import { truncateAddress } from '~/shared/utils/wallet'`

apps/web/src/components/HoverCard/HoverCardCommunityInner.tsx
```diff
@@ -138,16 +138,25 @@ export function HoverCardCommunityInner({
...
+
+  // check if name is eth address and truncate
+  const hasAddress = Boolean(community.name.startsWith('0x') && community.name.length > 20);
```
Robinnnnn, line 146: hmm how did we decide on `20` to be the arbitrary cutoff for name length? i think a more rigorous check would be something like this:

```ts
function isValidEthereumAddress(address: string): boolean {
  const pattern = /^0x[a-fA-F0-9]{40}$/;
  return pattern.test(address);
}
```

^ let's define this function in a new file, like `shared/utils/isValidEthereumAddress`, then use it in this component:


```tsx
import { isValidEthereumAddress } from '~/shared/utils/isValidEthereumAddress'
import { truncateAddress } from '~/shared/utils/wallet';

...

<StyledCardTitle>
  {isValidEthereumAddress(community.name) ? truncateAddress(community.name) : community.name}
</StyledCardTitle>
```

Robinnnnn (commented):
when making changes to the UI, let's make sure to add before/after screenshots 🙏

apps/web/src/components/HoverCard/HoverCardCommunityInner.tsx
ka17068: moving this function inside wallet utils, since it checks thats its a valid eth wallet address.

Robinnnnn (approved):

PR #1913, opened 2023-09-14
Marfa Screen

kaitoo1:
### Summary of Changes

This PR adds the presentation view that will display Posts live on screen at Marfa.
It will display and cycle through Posts featuring our partner artists' project.

organization of components:
CleanShot 2023-09-14 at 17 00 15@2x

### Demo or Before/After Pics
CleanShot 2023-09-19 at 21 37 16@2x

### Edge Cases
- empty state: no posts
- there aren't any new posts to load
- there are new posts to load

### Testing Steps
for now, go to http://localhost:3000/community/ethereum/0x7e619a01e1a3b3a6526d0e01fbac4822d48f439b?displayMode=presentation

### Checklist


apps/web/src/components/Feed/PostItem.tsx
```diff
@@ -55,9 +61,18 @@ export function PostItem({
...
 
-  if (!isDesktop) {
+  const useVerticalLayout = !isDesktop || bigScreenMode;
+
+  const reportError = useReportError();
+
+  if (!post.author) {
+    reportError('Post author is undefined', { tags: { postId: post.dbid } });
+    return null;
+  }
```
Robinnnnn, line 71: we want to make sure to avoid calling `reportError` directly in the render function

here we can just `throw` instead, since everywhere the `PostItem` component is rendered is already wrapped in a Suspense boundary that will handle it:

```ts
  if (!post.author) {
    throw new ErrorWithSentryMetadata('Post author is undefined', { postId: post.dbid }); // no need to specify `tags`
  }
```

apps/web/src/components/Feed/PostItem.tsx
```diff
@@ -145,30 +165,36 @@ export function PostItemWithBoundary({
...
 
   padding: 12px 0px;
 
   @media only screen and ${breakpoints.desktop} {
     padding: 24px 16px;
     max-width: initial;
-    width: ${FEED_EVENT_ROW_WIDTH_DESKTOP}px;
+    width: ${({ bigScreenMode }) => (bigScreenMode ? 544 : FEED_EVENT_ROW_WIDTH_DESKTOP)}px;
+
+    ${({ bigScreenMode }) => bigScreenMode && ``}
```
Robinnnnn, line 198: wat dis

apps/web/src/components/Feed/Socialize/RemainingCommentCount.tsx
```diff
@@ -43,13 +45,17 @@ export function RemainingCommentCount({
...
+  const { query: routerQuery } = useRouter();
+
   const openCommentsModal = useCommentsModal({ eventRef: event, queryRef: query });
 
   if (totalComments <= 1) {
     return null;
   }
 
-  return (
+  return routerQuery.displayMode === CommunityPageDisplayMode.PRESENTATION ? (
```
Robinnnnn, line 56: supernit: something like this could probably be a nice util in the future:

```ts
useRouteQuery(key, value): boolean

// e.g.

const isPresentationMode = useRouteQuery('displayMode', CommunityPageDisplayMode.PRESENTATION)
```

apps/web/src/components/Feed/PostItem.tsx
kaitoo1: ooo perfect ok good call! ty

apps/web/src/components/Feed/PostItem.tsx
kaitoo1: what the

kaitoo1: deleted

PR #1937, opened 2023-09-21
Cache the video on load

jakzaizzat:
### Summary of Changes

The idea:

1. Download the video when the user in the splash screen (same as phase when we setup our color scheme and font)
2. Store it inside the user device cache folder
3. Save the path inside the local storage
4. We just load the video from the local storage path

### Demo or Before/After Pics

https://github.com/gallery-so/gallery/assets/4480258/7c934240-ef0c-4b8c-8624-da7be048ea33

### Edge Cases

1. Test open the app with clear cache + local storage
2. Reopen the app and check whether are we downloading the video again

### Testing Steps

**Before**
1. Clear the local storage through expo debugger
2. Run with this command `yarn workspace mobile expo start --clear` to fresh start the app with clear cache

**Steps**
1. Enter with secret gesture
2. You should see the intro video instantly

### Checklist

### DIFF
--- a/apps/mobile/src/App.tsx
+++ b/apps/mobile/src/App.tsx
@@ -29,6 +29,7 @@ import SyncTokensProvider from './contexts/SyncTokensContext';
 import ToastProvider from './contexts/ToastContext';
 import { TokenStateManagerProvider } from './contexts/TokenStateManagerContext';
 import { magic } from './magic';
+import { useCacheIntroVideo } from './screens/Onboarding/useCacheIntroVideo';
 
 SplashScreen.preventAutoHideAsync();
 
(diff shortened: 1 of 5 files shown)


apps/mobile/src/App.tsx
```diff
@@ -116,7 +118,7 @@ export default function App() {
     [colorSchemeLoaded, fontsLoaded]
   );
 
-  if (!fontsLoaded || !colorSchemeLoaded) {
+  if (!fontsLoaded || !colorSchemeLoaded || !introVideoLoaded) {
     return null;
```
kaitoo1, line 122: what happens when we return null?

kaitoo1 (approved):

apps/mobile/src/App.tsx
jakzaizzat: Still showing Splash screen (the Gallery logo)

PR #1940, opened 2023-09-21
Wrap deeplink handler in try/catch to prevent crashes

Robinnnnn:
### Summary of Changes

Deeplink URL parsing was a likely cause for deeplink crashes

### Recommended Testing Steps

- [x] Open a link that is known to crash the app in previous versions
- [x] Ensure the link no longer crashes the app on this branch
- [x] Ensure a compose link opens the app. Logged in -> home page, Logged out -> sign in view (example URL: `https://gallery.so/~/compose?tokenId=98&contractAddress=0xE480a895DE49B49e37A8f0a8bD7e07FC9844CDb9&chain=Arbitrum&collection_title=Dots&token_title=Dots+%2398&caption=I+just+minted+Dots+%2398+by+wutsit2u+on+Prohibition`)
- out of scope: opening the post composer screen specifically

### Checklist

### DIFF
--- a/apps/mobile/src/components/DeepLinkRegistrar.tsx
+++ b/apps/mobile/src/components/DeepLinkRegistrar.tsx
@@ -11,6 +11,8 @@ import { useReportError } from '~/shared/contexts/ErrorReportingContext';
 
 const MARFA_EVENT_HASH = -1082633448;
 
+const KNOWN_NON_DEEPLINK_ROUTES = ['community', '~'];
+
 export function DeepLinkRegistrar() {
   const relayEnvironment = useRelayEnvironment();
   const track = useTrack();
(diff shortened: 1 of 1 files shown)


apps/mobile/src/components/DeepLinkRegistrar.tsx
```diff
@@ -7,141 +7,154 @@ import { simpleHash } from 'src/utils/hashString';
...
     async function handleDeepLinkToUrl({ url }: { url: string }) {
-      track('Deep Link Opened', { url });
-
-      const response = await fetchQuery<DeepLinkRegistrarAuthCheckQuery>(
-        relayEnvironment,
-        graphql`
-          query DeepLinkRegistrarAuthCheckQuery {
-            viewer {
-              __typename
+      try {
```
Robinnnnn, line 22: this file looks like a big diff but i only wrapped it in a try/catch

kaitoo1:
noticed that for routes we have but don't support deep linking for, the deep link registrar treated the url params as a possible username, collectionId, etc which led to the app navigating and failing to load user profile/gallery screens when the url was actually for `gallery.so/~/compose` or a collection page. 
so I added a check to return early from the deep link registrar for routes that we know we don't support deep linking for, so that the app doesn't try to check if it's a user or gallery

before
gallery.so/~/ link navigates to gallery collection page 
https://github.com/gallery-so/gallery/assets/80802871/6369fee3-5e5c-4945-9645-9699071fd10d


after:
video of links opening the app without crashing
https://github.com/gallery-so/gallery/assets/80802871/5246cd84-a760-440e-bcb0-08112c025667

kaitoo1 (approved):

PR #1970, opened 2023-09-28
[GAL-4387] Clear up provider after connect wallet

jakzaizzat:
### Summary of Changes

Fix a bug where the wallet connect still ask for signature when re-open the apps. 

**Solution**
We cleared up the web3 provider after the user attempt to sign the signature message (success or failed)

### Edge Cases

1. Sign with with wallet connect -> Close app -> Re-open the app again 
2. Add wallet to existing user -> Close app -> Re-open the app again

### Testing Steps

1. Sign with with wallet connect -> Close app -> Re-open the app again 

### Checklist

### DIFF
--- a/apps/mobile/src/contexts/ManageWalletContext.tsx
+++ b/apps/mobile/src/contexts/ManageWalletContext.tsx
@@ -99,6 +99,11 @@ const ManageWalletProvider = memo(({ children }: Props) => {
     bottomSheet.current?.dismiss();
   }, []);
 
+  const clearState = useCallback(() => {
+    setIsSigningIn(false);
+    provider?.disconnect();
+  }, [provider]);
+
   const handleSignMessage = useCallback(async () => {
     if (!web3Provider || !address || hasSigned.current) {
       return;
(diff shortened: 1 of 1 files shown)


apps/mobile/src/contexts/ManageWalletContext.tsx
```diff
@@ -99,6 +99,11 @@ const ManageWalletProvider = memo(({ children }: Props) => {
     bottomSheet.current?.dismiss();
   }, []);
 
+  const clearState = useCallback(() => {
```
kaitoo1, line 102: it still seems a bit odd to me that we call `setIsSigningIn(false)` and `provider?.disconnect();` when the app is opened 🤔 

I would expect the opposite where signing in is false by default and provider is disconnected, and they get set to true + connected when the user triggers a sign in or wallet connect flow

jakzaizzat: @kaitoo1 Make sense! I think I need to rename the `isSigningIn` naming.

The current `isSigninIn` state is to describe that the user is in the progress of sign in (not failed or success).

kaitoo1 (approved):

PR #1986, opened 2023-10-04
GAL-4161 refactor how comments are rendered to solve wrapping

kaitoo1:
### Summary of Changes

This PR refactors how comments are rendered. Previously we used the Markdown to display the comment with links.

Problem:
For comments specifically, (vs other text like the Post caption), we don’t need to support markdown. Instead we only need to support links and mentions. The Markdown component introduce unnecessary complexity and rigidity because it handles a lot of other cases that we don't need to for displaying comments. Specifically, the Markdown component prevented us from being able to wrap the comment under the username in the preview.

Solution
This PR adds a new `ProcessedCommentText` that we'll now use instead of `Markdown` to render comments. This component will parse the raw comment for any urls and display them as tappable links. In the future, this can be updated to also detect Mentions.

I also changed the layout of the "Socialize" section where we display admires and comments, so that the comment line has space to take the full screen width. 

### Demo or Before/After Pics

|CleanShot 2023-10-04 at 18 34 44@2x|CleanShot 2023-10-05 at 13 25 07@2x|

### Edge Cases
- short comments, long comments that wrap
- should look good on both the Comment Preview on the feed, as well as comment list in bottom sheet

### Testing Steps
Enter a long comment that wraps 

### Checklist

### DIFF
--- a/apps/mobile/src/components/Feed/CommentsBottomSheet/CommentsBottomSheetLine.tsx
+++ b/apps/mobile/src/components/Feed/CommentsBottomSheet/CommentsBottomSheetLine.tsx
@@ -1,19 +1,19 @@
 import { useNavigation } from '@react-navigation/native';
 import { useCallback, useRef, useState } from 'react';
-import { StyleSheet, View } from 'react-native';
+import { View } from 'react-native';
 import { useFragment } from 'react-relay';
 import { graphql } from 'relay-runtime';
 
-import { WarningLinkBottomSheet } from '~/components/Feed/Posts/WarningLinkBottomSheet';
 import { GalleryBottomSheetModalType } from '~/components/GalleryBottomSheet/GalleryBottomSheetModal';
 import { GalleryTouchableOpacity } from '~/components/GalleryTouchableOpacity';
-import { Markdown } from '~/components/Markdown';
 import { ProfilePicture } from '~/components/ProfilePicture/ProfilePicture';
 import { Typography } from '~/components/Typography';
 import { CommentsBottomSheetLineFragment$key } from '~/generated/CommentsBottomSheetLineFragment.graphql';
 import { MainTabStackNavigatorProp } from '~/navigation/types';
 import { replaceUrlsWithMarkdownFormat } from '~/shared/utils/replaceUrlsWithMarkdownFormat';
 import { getTimeSince } from '~/shared/utils/time';
+
+import ProcessedCommentText from '../Socialize/ProcessedCommentText';
 
 type CommentLineProps = {
   commentRef: CommentsBottomSheetLineFragment$key;
@@ -39,22 +39,16 @@ export function CommentsBottomSheetLine({ commentRef }: CommentLineProps) {
   const timeAgo = getTimeSince(comment.creationTime);
   const navigation = useNavigation<MainTabStackNavigatorProp>();
 
-  const [redirectUrl, setRedirectUrl] = useState('');
-  const captionWithMarkdownLinks = replaceUrlsWithMarkdownFormat(comment.comment ?? '');
-
-  const bottomSheetRef = useRef<GalleryBottomSheetModalType | null>(null);
-
-  const handleLinkPress = useCallback((url: string) => {
-    bottomSheetRef.current?.present();
-    setRedirectUrl(url);
-  }, []);
-
   const handleUserPress = useCallback(() => {
     const username = comment?.commenter?.username;
     if (username) {
       navigation.push('Profile', { username: username, hideBackButton: false });
     }
   }, [comment?.commenter?.username, navigation]);
+
+  if (!comment.comment) {
+    return null;
+  }
 
   return (
     <GalleryTouchableOpacity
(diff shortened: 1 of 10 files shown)


apps/mobile/src/components/Feed/CommentsBottomSheet/CommentsBottomSheetLine.tsx
```diff
@@ -80,18 +74,10 @@ export function CommentsBottomSheetLine({ commentRef }: CommentLineProps) {
             {timeAgo}
           </Typography>
         </View>
-        <View className="flex">
-          <Markdown onBypassLinkPress={handleLinkPress} style={markdownStyles}>
-            {captionWithMarkdownLinks}
-          </Markdown>
-          <WarningLinkBottomSheet redirectUrl={redirectUrl} ref={bottomSheetRef} />
-        </View>
+        {/* <View className="flex"> */}
```
kaitoo1, line 77: need to test if its fine to remove this

apps/mobile/src/components/Feed/Socialize/ProcessedCommentText.tsx
```diff
@@ -0,0 +1,63 @@
...
+      if (urls && urls[index]) {
+        result.push(<LinkComponent key={`link-${index}`} url={urls[index] ?? ''} />);
+      }
+    });
+
+    return result;
+  }, [comment]);
+
+  return (
+    <>
```
kaitoo1, line 53: remove

apps/mobile/src/components/UsernameDisplay.tsx
```diff
@@ -12,10 +12,9 @@ type Props = {
   userRef: UsernameDisplayFragment$key;
   style?: GalleryTouchableOpacityProps['style'];
   size?: 'xs' | 'sm';
-  text?: string;
```
kaitoo1, line 15: wasn't being used

apps/mobile/src/components/Feed/Socialize/Comments.tsx
```diff
@@ -0,0 +1,53 @@
...
+export default function Comments({ commentRefs, totalComments, onCommentPress }: Props) {
+  const comments = useFragment(
+    graphql`
+      fragment CommentsFragment on Comment @relay(plural: true) {
+        dbid
+        ...CommentLineFragment
+      }
+    `,
+    commentRefs
+  );
```
jakzaizzat, line 26: Since we fetch the comments inside this component, can we remove the `totalComments` props?

apps/mobile/src/components/Feed/Socialize/ProcessedCommentText.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const bottomSheetRef = useRef<GalleryBottomSheetModalType | null>(null);
+  const handleLinkPress = useCallback(() => {
+    bottomSheetRef.current?.present();
+  }, []);
+
+  return (
+    <>
+      <Text className="text-shadow" onPress={handleLinkPress}>
+        {url}
+      </Text>
```
jakzaizzat, line 24: Should we wrap this with `TouchableOpacity` or  `Pressable`?

apps/mobile/src/components/Feed/Socialize/ProcessedCommentText.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+    chunks.forEach((chunk, index) => {
+      result.push(<Text key={`text-${index}`}>{chunk}</Text>);
+      if (urls && urls[index]) {
+        result.push(<LinkComponent key={`link-${index}`} url={urls[index] ?? ''} />);
+      }
+    });
+
+    return result;
+  }, [comment]);
```
jakzaizzat, line 50: 😍

apps/mobile/src/components/Feed/Socialize/Comments.tsx
kaitoo1: unfortunately I don't think we can because this fragment is on the edge node, but the total comments comes from the pageInfo.total field 

that lets us get the total number of comments not just the number of comments loaded so far

apps/mobile/src/components/Feed/Socialize/ProcessedCommentText.tsx
kaitoo1: unfortunately when I use either it breaks the wrapping and I wasn't able to find tailwind classes that would make it behave like Text :(

I figured since we don't need to show any interaction effect in this case (since we immediately show the bottom sheet), the trade off is worth it

without pressable/touchable opacity
CleanShot 2023-10-05 at 21 04 16@2x

with it
CleanShot 2023-10-05 at 21 04 22@2x

kaitoo1: good catch though

apps/mobile/src/components/Feed/Socialize/Comments.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const previewComments = comments.slice(-1);
```
Robinnnnn, line 28: let's make sure to memo this! unmemoized arrays + RN = 💥

Robinnnnn (approved):

PR #2051, opened 2023-10-26
Add collection related links

Rohan-cp:
### Summary of Changes

### Demo Mobile

| Screenshot 2023-11-02 at 3 30 41 AM | Screenshot 2023-11-02 at 3 23 34 AM |

| After                                     | After    (clicking on opensea icon)                                  |
| Screenshot 2023-11-02 at 3 21 16 AM | Screenshot 2023-11-02 at 3 21 33 AM |

### Demo Web (After)
Screenshot 2023-11-02 at 4 12 04 AM
Screenshot 2023-11-02 at 4 13 18 AM

### Edge Cases

### Testing Steps

### Checklist


packages/shared/src/utils/extractRelevantMetadataFromCommunity.ts
```diff
@@ -0,0 +1,52 @@
...
+  const contractAddress = contract?.contractAddress?.address;
+  const result = {
+    contractAddress,
+    openseaUrl: '',
+    objktUrl: '',
+    externalAddressUrl: '',
+  };
+
+  if (contractAddress) {
+    result.objktUrl = getObjktExternalUrlDangerouslyForCollection(contractAddress);
```
Robinnnnn, line 39: we only want to provide the `objktUrl` for tezos right?

i think we also want to support Prohibition and FxHash URLs, similar to how we support those in `extractRelevantMetadataFromToken`. can be a follow-up PR though as we'll want the icons to go along with it.

image

Robinnnnn (approved):

packages/shared/src/utils/extractRelevantMetadataFromCommunity.ts
Rohan-cp: 1. so I added that logic to check the chain where we actually display the icon but I guess it makes more sense to have that here, so will change 
2. we don't currently support sub-communities in prohibition so there is no real prohibition or fxhash URL for the Prohibition community right?
3. web currently already shows projectUrl (globe icon), will add the openseaUrl and objktUrl stuff

Robinnnnn (approved):

PR #2077, opened 2023-11-06
Add creator announcement to composer on web

kaitoo1:
### Summary of Changes
This PR adds a Gallery for Creators is now in beta announcement to the composer on web.
CleanShot 2023-11-06 at 13 39 48@2x

This is shown when the user selects the "Created" filter and hasn't been shown this announcement before. 
Once they click continue, this state is saved in local storage.

### Demo or Before/After Pics
CleanShot 2023-11-06 at 13 39 48@2x

### Edge Cases
- one edge case is an existing creator who already uses this feature. they'll see this announcement once, but that's ok.

### Testing Steps
Open the composer and select the Created filter.

### Checklist


apps/web/src/components/NftSelector/NftSelector.tsx
```diff
@@ -350,20 +350,18 @@ function NftSelectorInner({ onSelectToken, headerText, preSelectedContract, even
...
-          selectedContractAddress={selectedContract?.address ?? null}
-          onSelectContract={handleSelectContract}
-          onSelectToken={onSelectToken}
-          eventFlow={eventFlow}
-          selectedNetworkView={network}
-          hasSearchKeyword={isSearching}
-          handleRefresh={handleRefresh}
-        />
-      )}
+      <NftSelectorTokens
```
kaitoo1, line 353: moved the loading state and selector view into a single component, so that I could return early with the announcement view in certain cases

apps/web/src/components/NftSelector/NftSelectorTokens.tsx
```diff
@@ -0,0 +1,118 @@
...
+    CREATOR_BETA_ANNOUNCEMENT_SEEN,
+    'false'
+  );
+
+  const handleContinueCreatorBetaClick = useCallback(() => {
+    setCreatorBetaAnnouncementSeen('true');
+  }, [setCreatorBetaAnnouncementSeen]);
+
+  const showCreatorBetaAnnouncement = useMemo(() => {
+    console.log(selectedFilter, creatorBetaAnnouncementSeen);
```
kaitoo1, line 63: oops toss

Robinnnnn, line 63: toss

apps/web/src/components/NftSelector/NftSelectorTokens.tsx
```diff
@@ -0,0 +1,118 @@
...
+    tokenRefs
+  );
+
+  const [creatorBetaAnnouncementSeen, setCreatorBetaAnnouncementSeen] = usePersistedState(
+    CREATOR_BETA_ANNOUNCEMENT_SEEN,
+    'false'
+  );
+
+  const handleContinueCreatorBetaClick = useCallback(() => {
+    setCreatorBetaAnnouncementSeen('true');
```
Robinnnnn, line 59: nit: you can just use raw `false` and `true` values without wrapping it in a string; `usePersistedState` will handle stringification

Robinnnnn: non-blocker: would prefer to do this via backend feature flag, unless we won't have announcement up for a while

apps/web/src/components/NftSelector/NftSelectorTokens.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }, [creatorBetaAnnouncementSeen, selectedFilter]);
+
+  if (showCreatorBetaAnnouncement) {
+    return (
+      <StyledAnnouncement align="center" justify="center" gap={24}>
+        <StyledTitle>Gallery for Creators is now in beta</StyledTitle>
+        <BaseM>
+          Welcome to our new creator support feature, currently in beta. Share and display works
+          you’ve created on Gallery. Learn more about how it works{' '}
+          <GalleryLink href={'https://gallery.so'}>in our FAQ here</GalleryLink>.
```
Robinnnnn, line 74: nit: would prefer to link this in a couple other ways
- `works in <our FAQ>.`
- or `works in our FAQ <here>.`

Robinnnnn (approved):

kaitoo1:
added announcement to Editor as well
https://github.com/gallery-so/gallery/assets/80802871/100d72e0-62f3-493b-addd-575c62c23c25

PR #2091, opened 2023-11-13
[GAL-4757] Fix duplicate mention tagging

jakzaizzat:
### Summary of Changes

Fix duplicate mention tagging. Found out at the issue happens when there is multiple `mention` interval

### Demo or Before/After Pics

| image | CleanShot 2023-11-13 at 18 11 24@2x |
| image | CleanShot 2023-11-13 at 18 12 35@2x|

### Testing Steps

- Open any post/comment that has multiple tagging

### Checklist

### DIFF
--- a/packages/shared/src/components/GalleryProccessedText/GalleryTextElementParser.ts
+++ b/packages/shared/src/components/GalleryProccessedText/GalleryTextElementParser.ts
@@ -1,3 +1,5 @@
+import isEqual from 'lodash/isEqual';
+import uniqWith from 'lodash/uniqWith';
 import { graphql, readInlineData } from 'react-relay';
 
 import { GalleryProcessedTextFragment$data } from '~/generated/GalleryProcessedTextFragment.graphql';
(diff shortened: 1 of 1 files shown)


packages/shared/src/components/GalleryProccessedText/GalleryTextElementParser.ts
```diff
@@ -51,12 +53,15 @@ export function getMentionElements(
...
 
-  const mentions: GalleryTextElementParserMentionsFragment$data[] = [];
+  let mentions: GalleryTextElementParserMentionsFragment$data[] = [];
 
   mentionRefs.forEach((mentionRef) => {
     mentions.push(fetchMention(mentionRef));
   });
 
+  // Remove duplicate mentions
+  mentions = uniqWith(mentions, isEqual);
```
kaitoo1, line 63: do you know what causes duplicate mentions? I wonder if there is a separate bug that causes duplicate mentions in the first place that we should also fix

jakzaizzat: @kaitoo1 Not sure 🤔 I'm currently to find what cause the issue. There must be something that triggered it

kaitoo1: replied on slack!

kaitoo1 (approved):

PR #2100, opened 2023-11-14
Upgrade Moon

Robinnnnn:
### Summary of Changes

s/o @milesj for the assist

### Checklist
- [x] Web app still works
- [x] Mobile app still works

### DIFF
--- a/.moon/toolchain.yml
+++ b/.moon/toolchain.yml
@@ -14,10 +14,6 @@ node:
   # Accepts "npm" (default), "pnpm", or "yarn".
   packageManager: 'yarn'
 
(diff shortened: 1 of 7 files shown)


.moon/toolchain.yml
```diff
@@ -14,10 +14,6 @@ node:
   # Accepts "npm" (default), "pnpm", or "yarn".
   packageManager: 'yarn'
 
-  # The version of the package manager (above) to use.
```
milesj, line 17: you can probably bring this back now with the moon upgrade

Robinnnnn: great call, looks like it works now!

PR #2105, opened 2023-11-16
Update how we write/store mention

jakzaizzat:
### Summary of Changes

Improve on how we calculate the positioning of `Mention`. 

Previously, we are tracking on every keystroke and update the position manually. Its getting tricky to make it perfect and we didn't have much confident that it will works 100%. 

Current solution would be use regex to the whole string and get the index for each mention.

### Demo or Before/After Pics

https://github.com/gallery-so/gallery/assets/4480258/5579aadb-6594-4281-bbe7-5c061022fa08

### Edge Cases

### Testing Steps

### Checklist

### DIFF
--- a/packages/shared/src/hooks/useMentionableMessage.ts
+++ b/packages/shared/src/hooks/useMentionableMessage.ts
@@ -136,7 +136,7 @@ export function useMentionableMessage() {
   };
 }
 
-function updateMentionPositions(text: string, mentions: MentionDataType[]) {
+export function updateMentionPositions(text: string, mentions: MentionDataType[]) {
   const updatedMentions: MentionDataType[] = [];
   mentions.forEach((mention) => {
     // use regex to find all mentions in text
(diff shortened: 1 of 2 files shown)


packages/shared/src/hooks/useMentionableMessage.ts
```diff
@@ -172,16 +115,45 @@ export function useMentionableMessage() {
...
+    const regex = new RegExp(`@${mention.label}`, 'g');
+    let matches;
+
+    while ((matches = regex.exec(text))) {
+      const label = mention?.label || '';
+      updatedMentions.push({
+        ...mention,
+        interval: {
+          start: matches.index || 0,
+          length: label.length + 1, // +1 for the @
```
kaitoo1, line 152: good comment

kaitoo1 (approved):
NICE ONE

Robinnnnn (approved):

PR #2183, opened 2023-12-13
Fix feed randomly refetch on comment

jakzaizzat:
### Summary of Changes

- Fix optimistic update whenever user comment
- Fix random refresh feed whenever the user comment on any post

This bug only happens if the user comment on post (not reply). This line triggered the refresh

https://github.com/gallery-so/gallery/blob/main/apps/web/src/components/Feed/Socialize/CommentsModal/PostCommentsModal.tsx#L61

### Demo or Before/After Pics

**Optimistic Update**

https://github.com/gallery-so/gallery/assets/4480258/68742bd8-86c9-4f32-88a1-aa519cad0a46

**Random Feed refetch**

https://github.com/gallery-so/gallery/assets/4480258/3bf8cc74-bd1e-42c4-ac8a-d0e53b7eeb1f

### Edge Cases

### Testing Steps

### Checklist

### DIFF
--- a/apps/web/src/utils/useClearURLQueryParams.ts
+++ b/apps/web/src/utils/useClearURLQueryParams.ts
@@ -18,13 +18,18 @@ export function useClearURLQueryParams(param: string | string[]) {
     if (hasRenderedOnce.current) return;
     const params = new URLSearchParams(urlQuery as Record<string, string>);
     const paramsToClear = typeof param === 'string' ? [param] : param;
+    let paramDeleted = false;
     for (const p of paramsToClear) {
       if (params.has(p)) {
         params.delete(p);
+        paramDeleted = true;
       }
     }
-    // @ts-expect-error we're simply replacing the current page with the same path
-    replace({ pathname, query: params.toString() }, undefined, { shallow: true });
+    if (paramDeleted) {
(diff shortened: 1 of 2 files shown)


apps/web/src/utils/useClearURLQueryParams.ts
```diff
@@ -18,13 +18,18 @@ export function useClearURLQueryParams(param: string | string[]) {
...
+        paramDeleted = true;
       }
     }
-    // @ts-expect-error we're simply replacing the current page with the same path
-    replace({ pathname, query: params.toString() }, undefined, { shallow: true });
+    if (paramDeleted) {
+      // @ts-expect-error we're simply replacing the current page with the same path
+      replace({ pathname, query: params.toString() }, undefined, { shallow: true });
+    }
+
```
jakzaizzat, line 32: Add a new logic that we only replace the router if we deleted something from the `params`. 

Lmk what you think about this @Robinnnnn

Robinnnnn: wow, i'm super surprised this caused an issue 🤔 i would assume shallow-rendering wouldn't cause the page to rerender like that

great catch!

Robinnnnn (approved):

PR #2186, opened 2023-12-13
[GAL-3622] nft detail preloader on web

Rohan-cp:
### Summary of Changes

### Demo 

https://github.com/gallery-so/gallery/assets/49758803/8669e640-4b6c-4ccd-917b-8d8178175e1c

### Edge Cases

### Testing Steps

### Checklist


apps/web/src/components/NftPreview/NftPreviewAsset.tsx
```diff
@@ -16,17 +21,53 @@ function NftPreviewAsset({ tokenRef, onLoad }: Props) {
...
+          width: serverSourcedDimensions.width,
+          height: serverSourcedDimensions.height,
+        },
+      });
+    }
+
+    return {
+      height: DESKTOP_TOKEN_SIZE,
+      width: DESKTOP_TOKEN_SIZE,
+    };
```
Robinnnnn, line 55: what happens for mobile?

apps/web/src/contexts/fullPageNftDetailModalListener/FullPageNftDetailModalListener.tsx
```diff
@@ -41,7 +41,7 @@ export default function FullPageNftDetailModalListener() {
     }
 
     const content = collectionId ? (
-      <Suspense fallback={<NftDetailPageFallback />}>
```
Robinnnnn, line 44: instead of having to do `tokenId as string` here, on L39 let's update:
```ts
if (!username) {
  return
}
if (typeof tokenId !== 'string) {
  return
}
```

then you should safely be able to pass in `tokenId={tokenId}` without type casting

apps/web/src/contexts/nftPreviewFallback/NftPreviewFallbackContext.tsx
```diff
@@ -0,0 +1,80 @@
+import { createContext, ReactNode, useCallback, useContext, useMemo, useState } from 'react';
+
+export type Dimensions = {
+  width?: number;
+  height?: number;
+};
+
+export type CacheParams = {
+  [tokenId: string]: {
+    type: 'preview' | 'raw';
```
Robinnnnn, line 10: `'preview' | 'raw'` is repeated a few times here. let's assign it to its own type, like `LoadedNftAssetType`, and use that variable everywhere (including L51)

apps/web/src/scenes/NftDetailPage/NftDetailAsset.tsx
```diff
@@ -28,6 +32,8 @@ type NftDetailAssetComponentProps = {
   onLoad: () => void;
 };
 
+const DESKTOP_TOKEN_SIZE = 600;
```
Robinnnnn, line 35: 1. noticed this is defined in another file as well; can we define it one place and export it?
2. probably need mobile support as well

apps/web/src/scenes/NftDetailPage/NftDetailAsset.tsx
```diff
@@ -201,33 +210,73 @@ function NftDetailAsset({ tokenRef, hasExtraPaddingForNote }: Props) {
...
+          height: serverSourcedDimensions.height,
+        },
+      });
+    }
+
+    return {
+      height: DESKTOP_TOKEN_SIZE,
+      width: DESKTOP_TOKEN_SIZE,
+    };
+  }, [token.media?.dimensions]);
```
Robinnnnn, line 270: if this ends up getting more complicated to incorporate mobile dimensions, let's turn it into its own reusable hook so the other file can use it too

apps/web/src/scenes/NftDetailPage/NftDetailPageFallback.tsx
```diff
@@ -3,12 +3,35 @@ import styled from 'styled-components';
...
     <StyledFullPageLoader>
-      <StyledNftSkeleton />
-      <VStack gap={24}>
+      {hasPreviewUrl ? (
+        <VisibilityContainer>
+          <StyledImageWrapper className={'visible'}>
+            <StyledImage
+              src={cachedUrls[tokenId]?.url}
+              height={dimensions?.height ?? 600}
+              width={dimensions?.width ?? 600}
```
Robinnnnn, line 26: 1. should handle mobile
2. let's import the `600` from wherever we define the const

apps/web/src/scenes/TokenDetailPage/TokenDetailAsset.tsx
```diff
@@ -1,17 +1,23 @@
...
+import { useNftPreviewFallbackState } from '~/contexts/nftPreviewFallback/NftPreviewFallbackContext';
 import { TokenDetailAssetFragment$key } from '~/generated/TokenDetailAssetFragment.graphql';
 import { useNftRetry } from '~/hooks/useNftRetry';
 import { useBreakpoint } from '~/hooks/useWindowSize';
 import { NftDetailAssetComponent } from '~/scenes/NftDetailPage/NftDetailAsset';
+import { useGetSinglePreviewImage } from '~/shared/relay/useGetPreviewImages';
+import { fitDimensionsToContainerContain } from '~/shared/utils/fitDimensionsToContainer';
 import { getBackgroundColorOverrideForContract } from '~/utils/token';
 
+const DESKTOP_TOKEN_SIZE = 600;
```
Robinnnnn, line 19: same comment

apps/web/src/scenes/TokenDetailPage/TokenDetailAsset.tsx
```diff
@@ -30,33 +36,71 @@ function TokenDetailAsset({ tokenRef, hasExtraPaddingForNote }: Props) {
...
+          height: serverSourcedDimensions.height,
+        },
+      });
+    }
+
+    return {
+      height: DESKTOP_TOKEN_SIZE,
+      width: DESKTOP_TOKEN_SIZE,
+    };
+  }, [token.media?.dimensions]);
```
Robinnnnn, line 87: same comment to standardize into a reusable hook

apps/web/src/components/NftPreview/NftPreviewAsset.tsx
Rohan-cp: fixed now

apps/web/src/scenes/NftDetailPage/NftDetailAsset.tsx
Rohan-cp: yess done now

apps/web/src/scenes/NftDetailPage/NftDetailAsset.tsx
Rohan-cp: it became a bit more complicated (1 line more), but I don't think making a reusable hook is the best because we can't pass `serverSourcedDimensions` directly into the hook since we'd need to do the if check (on L256) before anyway and create a new object that's a copy of it (as source), so there's no real part that's reusable I think.

apps/web/src/scenes/TokenDetailPage/TokenDetailAsset.tsx
Rohan-cp: replied above

apps/web/src/scenes/NftDetailPage/NftDetailPageFallback.tsx
```diff
@@ -22,26 +55,50 @@ export default function NftDetailPageFallback() {
...
 
 const StyledFullPageLoader = styled.div`
   display: flex;
-  gap: 60px;
   height: 100vh;
   width: 100%;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  flex-direction: row;
```
kaitoo1, line 70: flex-direction is defined twice, guessing we only want `flex-direction: row;`?

apps/web/src/scenes/NftDetailPage/NftDetailPageFallback.tsx
```diff
@@ -22,26 +55,50 @@ export default function NftDetailPageFallback() {
...
+
+  @media only screen and (max-width: ${size.tablet}px) {
```
kaitoo1, line 72: lets use mobile first responsive styling for consistency, that way it's easier to reason about when the style gets applied

so each `@media only...` will apply styles for the defined screen size or larger

apps/web/src/scenes/NftDetailPage/NftDetailPageFallback.tsx
```diff
@@ -1,14 +1,47 @@
...
+    : DESKTOP_TOKEN_DETAIL_VIEW_SIZE;
 
-export default function NftDetailPageFallback() {
   return (
     <StyledFullPageLoader>
-      <StyledNftSkeleton />
-      <VStack gap={24}>
+      {hasPreviewUrl ? (
+        <VisibilityContainer>
+          <StyledImageWrapper className={'visible'}>
```
kaitoo1, line 32: we usually use props to pass in a styled state, ie `isVisible={isVisible}` , is there a special need to use a classname here?

kaitoo1 (approved):

Robinnnnn (approved):

PR #2221, opened 2024-01-03
GAL-5015 wrap nft detail preview in a tag

kaitoo1:
### Summary of Changes

Make it possible to command-click an nft on the feed or gallery to open in new tab by wrapping it in an a tag

### Demo or Before/After Pics

command-click opens in new tab, regular click opens in same window

https://github.com/gallery-so/gallery/assets/80802871/d9d882d0-79c8-4663-a2e1-33aea9a8a291

### Edge Cases
this still won't work for iframes because they don't register the click to open the nft detail page in the first place. this change only applies to nfts that already open an nft detail page when clicked

### Testing Steps
command click an nft on the feed or in a gallery

### Checklist

### DIFF
--- a/apps/web/src/scenes/NftDetailPage/LinkToFullPageNftDetailModal.tsx
+++ b/apps/web/src/scenes/NftDetailPage/LinkToFullPageNftDetailModal.tsx
@@ -108,7 +108,7 @@ export default function LinkToFullPageNftDetailModal({
   );
 }
 
-const StyledAnchor = styled.div`
+const StyledAnchor = styled.a`
   display: flex;
   justify-content: center;
   align-items: center;


apps/web/src/scenes/NftDetailPage/LinkToFullPageNftDetailModal.tsx
```diff
@@ -108,7 +108,7 @@ export default function LinkToFullPageNftDetailModal({
   );
 }
 
-const StyledAnchor = styled.div`
+const StyledAnchor = styled.a`
```
kaitoo1, line 111: not sure why this was a div in the first place - maybe changing it an `a` tag breaks something im not aware of?

kaitoo1: does it mess with the Next Link ? 🤔

Robinnnnn (approved):

PR #2256, opened 2024-01-21
vercel og image generation

Rohan-cp:
### Summary of Changes

### Demo or Before/After Pics

|  |  |

### Edge Cases

### Testing Steps

### Checklist

### DIFF
--- a/apps/web/pages/api/og.tsx
+++ b/apps/web/pages/api/og.tsx
@@ -1,16 +1,37 @@
 import { ImageResponse } from '@vercel/og';
 import { NextRequest } from 'next/server';
+import { ogPostQuery } from '~/generated/ogPostQuery.graphql';
+import { graphql, useLazyLoadQuery } from 'react-relay';
 
 import styled from 'styled-components';
 import { TITLE_FONT_FAMILY } from '~/components/core/Text/Text';
 import colors from '~/shared/theme/colors';
+import { HEIGHT_OPENGRAPH_IMAGE, WIDTH_OPENGRAPH_IMAGE } from '~/constants/opengraph';
+import { fetchWithJustQueryText } from 'shared/relay/network';
+import { postIdQuery } from './postIdQuery';
 
 export const config = {
   runtime: 'edge',
 };
 
+const ABCDiatypeRegular = fetch(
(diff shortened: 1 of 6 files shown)


apps/web/pages/api/og.tsx
```diff
@@ -0,0 +1,291 @@
...
+    }
+  }
+
+  postImageUrl = postPreviewUrls?.large ?? '';
+
+  const ABCDiatypeRegularFontData = await ABCDiatypeRegular;
+  const ABCDiatypeBoldFontData = await ABCDiatypeBold;
+  const alpinaLightFontData = await alpinaLight;
+
+  return new ImageResponse(
```
Robinnnnn, line 125: lower priority - let's wrap this whole thing in a try/catch and in the catch return a new `ImageResponse` featuring the gallery logo as an SVG (this should be in our asset folders in the gallery repo)

we should also throw an error somewhere in here if the media isn't available. if so, `catch` will handle it and render the backup

apps/web/pages/api/og.tsx
```diff
@@ -0,0 +1,291 @@
...
+  new URL('../../../../packages/shared/src/fonts/GT-Alpina-Standard-Light.otf', import.meta.url)
+).then((res) => res.arrayBuffer());
+
+type UrlSet = {
+  small: string | null;
+  medium: string | null;
+  large: string | null;
+};
+
+export default async function handler(request: NextRequest) {
```
Robinnnnn, line 36: we should be able to stick this into `/api/og/post/[postId]`

and then grab the postId from `req.query` or `req.params` or `req.nextUrl`

this makes it so that when redirecting users here, we don't have to use:
```
gallery-og.vercel.app/api/og/post?postId=xxx
```

and instead
```
gallery-og.vercel.app/api/og/post/xxx
```

Robinnnnn: `curl http://localhost:3001/api/og/post/12345` -> should test a very basic handler to ensure this works

file should be placed in `/api/og/post/[postId].tsx`

Robinnnnn: collection handler would be placed in `/api/og/collection/[collectionId].tsx`

PR #2215, opened 2023-12-28
add suggested users to follow in onboarding

Rohan-cp:
### Summary of Changes
As part of onboarding a user, a screen with suggested users to follow are shown that can be skipped

### Demo

https://github.com/gallery-so/gallery/assets/49758803/cba6af09-a4c6-4dee-96d1-8af243246c9c

### Edge Cases

List any common edge cases that you have considered and tested.

### Testing Steps
Can test locally on mobile

### Checklist

Please make sure to review and check all of the following:

- [x] I've tested the changes and all tests pass.
- [ ] (if web) I've tested the changes on various desktop screen sizes to ensure responsiveness.
- [x] (if mobile) I've tested the changes on both light and dark modes.

PR #2266, opened 2024-01-24
[GAL-5069] UserFollowList regression

Rohan-cp:
### Summary of Changes

1. fix a regression in UserFollowList from the recent pr #2215 where there is no horizontal padding for the case `isPresentational = false`
2. in the UserFollowList we should display user bios as plaintext without markdown formatting or styling

### Demo or Before/After Pics

| Screenshot 2024-01-23 at 11 01 23 PM | Screenshot 2024-01-23 at 10 17 42 PM |

## Change 2
Now clicking on the link in the user list, takes you to the user profile and does not redirect you to the clicked URL. Also, the link styled with markdown is the same color as rest of the plaintext and does not appear as a link anymore.
### Before
https://github.com/gallery-so/gallery/assets/49758803/a1fa4f32-8903-45a5-b29b-a05fe1a20a60

### After
https://github.com/gallery-so/gallery/assets/49758803/365618b9-5ab3-4e63-84ee-5f4e5e3e84a9

### Edge Cases

### Testing Steps

### Checklist


apps/mobile/src/components/ProcessedText/ProcessedTextAsPlaintext.tsx
```diff
@@ -0,0 +1,35 @@
...
+import GalleryProcessedText from '~/shared/components/GalleryProccessedText/GalleryProcessedText';
+
+import { MentionComponent } from './elements/MentionComponent';
+import { TextComponent } from './elements/TextComponent';
+
+type ProcessedTextProps = {
+  text: string;
+} & TextProps;
+
+export default function ProcessedTextAsPlaintext({ text, ...props }: ProcessedTextProps) {
```
Robinnnnn, line 13: looks like this is mostly the same as the existing `ProcessedText` component.

could we instead use that component and add an existing prop, like `plaintextOnly`?

Robinnnnn (approved):

PR #2269, opened 2024-01-25
Community Grid View (Web)

jakzaizzat:
### Summary of Changes

- Add grid token view to the collection page
- Clean up the specific contract that we enabled grid view previously

### Demo or Before/After Pics

**Desktop**

https://github.com/gallery-so/gallery/assets/4480258/7870fd94-b194-43f4-9e49-7c71b6dae448

**Mobile**

image

### Edge Cases

1. Tested with svg collection
2. Tested with the different aspect ration nft

### Testing Steps

### Checklist

### DIFF
--- a/apps/web/src/components/CommunityHolderGrid/CommunityHolderGridItem.tsx
+++ b/apps/web/src/components/CommunityHolderGrid/CommunityHolderGridItem.tsx
@@ -1,33 +1,38 @@
-import { useCallback } from 'react';
 import { graphql, useFragment } from 'react-relay';
 import styled from 'styled-components';
 
 import breakpoints from '~/components/core/breakpoints';
-import GalleryLink from '~/components/core/GalleryLink/GalleryLink';
 import { HStack, VStack } from '~/components/core/Spacer/Stack';
 import { BaseM } from '~/components/core/Text/Text';
-import { useModalActions } from '~/contexts/modal/ModalContext';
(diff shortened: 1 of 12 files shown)


packages/shared/src/utils/communities.ts
```diff
@@ -1,12 +1,3 @@
-export const GRID_ENABLED_COMMUNITY_ADDRESSES = [
```
jakzaizzat, line 1: Clean up this since we enabled this layout to all community

apps/web/src/components/CommunityHolderGrid/CommunityHolderGridItem.tsx
```diff
@@ -94,21 +89,27 @@ export default function CommunityHolderGridItem({ holderRef, queryRef }: Props)
       </StyledGalleryLink>
```
jakzaizzat, line 89: Still remain to render as `image` for all the tokens. Tried to replace with live render and seems like its consume too much memory. Especially for svg token

jakzaizzat: Updated it to use `NftPreview` component

packages/shared/src/utils/communities.ts
kaitoo1: good call thank you!

kaitoo1 (approved):

PR #2283, opened 2024-01-31
bug fix for tokens adjacent to broken token media in gallery

Rohan-cp:
### Summary of Changes
From a gallery, when you click on nft detail page of a token adjacent to a token with broken token media, we render an error page. This ignores the error for now so that the token clicked can render as it's media is valid

### Before

https://github.com/gallery-so/gallery/assets/49758803/0b19d422-674e-4093-ae42-eb664da70c8b

### After

https://github.com/gallery-so/gallery/assets/49758803/5cfc2d3d-b6f5-441f-91a4-aa73cf2754e9

### Edge Cases

### Testing Steps

### Checklist

### DIFF
--- a/packages/shared/src/relay/getPreviewImageUrlsInlineDangerously.ts
+++ b/packages/shared/src/relay/getPreviewImageUrlsInlineDangerously.ts
@@ -23,10 +23,6 @@ export type GetPreviewImageUrlsResult =
   | {
       type: 'error';
       error: CouldNotRenderNftError;
-    }
-  | {
-      // TODO: rohan - remove later and throw error
-      type: 'temp';
     };
 
 type Props = {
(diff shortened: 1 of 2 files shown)


packages/shared/src/relay/getPreviewImageUrlsInlineDangerously.ts
```diff
@@ -238,15 +242,7 @@ export function getPreviewImageUrlsInlineDangerously({
...
-      type: 'error',
-      error: new CouldNotRenderNftError(
-        'getPreviewImageUrlsInlineDangerously',
-        'No preview URLs returned for token!',
-        {
-          id: token?.dbid,
-          assetType: media?.__typename,
-        }
-      ),
+      type: 'temp',
```
jakzaizzat, line 245: I don't have a concrete fix yet for this issue, need to investigate it more deeper. 

Less hacky stuff could be like this. We just ignore the error only for the specific user. 

We could filter it by `username` or user id` or use `UnknownMedia` as an indicator

Example

```tsx
  if (!previewUrls && token.owner?.username === 'welovetheart') {
    return {
      type: 'ignore-invalid-media',
    };
  }

```

Robinnnnn (approved):

PR #2317, opened 2024-02-14
tab to expand mobile and scroll in web

estiventhneira:
### Summary of Changes

Adding the feature on mobile on touch to expand description, and scroll in web with overflow, for this we need to set a maximum height I use 300px but not sure if we want to use another value

### Demo or Before/After Pics

| not expanding | < <video src="https://github.com/gallery-so/gallery/assets/51178245/95aac612-e431-4a90-b325-ec575045eeb3" /> 
 <video src="https://github.com/gallery-so/gallery/assets/51178245/98711ef5-47b9-4b06-88de-86b80df48853" /> 
|

### Edge Cases

### Testing Steps

### Checklist

### DIFF
--- a/apps/mobile/src/components/Feed/Posts/PostListCaption.tsx
+++ b/apps/mobile/src/components/Feed/Posts/PostListCaption.tsx
@@ -1,17 +1,29 @@
-import { useMemo } from 'react';
+import { useMemo, useRef, useState } from 'react';
 import { View } from 'react-native';
 import { graphql, useFragment } from 'react-relay';
 
+import { GalleryTouchableOpacity } from '~/components/GalleryTouchableOpacity';
 import ProcessedText from '~/components/ProcessedText/ProcessedText';
 import { PostListCaptionFragment$key } from '~/generated/PostListCaptionFragment.graphql';
+import { GalleryElementTrackingProps } from '~/shared/contexts/AnalyticsContext';
 import { removeNullValues } from '~/shared/relay/removeNullValues';
 import { replaceUrlsWithMarkdownFormat } from '~/shared/utils/replaceUrlsWithMarkdownFormat';
 
 type Props = {
   feedPostRef: PostListCaptionFragment$key;
+  eventContext: GalleryElementTrackingProps['eventContext'];
 };
 
-export function PostListCaption({ feedPostRef }: Props) {
+export function PostListCaption({ feedPostRef, eventContext }: Props) {
+  const [showAll, setShowAll] = useState(false);
+  const ref = useRef(null);
+  const toggleText = () => {
+    if (!showAll) {
+      setShowAll(!showAll);
+    }
+  };
+
+
   const feedPost = useFragment(
     graphql`
       fragment PostListCaptionFragment on Post {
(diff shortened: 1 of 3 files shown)


apps/mobile/src/components/Feed/Posts/PostListCaption.tsx
```diff
@@ -1,17 +1,28 @@
...
 
-export function PostListCaption({ feedPostRef }: Props) {
+export function PostListCaption({ feedPostRef, eventContext }: Props) {
+  const [showAll, setShowAll] = useState(false);
+  const ref = useRef(null);
+  const toggleText = () => {
+    if (!showAll) {
+      setShowAll(!showAll);
+    }
+  };
```
Robinnnnn, line 24: some small things:
1. let's wrap this in `useCallback`! we want to memoize things as much as possible on mobile for performance
2. `setShowAll` can actually accept a function that provides the previous state as the parameter, and you can `return` the next state. in general this should make it more functional and allow the dependecy list for the callback to be stateless:

```js
// good
const toggleText = useCallback(() => {
  if (!showAll) {
    setShowAll(!showAll);
  }
}, [showAll]);

// better
const toggleText = useCallback(() => {
  setShowAll((prev) => {
    if (!prev) return !prev;
    return prev;
  })
}, []); // no more dependency
```

now that i'm looking at this logic though: what is actually happening?
- if `showAll` is `true`, then nothing happens, and `showAll` remains `true`
- if `showAll` is `false`, then we `setShowAll(!false)`, which means we set it to `true`

so in every scenario we set the value to `true`, in which case we should really just make it:

```js
const toggleText = useCallback(() => {
  setShowAll(true)
}, []);
```

you might say: well, if `showAll` is already set to `true`, do we need to actively set it to `true` again? good thing is the react engine will notice that the value hasn't changed `true -> true` and will not trigger a re-render!

apps/mobile/src/components/Feed/Posts/PostListCaption.tsx
```diff
@@ -32,7 +43,24 @@ export function PostListCaption({ feedPostRef }: Props) {
 
   return (
     <View className="px-4 pb-4">
-      <ProcessedText text={captionWithMarkdownLinks} mentionsRef={nonNullMentions} />
+      <View>
+        <GalleryTouchableOpacity
+          eventElementId="Show more lines"
+          eventName="Show more lines"
+          eventContext={eventContext || null}
```
Robinnnnn, line 50: instead of receiving `eventContext` to props of this component, let's just hard-code here:

```js
import { contexts } from '~/shared/analytics/constants';

...

eventContext={contexts.Posts}
```

apps/mobile/src/components/Feed/Posts/PostListCaption.tsx
estiventhneira: done!

apps/mobile/src/components/Feed/Posts/PostListCaption.tsx
estiventhneira: done! normally all the contexts are just hard-code? or should I get it from props normally?

PR #2356, opened 2024-03-09
add Email Notification Setting unsub options

Rohan-cp:
### Summary of Changes
You can now unsubscribe to specific mailing lists from the notification settings, and when you toggle an email notification subscription, a toast is pushed when the mutation succeeds or fails.

## After (video)
https://github.com/gallery-so/gallery/assets/49758803/11f07815-3f5d-4a10-8483-b43807ef614b

| Screenshot 2024-03-09 at 14 19 47 | Screenshot 2024-03-09 at 14 05 46 |

## After (video)
https://github.com/gallery-so/gallery/assets/49758803/e35842b3-6484-465e-8c62-d8cecd77f63e

### Edge Cases

Dark mode
Screenshot 2024-03-11 at 17 56 42

### Testing Steps

### Checklist

### DIFF
--- a/apps/mobile/src/screens/SettingsScreen/NotificationSettingsScreen.tsx
+++ b/apps/mobile/src/screens/SettingsScreen/NotificationSettingsScreen.tsx
@@ -3,10 +3,13 @@ import * as Notifications from 'expo-notifications';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { Alert, Linking, View } from 'react-native';
 import { useRelayEnvironment } from 'react-relay';
+import colors from 'shared/theme/colors';
+import { CircleCheckIcon } from 'src/icons/CircleCheckIcon';
 
 import { BackButton } from '~/components/BackButton';
 import { Button } from '~/components/Button';
 import { registerNotificationToken } from '~/components/Notification/registerNotificationToken';
+import { Toggle } from '~/components/Toggle';
 import { Typography } from '~/components/Typography';
 import { contexts } from '~/shared/analytics/constants';
 
@@ -20,6 +23,42 @@ export function NotificationSettingsScreen() {
     const existingPermissions = await Notifications.getPermissionsAsync();
     setPushNotificationPermissions(existingPermissions);
   }, []);
+
+  const query = { viewer: null };
+  const hasEarlyAccess = useMemo(() => {
+    return query.viewer?.user?.roles?.includes('EARLY_ACCESS') ?? false;
+  }, [query]);
+
+  const emailNotificationSettingData = useMemo(
+    () => [
+      {
+        key: 'notifications',
+        title: 'Notifications',
+        description: 'Weekly summary of your unread notifications',
+      },
+      {
+        key: 'marketing',
+        title: 'General Marketing',
+        description: 'Product updates, artist collabs, and airdrops',
+      },
+      // Conditionally add the 'Members Club' entry if `hasEarlyAccess` is true
+      ...(hasEarlyAccess
+        ? [
+            {
+              key: 'membersClub',
+              title: 'Members Club',
+              description: 'Exclusive updates for Members Club Holders',
+            },
+          ]
+        : []),
+      {
+        key: 'digest',
+        title: 'Digest',
+        description: 'Weekly digest of top interacted galleries, artists, and posts',
+      },
+    ],
+    [hasEarlyAccess]
+  );
 
   useEffect(() => {
     checkPushNotificationPermission();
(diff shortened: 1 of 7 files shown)


apps/mobile/src/screens/SettingsScreen/NotificationSettingsScreen.tsx
```diff
@@ -85,3 +326,22 @@ export function NotificationSettingsScreen() {
...
+  readonly unsubscribedFromNotifications: boolean;
+};
+
+type EmailNotificationSettings = {
+  notifications: boolean;
+  marketing: boolean;
+  digest: boolean;
+  membersClub: boolean;
+  all: boolean;
+  [key: string]: boolean;
```
Robinnnnn, line 346: do we need this line?

apps/mobile/src/screens/SettingsScreen/NotificationSettingsScreen.tsx
```diff
@@ -55,11 +121,186 @@ export function NotificationSettingsScreen() {
...
+  }, [query]);
+
+  const shouldShowEmailSettings = useMemo(
+    () => userEmail && !isEmailUnverified,
+    [userEmail, isEmailUnverified]
+  );
+
+  const currentEmailNotificationSettings = query?.viewer?.email?.emailNotificationSettings;
+
+  const [emailSettings, setEmailSettings] = useState<EmailNotificationSettings>({
```
Robinnnnn, line 137: so i think we can definitely share a lot of this business logic across mobile <> web!

can use something like
```js
export function useEmailNotificationSettings() {
  const [emailSettings, setEmailSettings] = useState...
  
  const computeIsToggleChecked = ... // small rename here since it's a func

  const updateEmailNotificationSettings = useUpdateEmailNotificationSettings()

  const handleToggle = async ...

  return useMemo(() => ({
    emailSettings,
    computeToggleChecked,
    updateEmailNotificatinSettings,
    handleToggle,
    ...
  }), [...])
}
```

as you play with it you'll start to see what can be self-contained within the hook vs. returned out of the hook for access

Robinnnnn: for the areas that are slightly different for mobile vs. web, such as displaying a toast, that can be passed in as a callback to the function from each platform.

e.g.
```js
const handleToggle = useCallback(async (settingType: string, settingTitle: string, onSuccess, onFailure) => {
  ...
  try {
     ...
     onSuccess()
  } catch (e) {
     onFailure(e)
  }
```

let's also make sure that if we're passing in a bunch of variables into `handleToggle`, we pass in a single object for more explicit parameterization

apps/mobile/src/screens/SettingsScreen/NotificationSettingsScreen.tsx
Rohan-cp: yeah adding the index signature just fixes this ts flag 
Screenshot 2024-03-12 at 15 19 15

apps/mobile/src/screens/SettingsScreen/NotificationSettingsScreen.tsx
Rohan-cp: got ittt, made the changes now!

packages/shared/src/hooks/useUpdateEmailNotificationSettings.ts
```diff
@@ -0,0 +1,246 @@
...
+  readonly unsubscribedFromNotifications: boolean;
+};
+
+type updateEmailNotificationSettingsProps = {
+  unsubscribedFromNotifications: boolean;
+  unsubscribedFromDigest: boolean;
+  unsubscribedFromMarketing: boolean;
+  unsubscribedFromMembersClub: boolean;
+  unsubscribedFromAll: boolean;
+};
```
Robinnnnn, line 233: these look like the same types, could we consolidate?

maybe we can just use the `EmailSettings` type everywhere?

packages/shared/src/hooks/useUpdateEmailNotificationSettings.ts
```diff
@@ -0,0 +1,246 @@
...
+  settingTitle: string;
+  pushToast: ({ message }: { message: string }) => void;
+};
+
+type handleEmailNotificationChangeProps = {
+  settingType: string;
+  newSettingValue: boolean;
+  settingTitle: string;
+  pushToast: ({ message }: { message: string }) => void;
+};
```
Robinnnnn, line 246: we can build the second type on top of the first :)

```
type handleEmailNotificationChangeProps = handleToggleProps & { newSettingsValue: boolean; }
```

packages/shared/src/hooks/useUpdateEmailNotificationSettings.ts
```diff
@@ -0,0 +1,246 @@
...
+  );
+}
+
+type EmailNotificationSettings = {
+  notifications: boolean;
+  marketing: boolean;
+  digest: boolean;
+  membersClub: boolean;
+  all: boolean;
+  [key: string]: boolean;
```
Robinnnnn, line 216: ah i see what you're saying, it's fine we can fix this, will comment separately :)

packages/shared/src/hooks/useUpdateEmailNotificationSettings.ts
```diff
@@ -0,0 +1,246 @@
...
+  const computeToggleChecked = useCallback(
+    (notifType: string | number) => {
+      // if the user dont have an email or not verified, we want to toggle off
+      if (!shouldShowEmailSettings) {
+        return false;
+      }
+      return emailSettings[notifType];
+    },
+    [shouldShowEmailSettings, emailSettings]
+  );
```
Robinnnnn, line 62: here we can update L54 to accept this type of input. then we shouldn't have to do the `[key: string]: boolean` trickery below

image

packages/shared/src/hooks/useUpdateEmailNotificationSettings.ts
```diff
@@ -0,0 +1,246 @@
...
+            response.updateEmailNotificationSettings.viewer.email.emailNotificationSettings;
+          if (settings[settingType as keyof EmailSettings] === newSettingValue) {
+            setEmailSettings(settingsUpdate);
+            pushToast({
+              message: `Settings successfully updated. You will ${
+                settingsUpdate[settingType] ? 'no longer' : 'now'
+              } receive ${settingTitle} emails.`,
+            });
+          }
+        }
```
Robinnnnn, line 149: you may encounter more typecheck issues here. to fix, simply edit the `settingType` field on `handleToggleProps` below

image

then we should be able to toss this `settingType as keyof EmailSettings` and everything should just work!

apps/web/src/components/Settings/ManageEmailSection/ManageEmailSection.tsx
```diff
@@ -20,112 +19,76 @@ type Props = {
...
+            },
+          ]
+        : []),
+      {
+        key: 'digest',
+        title: 'Weekly Digest',
+        description: 'Featured Galleries, Collections and more',
+      },
+    ],
+    [hasEarlyAccess]
```
Robinnnnn, line 83: 1. is there a reason this is copy is distinct from mobile? e.g. marketing email description is `'Product updates, artist collabs, and airdrops'` whereas on web it's just `'Product marketing emails'`
2. we should consolidate this into the hook as well :) and since the hook receives `hasEarlyAccess` bool, we'll be able to return the relevant settings dynamically

packages/shared/src/hooks/useUpdateEmailNotificationSettings.ts
Rohan-cp: done!

packages/shared/src/hooks/useUpdateEmailNotificationSettings.ts
Rohan-cp: made the change!

apps/web/src/components/Settings/ManageEmailSection/ManageEmailSection.tsx
Rohan-cp: I consolidated the hooks and regarding the copy, the designs had a different one for mobile and the one for web was from this other screenshot in your loom. Should I bring this up on the slack thread maybe?

packages/shared/src/hooks/useUpdateEmailNotificationSettings.ts
Rohan-cp: working now!

Robinnnnn (approved):

PR #2373, opened 2024-03-19
Refactor mobile nft selector

jakzaizzat:
### Summary of Changes

Refactor on how we implement the `NftSelector` for `Onboarding`, `Post`, `CommunityPost` and `PFP`. This refactoring will be benefit for our gallery editor feature.

**Before**

The single screen `NftSelector` handle all the logic for all the scenario above and it make its hard to debug and dev dont have much flexibility.

**After**

We split up all the screen for each flow. 

Example: **Onboarding** gonna have their own screen instead of sharing between other use cases. 

We have a independent component to render NFT selector

1. `<NftSelectorHeader/>`
2. `<NftSelectorToolbar/>`
3. `<NftSelectorPickerGrid/>`

All the logic that defined after the user select the NFT will be defined in each screen instead of component itself

### Edge Cases

1. Onboarding flow
2. PFP picker from profile screen
3. Create a post from bottom navbar
4. Create a post from community screen
5. Switching between collector and creator profile
6. Sync owned and created tokens

### Testing Steps

### Checklist


apps/mobile/src/screens/Onboarding/OnboardingNftSelectorContractScreen.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+import { NftSelectorContractHeader } from '~/components/NftSelector/NftSelectorContract/NftSelectorContractHeader';
+import { NftSelectorContractPickerGrid } from '~/components/NftSelector/NftSelectorContract/NftSelectorContractPickerGrid';
+import { NftSelectorContractWrapper } from '~/components/NftSelector/NftSelectorContract/NftSelectorContractWrapper';
+import { OnboardingNftSelectorContractScreenQuery } from '~/generated/OnboardingNftSelectorContractScreenQuery.graphql';
+import { MainTabStackNavigatorParamList, MainTabStackNavigatorProp } from '~/navigation/types';
+
+import { useProfilePicture } from '../NftSelectorScreen/useProfilePicture';
+
+export function OnboardingNftSelectorContractScreen() {
```
kaitoo1, line 13: I love separating these out into different components depending on usage, but it looks like `OnboardingNftSelectorContractScreen` and `PfpSelectorContractScreen` are 99% the same code.Is it possible to consolidate the shared code a bit more? 

I'm imagining `OnboardingNftSelectorContractScreen` and `PfpSelectorContractScreen` to be very thin wrappers around a shared component that contains most of the code that's the same, and the wrappers only add the unique things like the `fullscreen` prop or the custom `handleSelectNft` handler in the case of `PostNftSelectorContractScreen`. 

the main concern would be having to maintain the same logic in 3 different places, and repeating it again when we add another variant of the SelectorContractScreen component

kaitoo1: Same comment for `SelectorScreen` components like `PostNftSelectorScreen`, `PfpSelectorScreen`,

jakzaizzat: That make sense! Clean up majority of the code so that no redundant code

kaitoo1 (approved):
sick!!! thanks for addressing the comment, this looks great. each of the unique wrapper components feels a lot slimmer now

PR #2378, opened 2024-03-23
fix to remove empty feed items

Rohan-cp:
### Demo or Before/After Pics

Before
Screenshot 2024-03-23 at 15 45 18

now removed

### Edge Cases

### Testing Steps

### Checklist

### DIFF
--- a/apps/web/src/components/Feed/FeedEventItem.tsx
+++ b/apps/web/src/components/Feed/FeedEventItem.tsx
@@ -86,6 +86,11 @@ export default function FeedEventItemWithBoundary({
           ... on UserFollowedUsersFeedEventData {
             __typename
           }
+          ... on GalleryUpdatedFeedEventData {
+            subEventDatas {
+              __typename
+            }
+          }
         }
         ...FeedEventSocializeSectionFragment
         ...FeedEventItemFragment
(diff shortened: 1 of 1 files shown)


apps/web/src/components/Feed/FeedEventItem.tsx
```diff
@@ -110,6 +115,18 @@ export default function FeedEventItemWithBoundary({
...
+    // ignore GalleryInfoUpdatedFeedEventData events because we have no component to handle that right now
+    const subEvents = event?.eventData?.subEventDatas
+      ?.slice(0, 4)
+      .filter((event) => event.__typename !== 'GalleryInfoUpdatedFeedEventData');
+
+    if (!subEvents?.length) {
+      return null;
+    }
+  }
+
```
Rohan-cp, line 129: kind of a hacky fix but basically was trying to ignore the eventDataType`GalleryInfoUpdatedFeedEventData` because we don't have a component for it from this file so it would always be an empty item for this event type. I wanted to know if you had a better idea of how I could fix this @Robinnnnn! thanks

Screenshot 2024-03-23 at 15 43 19

Robinnnnn: sorry but why can't we just introduce a new component in the screenshot you provided?

PR #2417, opened 2024-04-09
add SuggestedUser card row in mobile feed

Rohan-cp:
### Summary of Changes
Adds a new suggested profile section in the Feed, that shows up `4` feed items down the list.

### Demo 

https://github.com/gallery-so/gallery/assets/49758803/a3c4daa9-777c-49fd-adf3-71fab524c33a

### Edge Cases

light mode

Screenshot 2024-04-11 at 22 13 23

### Testing Steps

### Checklist

### DIFF
--- a/apps/mobile/src/components/Feed/createVirtualizedFeedEventItems.ts
+++ b/apps/mobile/src/components/Feed/createVirtualizedFeedEventItems.ts
@@ -398,7 +398,7 @@ export function createVirtualizedFeedEventItems({
           key: 'suggested-profile-row',
           eventId: 'suggested-profile-row',
           itemType: null,
-        });
+        } as FeedListItemType);
         insertedRow = true;
       }
     }
--- a/apps/mobile/src/components/Feed/createVirtualizedFeedEventItems.ts
+++ b/apps/mobile/src/components/Feed/createVirtualizedFeedEventItems.ts
@@ -382,9 +382,15 @@ export function createVirtualizedFeedEventItems({
     }
   }
 
+  /**
+   * insert suggested profiles in between posts.
+   *
+   * we need to do it this way as opposed to a simple splice based on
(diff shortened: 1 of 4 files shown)


apps/mobile/src/components/Feed/createVirtualizedFeedEventItems.ts
```diff
@@ -370,5 +382,28 @@ export function createVirtualizedFeedEventItems({
...
+          queryRef: query,
+          key: 'suggested-profile-row',
+          eventId: 'suggested-profile-row',
+          itemType: null,
+        } as FeedListItemType);
+        insertedRow = true;
+      }
+    }
+    itemsWithSuggestedProfileRow.push(item);
+  }
```
Robinnnnn, line 406: 1. could we add an explainer / comment block for what's going on here?
2. do we need to mark `insertedRow` as `true`? i assume `postCount === SUGGESTED_PROFILE_ROW_IDX` would only be true a single time anyway?

overall i'm wondering if there could be a simpler way to do this without the loop, like:
```ts
const itemsWithSuggestedProfileRow = [
  ...items.slice(0, 8),
  { kind: 'suggested-profile-row', /* rest of the object */ }
  ...items.slice(8)
]
```

Rohan-cp: yo @Robinnnnn, you're right we don't need the `insertedRow` variable, removed that now. 

Regarding using the splice, I think we can't use the hardcoded value for the index because the items of the list are `feed-post-header`, `feed-post-socialize-section`, etc (basically subsections of a single post in the feed for all the posts) so we want to add the row only when it is before a `feed-post-header` 

Or it might end up after a post and before its mint button for example.

Screenshot 2024-04-12 at 00 28 45

Robinnnnn (approved):

PR #2462, opened 2024-05-07
Farcaster Improvements

pvicensSpacedev:
### Summary of Changes

- Added activity indicator when signing up with farcaster (mobile). Also added function to check wether the user canceled warpcast login (the onError from the library does not work so this is just a workaround) 
- Prefilled username input when farcaster signup (mobile & web)
- Aded necessary types 

https://linear.app/galleryso/issue/GAL-5603/various-auth-polish

### Demo or Before/After Pics
- Username from farcaster

https://github.com/gallery-so/gallery/assets/112896251/3109ce31-ce13-46be-8818-2a26b8590dcf

- Simple activity indicator
image

- Hide modal when email signup
Screenshot 2024-05-07 at 12 23 43

### Edge Cases

### Testing Steps

### Checklist

### DIFF
--- a/apps/mobile/src/components/Login/AuthProvider/Farcaster/FarcasterAuthProvider.tsx
+++ b/apps/mobile/src/components/Login/AuthProvider/Farcaster/FarcasterAuthProvider.tsx
@@ -34,7 +34,11 @@ export function FarcasterAuthProvider({ children }: { children: ReactNode }) {
   return <AuthKitProvider config={config}>{children}</AuthKitProvider>;
 }
 
-export function useLoginWithFarcaster() {
+type FarcasterBottomSheetRowProps = {
+  setIsFarcasterLoading: (isLoading: boolean) => void;
+};
+
+export function useLoginWithFarcaster({ setIsFarcasterLoading }: FarcasterBottomSheetRowProps) {
   const { hideBottomSheetModal } = useBottomSheetModalActions();
 
   const getUsersByWalletAddresses = useGetUsersByWalletAddressesImperatively();
@@ -53,6 +57,7 @@ export function useLoginWithFarcaster() {
 
   const handleFarcasterLoginError = useCallback(
     (error?: AuthClientError | Error) => {
+      setIsFarcasterLoading(false);
       const errorMessage = error?.message ?? 'unknown error';
       pushToast({
         message: `There was an error signing in with Farcaster: ${errorMessage}`,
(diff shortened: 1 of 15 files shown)


apps/mobile/src/components/Login/SignInBottomSheet.tsx
```diff
@@ -87,13 +87,46 @@ function SignInBottomSheet({ onQrCodePress, openManageWallet }: Props) {
...
+    });
+
+    return () => {
+      subscription.remove();
+    };
+  }, [appState, isPolling, isConnected, isSuccess]);
+
+  const handleConnectFarcasterPress = useCallback(() => {
+    handleConnectFarcaster(attemptReconnect);
+  }, [attemptReconnect, handleConnectFarcaster]);
```
Robinnnnn, line 122: so the only thing i'm concerned about with this PR is farcaster-related-logic being split in this component (`FarcasterBottomSheetRow`) as well as the main hook `useLoginWithFarcaster`

from what i can tell, the `appState / useState` and `useEffect` can both live inside `useLoginWithFarcaster`. the only variable used by this component is `isFarcasterLoading`, but even that i think can be a state that's defined inside `useLoginWithFarcaster` and exported out for this component to use

apps/mobile/src/components/Login/SignInBottomSheet.tsx
```diff
@@ -87,13 +87,46 @@ function SignInBottomSheet({ onQrCodePress, openManageWallet }: Props) {
...
+  });
+
+  const [appState, setAppState] = useState(AppState.currentState);
+
+  useEffect(() => {
+    const subscription = AppState.addEventListener('change', (nextAppState) => {
+      if (appState.match(/inactive|background/) && nextAppState === 'active') {
+        if (isPolling || (!isConnected && !isSuccess)) {
+          setAttemptReconnect(true);
+        }
```
Robinnnnn, line 109: this logic makes sense in that if the connection did not go through, we attempt to reconnect from scratch.

_is there any chance the connection is actually successful_, but our app doesn't pick up on it yet by the time the user arrives back on our app? just wondering if we need to wait at all before immediately going into reconnection mode

apps/mobile/src/components/Login/AuthProvider/Farcaster/FarcasterAuthProvider.tsx
```diff
@@ -182,19 +192,44 @@ export function useLoginWithFarcaster() {
...
+    async (shouldAttemptReconnect = false) => {
+      setIsFarcasterLoading(true);
 
-      if (isConnectError) {
+      if (shouldAttemptReconnect && !isSuccess && url) {
         reconnect();
+        const { nonce } = await createNonce();
+        setNonce(nonce);
+        signIn();
+        Linking.openURL(url);
```
Robinnnnn, line 204: if we're attempting to reconnect i don't think we need to create a new nonce! so we can probably remove L201-202

